PR #211, opened 2019-11-26
Removes the circular dependency of the chart heatfield.

lukasholzer:
Removes the circular dependency of the chart

### DIFF
--- a/components/chart/src/chart.ts
+++ b/components/chart/src/chart.ts
@@ -41,25 +41,31 @@ import {
   ViewChild,
   ViewEncapsulation,
 } from '@angular/core';
+import {
+  createInViewportStream,
+  DtViewportResizer,
+  isDefined,
+} from '@dynatrace/barista-components/core';
+import { DtTheme } from '@dynatrace/barista-components/theming';
 // tslint:disable-next-line:no-duplicate-imports
 import * as Highcharts from 'highcharts';
 // tslint:disable-next-line:no-duplicate-imports
 import {
-  ChartObject,
-  Options as HighchartsOptions,
-  IndividualSeriesOptions,
   addEvent as addHighchartsEvent,
   chart,
+  ChartObject,
+  IndividualSeriesOptions,
+  Options as HighchartsOptions,
   setOptions,
 } from 'highcharts';
 import { merge as lodashMerge } from 'lodash';
 import {
   BehaviorSubject,
+  defer,
+  merge,
   Observable,
   Subject,
   Subscription,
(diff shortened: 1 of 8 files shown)


components/chart/src/heatfield/chart-heatfield.ts
```diff
@@ -201,30 +188,44 @@ export class DtChartHeatfield extends _DtHeatfieldMixinBase
...
+  /**
+   * Highcharts charts object is needed for the xAxis is
+   * set by the chart via the `_initHeatfield` function that
+   * is called by the chart.
+   */
+  private _chartObject: Highcharts.ChartObject;
+
   constructor(
     elementRef: ElementRef,
-    // tslint:disable-next-line: no-forward-ref
```
tomheller: You are removing parameters from the constructor, which in this instance is considered a breaking change. 
Please change the commit message to a `fix` component and add a `Breaking Change` codename.

components/chart/src/heatfield/chart-heatfield.ts
```diff
@@ -201,30 +188,44 @@ export class DtChartHeatfield extends _DtHeatfieldMixinBase
...
     private _changeDetectorRef: ChangeDetectorRef,
   ) {
     super(elementRef);
-    this._checkChartSupport();
   }
 
   ngOnDestroy(): void {
-    this._destroyed.next();
-    this._destroyed.complete();
+    this._destroy$.next();
```
tomheller: 💚 for consistency and actually following our coding standards

components/chart/src/chart.ts
```diff
@@ -41,25 +41,31 @@ import {
...
 // tslint:disable-next-line:no-duplicate-imports
 import * as Highcharts from 'highcharts';
 // tslint:disable-next-line:no-duplicate-imports
 import {
-  ChartObject,
-  Options as HighchartsOptions,
-  IndividualSeriesOptions,
   addEvent as addHighchartsEvent,
   chart,
+  ChartObject,
```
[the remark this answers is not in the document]
lukasholzer: There is no import sorter, but due to the fact that my build breaks on every unused import, i use order imports of vs code to let my build pass. @ffriedl89, @thomaspink in my opinion, we should discuss this in a separate topic in case that this slows down developer productivity extremely!

components/chart/src/heatfield/chart-heatfield.ts
lukasholzer: haha thx :)

Issue #148, opened 2019-11-20
Barista: Color grid component


lara-aigmueller:
Make a component that displays a grid of all colors of one shade, based on the colors defined in the colors-CSS-file.

Usage could be: `<ba-color-grid color="green"></ba-color-grid>`

Used on `resources/colors`

lara-aigmueller:
Not part of "Barista v2", will be hardcoded for now.

Issue #147, opened 2019-11-20
Barista: Color component


lara-aigmueller:
We need an Angular component that displays a color + HEX and CMYK value and has those fancy copy hex/cmyk buttons.

The component's input should be the color-name, e.g. `green-100` - look up the actual color hex value in a CSS file.

lara-aigmueller:
Not part of "Barista v2", will be hardcoded for now.

PR #227, opened 2019-11-27
Barista/create color page

NinaKammerlander:

<hr>
Hi, thank you for contributing to the Dynatrace Angular Component library (PR).

  often helpful).
  guidelines.

Please fill out the checkbox below: <br>

#### Type of PR

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or change that would cause existing functionality to
      not work as expected)
- [ ] Documentation update (changes to documentation)
- [ ] Other (if none of the above apply)

#### Checklist

### DIFF
--- a/apps/barista/src/styles/items/_color.scss
+++ b/apps/barista/src/styles/items/_color.scss
@@ -1,0 +1,111 @@
+.ba-color {
+  display: inline-block;
+  vertical-align: top;
+  min-width: 10rem;
+  width: 10rem;
+  margin-bottom: 1.5rem;
+
+  /* Color width if used without a colorgroup */
+  @media (min-width: $screen-lg) {
+    width: 24%; /* 4 colors per row */
+  }
+
+  @media (min-width: $screen-xl) {
+    width: 19%; /* 5 colors per row */
+  }
+
+  @media (min-width: $screen-xxl) {
+    width: 16%; /* 6 colors per row */
+  }
+
+  @media (min-width: $screen-xxxl) {
+    width: 14%; /* 7 colors per row */
+  }
+}
+
+.ba-color__block {
+  height: 7rem;
+  position: relative;
+  overflow: hidden;
+  margin-bottom: 0.5rem;
+  display: flex;
+  flex-direction: column;
+}
+
+.ba-color__copy {
+  position: relative;
+  appearance: none;
+  display: block;
+  width: 100%;
+  flex: 1;
+  background-color: $turquoise-600;
+  border: 0;
+  font-weight: 600;
+  opacity: 0;
+  color: #fff;
+  outline: none;
+  box-shadow: 0;
+  overflow: hidden;
+}
+
+.ba-color .ba-color__name {
+  margin-bottom: 0;
+}
+
+.ba-color .ba-color__metadata {
+  color: $gray-400;
+  font-size: 0.9rem;
+  margin-bottom: 0;
+}
+
+/**
+* Colorgroup wrapper styles
+*/
+
+.ba-colorgroup {
+  width: 100%;
+
+  &.ba-colorgroup--textwidth {
+    @include textContentWidth();
+  }
+
+  .ba-color {
+    width: 100%;
+    min-width: unset;
+  }
+
+  @media (min-width: $screen-xs) {
+    display: grid;
+    grid-template-columns: repeat(2, 1fr);
+    grid-column-gap: 0.5rem;
+
+    &.ba-colorgroup--textwidth {
+      display: grid;
+    }
+  }
+
+  @media (min-width: $screen-sm) and (max-width: $screen-md) {
+    grid-template-columns: repeat(3, 1fr);
+  }
+
+  @media (min-width: $screen-lg) {
+    grid-template-columns: repeat(4, 1fr);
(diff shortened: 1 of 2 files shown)


apps/barista/src/pages/page-content.ts
```diff
@@ -35,4 +35,37 @@ export class BaPageContent {
...
+    input.select();
+    try {
+      target!.classList.add('ba-is-active');
+      setTimeout(() => {
+        target!.classList.remove('ba-is-active');
+      }, 800);
+      document.execCommand('copy');
+      input.blur();
+    } catch (err) {
+      /* #hupsi */
```
lara-aigmueller: 😅

apps/barista/src/styles/items/_color.scss
```diff
@@ -0,0 +1,147 @@
...
+  @media (min-width: $screen-xxl) {
+    width: 16%; /* 6 colors per row */
+  }
+
+  @media (min-width: $screen-xxxl) {
+    width: 14%; /* 7 colors per row */
+  }
+}
+
+.ba-color__block {
```
lara-aigmueller: Could we remove BEM-syntax?

NinaKammerlander: I did not change the BEM-syntax because I copied the HTML from the different color pages and they contained this syntax. It is only a temporary solution and BEM will be removed as soon as a color component is implemented.

apps/barista/src/pages/page-content.ts
```diff
@@ -35,4 +35,37 @@ export class BaPageContent {
...
+      allColorItems.forEach(colorItem => {
+        colorItem.addEventListener('click', this._handleCopyColor);
+      });
+    }
+  }
+
+  /**
+   * TODO: remove this function as soon as a color component is implemented!
+   * save a color code to the clipboard
+   */
```
tomheller: Maybe you can reference issue #147 and #148 here

apps/barista/src/pages/page-content.ts
```diff
@@ -35,4 +35,37 @@ export class BaPageContent {
   private _data = '';
 
   constructor(private _elementRef: ElementRef<HTMLElement>) {}
+
+  ngAfterContentInit(): void {
+    const allColorItems = document.querySelectorAll('.js-ba-copybutton');
```
thomaspink: Is this also removed once we have the color components in place? Can we change the class name?

lara-aigmueller:
> Not quite sure why we need this workaround until we have the color component in place? Because the color component is basically just this workaround? Why aren't we implementing the color component instead?

Because when we implement the color component and the color grid, we somehow want to get the colors from our css files and maybe refactor the existing code etc. I'm pretty sure if we start making 1-2 new components, it takes us more time until everything is working. The solution we have now works and this is what we want at the moment ;-)

tomheller:
In my opinion, this is something that can be touched later. If this works now, we should merge it and reiterate over this once Barista design system page is out the door. I don't want to block the release of it because we are rebuilding the whole colors page, which would also need some love designwise.

thomaspink:
I still think that implementing the color component and the grid should not take to long, but i am fine for now.

apps/barista/src/pages/page-content.ts
```diff
@@ -35,4 +35,37 @@ export class BaPageContent {
...
+    if (allColorItems && allColorItems.length > 0) {
+      allColorItems.forEach(colorItem => {
+        colorItem.addEventListener('click', this._handleCopyColor);
```
thomaspink: Event listener should be closed when the component is destroyed

apps/barista/src/pages/page-content.ts
```diff
@@ -35,4 +35,37 @@ export class BaPageContent {
   private _data = '';
 
   constructor(private _elementRef: ElementRef<HTMLElement>) {}
+
+  ngAfterContentInit(): void {
+    const allColorItems = document.querySelectorAll('.js-ba-copybutton');
+    if (allColorItems && allColorItems.length > 0) {
+      allColorItems.forEach(colorItem => {
```
thomaspink: Pls use forof here

tomheller:
> I still think that implementing the color component and the grid should not take to long, but i am fine for now. We should change it asap.
> Let's also be really careful with copying legacy code because it just works and we wanna publish it no matter what. We should stick to our guidelines and best practices, because if wo don't do it, why should others.

Yes, I agree. I was not entirely referring to the javascript part of the pull request when talking about the color page. 
I thought that we could reuse the markup and parts of the styling to not reinvent what we already have. I think that we can safely drop the copy to clipboard and non angular javascript for now. I'd rather have the information on the page and maybe miss one or two _convenience_ features, rather than not have the information at all.

apps/barista/src/pages/page-content.ts
tomheller: I think when the component is destroyed, all it's child dom elements are removed from the DOM, which usually clears their event handlers (unless angular caches the components for re-rendering somewhere?). 

To avoid non angular legacy scripts, can we drop the copy color functionality for now?

PR #255, opened 2019-12-02
ci: Run affected targets correctly on master and on PR as on 4.x and 1.15x

lukasholzer:

Calculate correct commit hashes so that affected knows always against which target it should compare. Currently, we were always running against master.

- `4.x` against latest commit between the changes and previous commits
- `4.14.x` against latest commit between the changes and previous commits
- `master` against latest commit between the changes and previous commits
- PRs check for the target hash
- commits without PR against origin/master
#### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,6 +42,61 @@ npm_cache: &npm_cache
     - npm-packages-{{ checksum "package-lock.json" }}
     - npm-packages-
 
+# The last sha of the branch that holds the previous commit sha.
+# could be either master, 4.x or 4.10.x
+sha_branch_cache: &sha_branch_cache
+  key: last-{{ .Environment.CIRCLE_BRANCH }}-sha
+
+set_env: &set_env
+  name: Setup Environment Variables
+  command: |
+    BASE_URL="https://api.github.com/repos/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}"
+    MASTER_SHA_FILE=./.config/last-${CIRCLE_BRANCH}-sha.txt
+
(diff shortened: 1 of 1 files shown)


ffriedl89:
I think this introduces a lot of complexity for the benefits we get, we have enough resources to run the entire pipeline on master, release branches. If we run into problems in the future we can optimize

lukasholzer:
We had the shortages of resources when we had the 4.x and 4.10.x branches and hat to cherry-pick them. 

Currently, we have build times about 10minutes per commit when we need to run all. And we don't run the `build` and `a11y` tests currently on the CI. That means we are using per commit all of the build resources in case we can run max 4 in parallel and we have about 10 jobs that need to be running.

When the release branches are added and we are cherrypicking always commits than those commits have to be built as well alongside the master.

We are trying to create small pull requests – defacto a lot of builds. If we can reduce our CI times and CI waiting times than this is a must in my opinion.

Furthermore, a configuration of the CI is not touched every time in best case it is an initial effort.
Nevertheless compared with the complexity of our old build (Jenkins + Gradle) this is very straight forward.

lukasholzer:
By the way, configuring that only on PRs the affected should be leveraged is quite the same complexity.

ffriedl89:
After talking yesterday, i am fine with the current approach, since it solves a CI bottleneck that we will again experience once we have a 5.x and a 5.0.x branch. Please add troubleshooting the CI information either in this PR or create a followup PR

lukasholzer:
Will add the debugging with ssh https://circleci.com/blog/optimizing-react-native-camera/ to some kind of documentation.

tomheller:
Please also link the referred wiki documentation on how to troubleshoot

PR #304, opened 2019-12-06
build(components): Fixes the components build

thomaspink:


thomaspink:
Closed in favour of #306

PR #319, opened 2019-12-10
barista: Hide sidenav on small screens and fix line-height.

lara-aigmueller:
`#hupsi`


NinaKammerlander:
hupsi... thanks! :monkey_face:

PR #333, opened 2019-12-11
fix(filter-field): Fixes and issue where filters are not immediately removed when clicking the clear all button

### LINKED ISSUE #176: Filter field: Filters are not immediately removed when clicking the clear all button
You have to click somewhere outside to trigger the removal of the filter field.
To me this look like a CD issue we have here. Maybe we have to trigger it by hand somewhere.

[See screencast](https://github.com/Dynatrace/barista/files/3874154/Filter.field._.Barista.Design.System.-.Google.Chrome.2019-11-21.11-08-01.mp4.zip)

thomaspink:
fix(filter-field): Fixes and issue where filters are not immediately removed when clicking the clear all button

Fixes #176

### DIFF
--- a/components/filter-field/src/filter-field.ts
+++ b/components/filter-field/src/filter-field.ts
@@ -617,8 +617,7 @@ export class DtFilterField<T> implements AfterViewInit, OnDestroy, OnChanges {
   }
 
   /** @internal Clears all filters and switch to root def. */
-  _clearAll(event: Event): void {
-    event.stopPropagation();
+  _clearAll(): void {
     // Only filters that are deletable should be removed
     // We need to aggregate the once that should be removed on the one hand,
     // so we can emit them to the consumer.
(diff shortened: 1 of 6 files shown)


tomheller:
PR looks good, but can we adde some tests that would test this behaviour?

components/filter-field/src/filter-field.ts
```diff
@@ -618,7 +618,6 @@ export class DtFilterField<T> implements AfterViewInit, OnDestroy, OnChanges {
 
   /** @internal Clears all filters and switch to root def. */
   _clearAll(event: Event): void {
```
tomheller: remove `event`, if we no longer need it

thomaspink:
Sadly, this can not be tested via unit tests because the `event.stopPropagation()` stops zone(angular) from running CD - but in unit tests you trigger CD by hand which means that the clearAll button will alway work there (thats why no tests are failing).

Will try to come up with an e2e test

tomheller:
I guess some e2e tests wont hurt the filter field anyways, let's see if we can add a test for to be on the save side.

Issue #343, opened 2019-12-16
TS2742 The inferred type of … cannot be named without a reference to … 


ermin-muratovic:
# Bug Report

When building the barista components library on Windows I get numerous TS2742 errors, like for example:

`../../components/table/src/expandable/expandable-row.ts:146:33 - error TS2742: The inferred type of '_collapsedStream' cannot be named without a reference to 'C:/Users/ermin.muratovic/IdeaProjects/keptn-workspace/barista/node_modules/rx
js'. This is likely not portable. A type annotation is necessary.
`

## Steps to Reproduce

1. Run `npm run build` on a Windows machine.

## Context (Environment)

This issue seems to appear only on Windows.

### Used Versions:

- **node**: v12.13.1
- **npm**: 6.12.1
- **typescript**: 3.4.3
- **angular**: 8.3.19
- **@angular/cdk**: 8.3.19
- **@dynatrace/barista-components**: 4.15.0

## Possible Implementation

Add explicit type annotation.

## Attachments

Log file of the `npm run build` is attached, see [build_log.txt](https://github.com/dynatrace-oss/barista/files/3967422/build_log.txt)

PR #344, opened 2019-12-16
Fix build on Windows by adding explicit type annotations

ermin-muratovic:
I have added explicit type annotations where they were missing and breaking the build. For details see issue #343

### DIFF
--- a/components/core/src/animations/error-animations.ts
+++ b/components/core/src/animations/error-animations.ts
@@ -14,14 +14,21 @@
  * limitations under the License.
  */
 
-import { animation, animate, style } from '@angular/animations';
+import {
+  animation,
+  animate,
+  style,
+  AnimationReferenceMetadata,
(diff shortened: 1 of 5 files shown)


ermin-muratovic:
Thank you @thomaspink for the suggestions. I hope I did the rebase correctly. I updated the commit message to your suggestion with Fixes and the issue number to autoclose the issue.

ermin-muratovic:
> LGTM: @ffriedl89 but we should investigate the root cause while the type inference is not working on windows. In case it is working on macOS.

The root cause seems to be an issue with `importHelpers` and when the @types directory is a symlink on windows. So maybe it is a file system issue (NTFS on Windows) and that is why it is working fine on Mac and Ubuntu. See also [importHelpers with symlinks breaks type resolution #29221](https://github.com/microsoft/TypeScript/issues/29221).

PR #366, opened 2019-12-18
Barista/live examples actions

lara-aigmueller:

<hr>

  often helpful).
  guidelines.

#### Checklist


apps/barista/src/components/live-example/live-example.html
```diff
@@ -18,21 +22,56 @@
...
+      [class.ba-live-example-control-copied]="(_copied$ | async)"
+      [disabled]="_sourceHidden"
+      (click)="_copyToClipboard()"
+    >
+      <dt-icon *ngIf="(_copied$ | async) === false" name="reports"></dt-icon>
+      <dt-icon *ngIf="(_copied$ | async)" name="checkmark"></dt-icon>
+    </button>
+    <button
+      title="Expand source code"
+      role="button"
```
lukasholzer: i think this a little bit too much button ;D 

```suggestion
```

apps/barista/src/components/live-example/live-example.html
```diff
@@ -18,21 +22,56 @@
...
+      type="button"
```
lukasholzer: here as well to much button

```suggestion
```

apps/barista/src/components/live-example/live-example.ts
```diff
@@ -126,18 +150,49 @@ export class BaLiveExample implements OnDestroy {
...
 
+  /** @internal Sets the active tab. */
   _setActiveTab(tab: 'html' | 'ts' | 'scss'): void {
     this._activeTab = tab;
     this._activeTabChanged = true;
   }
 
+  /** @internal Copies content of currently active tab to clipboard. */
+  _copyToClipboard(): void {
+    if (document) {
```
lukasholzer: make a platform check with cdks platform library in case that universal or jest has a document as an empty object where there is no execCommand on it.

apps/barista/src/components/live-example/live-example.ts
```diff
@@ -153,4 +208,59 @@ export class BaLiveExample implements OnDestroy {
       );
     }
   }
+
+  /**
+   * Creates a non-visible textarea with the given string as value.
+   */
+  private _createTextArea(value: string): HTMLTextAreaElement {
```
lukasholzer: May you move this into a pure function? Better testable and it needs no class context

apps/barista/src/components/live-example/live-example.ts
```diff
@@ -153,4 +208,59 @@ export class BaLiveExample implements OnDestroy {
...
+    // Add syntax highlighting using highlight.js
+    transformedCode = hljs.highlight(type, transformedCode).value;
+
+    return this._wrapCodeLines(transformedCode);
+  }
+
+  /**
+   * Wraps each line of the given code string in a span tag.
+   */
+  private _wrapCodeLines(code: string): string {
```
lukasholzer: Please move this to a pure function as well

apps/barista/src/components/live-example/live-example.ts
```diff
@@ -24,19 +24,24 @@ import {
   ViewContainerRef,
   ComponentRef,
   OnDestroy,
+  ViewEncapsulation,
 } from '@angular/core';
+import { coerceBooleanProperty } from '@angular/cdk/coercion';
 import { EXAMPLES_MAP } from '@dynatrace/barista-components/examples';
+import * as hljs from 'highlight.js';
```
lukasholzer: Please don't import all highlight this will increase the bundle size extremely only import the typescript, html and css packages. Please check bundle size after your changes

apps/barista/src/components/live-example/live-example.ts
```diff
@@ -126,18 +150,49 @@ export class BaLiveExample implements OnDestroy {
...
   }
 
-  /** Whether one of the three sources has been set. */
+  /** @internal Whether one of the three sources has been set. */
   _hasSources(): boolean {
     return Boolean(this.classSource || this.templateSource);
   }
 
+  /** @internal Sets the active tab. */
   _setActiveTab(tab: 'html' | 'ts' | 'scss'): void {
```
lukasholzer: NIT: you might could write a `type soureType = 'html' | 'ts' | 'scss'` or enum ;)

package.json
```diff
@@ -126,6 +127,7 @@
     "gulp-clean": "^0.4.0",
     "gulp-util": "^3.0.8",
     "highcharts": "^6.0.7",
+    "highlight.js": "^9.17.1",
```
lukasholzer: `highlight.js` is a dependency in case that we are using it in barista. Please reinstall it as dependency

apps/barista/src/components/live-example/live-example.ts
```diff
@@ -24,19 +24,24 @@ import {
...
 @Component({
   selector: 'ba-live-example',
   templateUrl: 'live-example.html',
   styleUrls: ['live-example.scss'],
   host: {
+    class: 'ba-live-example',
     '[class.ba-live-example-dark]': 'themedark',
     '[class.ba-live-example-full-width]': 'fullwidth',
   },
+  encapsulation: ViewEncapsulation.None,
```
tomheller: I'm guessing you disabled the ViewEncapsulation to have the global stylesheet working, but would it also be possible to `@import` the highlight styles in the `live-example.scss` file and keep viewEncapsulation on?

ffriedl89:
I'd suggest using https://www.npmjs.com/package/highlight-ts since this is treeshakable
see https://bundlephobia.com/result?p=highlight-ts@9.12.1-2
in comparison https://bundlephobia.com/result?p=highlight.js@9.17.1

apps/barista/src/components/live-example/live-example.ts
lara-aigmueller: I disabled the view encapsulation to be able to style the code-lines I'm adding in the `_wrapCodeLines` function. Is there another way to do it better and with view encapsulation turned on?

lukasholzer: you can always break out of the view encapsulation with the `:host` attribute but can you provide a simple example?

apps/barista/src/components/live-example/live-example.ts
```diff
@@ -111,33 +132,89 @@ export class BaLiveExample implements OnDestroy {
...
+          copyText = this._classSource;
+          break;
+        case 'scss':
+          // TODO: copyText = this._scssSource;
+          break;
+      }
+
+      if (copyText) {
+        const textarea = createInputElement(copyText, 'textarea');
+        document.body.append(textarea);
```
lukasholzer: please inject the document instead of accessing it directly.

apps/barista/src/utils/utils.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { wrapCodeLines } from './wrap-code-lines';
+import { createInputElement } from './create-input-element';
+
+describe('Barista app utils', () => {
```
lukasholzer: I like 🔥

apps/barista/src/components/live-example/live-example.html
```diff
@@ -4,12 +4,16 @@
   <div #demoPlaceholder></div>
 </div>
 
-<div class="ba-live-example-actions">
+<div
+  class="ba-live-example-actions"
+  [class.ba-live-example-source-hidden]="_sourceHidden"
```
thomaspink: Can we just `ngIf` here instead of applying a hidden class?

apps/barista/src/components/live-example/live-example.ts
```diff
@@ -111,33 +132,89 @@ export class BaLiveExample implements OnDestroy {
...
+  _activeTab: BaSourceType;
+
+  /** @internal Whether the source code is hidden. */
+  _sourceHidden = false;
+
+  /** @internal Whether the source code is expanded to its max height. */
+  _isExpanded = false;
+
+  /** @internal Whether the source code has been copied. */
+  _copied$: BehaviorSubject<boolean> = new BehaviorSubject(false);
```
thomaspink: In your case having an observable is not the necessary as it is just reflecting a state without any transformation or additional stuff. Having just a property is fine.

apps/barista/src/components/live-example/live-example.ts
```diff
@@ -111,33 +132,89 @@ export class BaLiveExample implements OnDestroy {
...
+        case 'ts':
+          copyText = this._classSource;
+          break;
+        case 'scss':
+          // TODO: copyText = this._scssSource;
+          break;
+      }
+
+      if (copyText) {
+        const textarea = createInputElement(copyText, 'textarea');
```
thomaspink: We have now multiple copy to clipboard occurrences (headline link). Can we combine these copy to clipboard logic into a single function or service

apps/barista/src/components/live-example/live-example.ts
thomaspink: @lukasholzer Not sure what you mean with breaking out the view encapsulation via the `:host` attribute. `:host` is there for accessing the host element, not for breaking out the encapsulation. I guess you mean `ng-deep`?

apps/barista/src/components/live-example/live-example.html
lara-aigmueller: I use `ngIf` for the source itself in line 65, but here I need a class for some minor styling updates when the source-block is hidden.

apps/barista/src/components/live-example/live-example.ts
lara-aigmueller: :ng-deep it was – thanks.

apps/barista/src/components/live-example/live-example.ts
```diff
@@ -153,4 +221,34 @@ export class BaLiveExample implements OnDestroy {
...
+    );
+  }
+
+  /**
+   * Transforms given code by replacing empty lines,
+   * and wrapping each line in a span element.
+   */
+  private _enhanceCode(code: string, type: BaSourceType): string {
+    // Remove empty lines at the start of the source
+    let transformedCode = code.replace(/^(\s*\r?\n)*/g, '');
```
lukasholzer: NIT: For those regexes, I would add a comment with an example regex101 to showcase the regex

Issue #251, opened 2019-12-02
Table: Favorite column


ffriedl89:
# Feature Request

## Summary

I want to have a column that handles marking rows as favorites. The column should contain a star icon only without fill when unselected and a gray stroke. When selected the star gets filled with a shade of yellow. UX can give information about specific color values.

## Feature Description

Currently we lack a workflow that allows a user to mark an entry in a table as a favorite. We need to consider sorting behaviour. 
I'd suggest extending the column from `DtSimpleColumnBase`. 
Secondly add a selection model for the favorites to the DtTableDatasource with multiSelect enabled and an empty array for initial selection. 

```
new SelectionModel<T>(true,[]);
``` 
see https://material.angular.io/components/table/overview#selection for inspiration
Things to consider is how to set the selection as part of the datasource.

PR #370, opened 2019-12-18
Feat/table favorite column

yngrdyn:
<hr>

Closes [#251](https://github.com/dynatrace-oss/barista/issues/251)

#### Type of PR

#### Checklist

### DIFF
--- a/components/table/README.md
+++ b/components/table/README.md
@@ -79,9 +79,9 @@ problem indicators.
 
 #### Variants
 
-Currently there are two predefined versions of the `dtSimpleColumn` exposed:
-`dt-simple-number-column` and `dt-simple-text-column`. There are only small
-differences between the number and text column:
+Currently there are three predefined versions of the `dtSimpleColumn` exposed:
+`dt-simple-number-column`, `dt-simple-text-column` and `dt-favorite-column`.
+There are only small differences between the number and text column:
 
 ##### dt-simple-text-column
 
@@ -95,6 +95,18 @@ differences between the number and text column:
 - Column alignment is set to `number` -> right
 - When no formatter is given, the `dtCount` formatter will automatically be
   applied, e.g. 1000 -> 1k
+
+##### dt-favorite-column
+
+- Column alignment is set to `icon` -> center
+- When sorting this column, the sort direction will start descending, e.g. ★ ->
+  ☆
+
+Default example of a favorite column:
+<ba-live-example name="DtExampleTableFavoriteColumn" fullwidth="true"></ba-live-example>
+
+When no label is given, nothing will be rendered inside the header-cell:
+<ba-live-example name="DtExampleTableFavoriteColumnNoHeader" fullwidth="true"></ba-live-example>
 
 ## Custom columns for advanced use cases
 
--- a/components/table/src/simple-columns/favorite-column.html
+++ b/components/table/src/simple-columns/favorite-column.html
@@ -1,0 +1,31 @@
+<ng-container
+  [dtColumnDef]="name"
+  dtColumnAlign="icon"
+  [dtColumnProportion]="proportion"
+>
+  <ng-container *dtHeaderCellDef>
+    <dt-header-cell [dt-sort-header]="sortable" start="desc"
+      >{{ label }}</dt-header-cell
+    >
+  </ng-container>
+  <dt-cell
+    *dtCellDef="let data"
+    [dtIndicator]="_getIndicator(data)"
+    [dtIndicatorColor]="_getIndicator(data)"
+    [attr.title]="_isFavorite(data)"
+  >
+    <ng-container
+      *ngTemplateOutlet="favoriteTemplate; context:{ favorite: _isFavorite(data), data: data }"
+    ></ng-container>
+  </dt-cell>
+</ng-container>
+
+<ng-template #favoriteTemplate let-favorite="favorite" let-data="data">
+  <button dt-icon-button variant="nested" (click)="_toggleFavorite(data)">
+    <dt-icon
+      [name]="favorite ? 'favorite' : 'star-empty'"
+      [class.dt-favorite-column-filled-star]="favorite"
+      [class.dt-favorite-column-empty-star]="!favorite"
+    ></dt-icon>
+  </button>
+</ng-template>
--- a/components/table/src/simple-columns/favorite-column.scss
+++ b/components/table/src/simple-columns/favorite-column.scss
@@ -1,0 +1,10 @@
+@import './simple-column.scss';
+@import '../../../core/src/style/colors';
+
+::ng-deep .dt-favorite-column-empty-star svg {
+  fill: $gray-400 !important;
+}
+
+::ng-deep .dt-favorite-column-filled-star svg {
+  fill: $yellow-500 !important;
+}
--- a/components/table/src/simple-columns/favorite-column.ts
+++ b/components/table/src/simple-columns/favorite-column.ts
@@ -1,0 +1,67 @@
+/**
+ * @license
+ * Copyright 2020 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  ChangeDetectionStrategy,
+  Component,
+  EventEmitter,
+  Optional,
+  Output,
+  ViewEncapsulation,
+} from '@angular/core';
+import { DtTable } from '../table';
+import { DtSimpleColumnBase } from './simple-column-base';
+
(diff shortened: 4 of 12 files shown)


components/table/src/simple-columns/favorite-column.scss
```diff
@@ -0,0 +1,12 @@
+@import './simple-column.scss';
+@import '../../../core/src/style/colors';
+
+.dt-icon {
+  width: 22px;
+  cursor: pointer;
+  fill: $gray-300;
+}
+
```
tomheller: I'm not sure if this will need a hover effect as well, maybe we will need some input from UX here?

libs/examples/src/table/table-favorite-column-example/table-favorite-column-example.html
```diff
@@ -0,0 +1,26 @@
+<dt-table [dataSource]="dataSource" dtSort #sortable>
```
tomheller: These examples show up in our documentation as live examples and are usually copy and pasted as a guide to get started somewhere. 
Maybe you could split this example up into multiple ones, which show the different options for the `favorite column`?

yngrdyn: I divided in two examples and added as a reference in the `README` file

components/table/src/simple-columns/favorite-column.scss
yngrdyn: You were right. I added the hover effect

components/table/README.md
```diff
@@ -96,6 +96,18 @@ differences between the number and text column:
...
+
+- Column alignment is set to `icon` -> center
+- When sorting this column, the sort direction will start descending, e.g. ★ ->
+  ☆
+
+<ba-live-example name="DtExampleTableFavoriteColumn" fullwidth="true"></ba-live-example>
+
+- When no label is given, nothing will be rendered inside the header-cell
+
+<ba-live-example name="DtExampleTableFavoriteColumnNoHeader" fullwidth="true"></ba-live-example>
```
tomheller: nitpick, maybe we can change this to have not a single, lonely list item to: 
```md

- Column alignment is set to `icon` -> center
- When sorting this column, the sort direction will start descending, e.g. ★ ->
  ☆

Default example of a favorite column: 
<ba-live-example name="DtExampleTableFavoriteColumn" fullwidth="true"></ba-live-example>

When no label is given, nothing will be rendered inside the header-cell:
<ba-live-example name="DtExampleTableFavoriteColumnNoHeader" fullwidth="true"></ba-live-example>
```

components/table/src/simple-columns/favorite-column.ts
```diff
@@ -0,0 +1,63 @@
...
+  constructor(@Optional() table: DtTable<T>) {
+    super(table);
+  }
+
+  /**
+   * @internal Get data either returns a data access with the given name or calls the
+   * displayAccessor function to get the simpleData for display.
+   */
+  // tslint:disable-next-line: no-any
+  _getData(data: T): any {
```
thomaspink: Could use a property instead of this function and update it only when needed. because right now, this function is called 3 times in the template on every change detection run. This means it gets called a lot even if the result does not change.

thomaspink:
Please also rebase and squash all commits into one when the review has been done as is marked as `pr: lgtm`.

components/table/src/simple-columns/favorite-column.ts
yngrdyn: Hi Thomas, I improved the template a bit but I tried to apply completely your suggestion but it seems it does not cover fully our case.
The main issue is that we need to update the reference in the template when clicking on the icon but if we do without _`getData` I always get the old reference (even by triggering the change detection manually does not solve this issue).
For example if the first row has favorite=true and the second=false, I always get the first row as it's false and the second as it's true.
How can I get the proper value without accessing the row?
Do you have any suggestion to solve this issue?
I tried to look also for other usages but in all the cases we always accessing through _getData (also more than one time as in our case)

thomaspink: @yngrdyn thanks for clarification. This makes now totally sense. LGTM :)

Only thing is, can me rename `getData` because as far as i tries to always return a `string` (title) for displaying it in the template. Maybe we could then also change the return type too?

components/table/src/simple-columns/favorite-column.html
```diff
@@ -0,0 +1,30 @@
...
+    ></ng-container>
+  </dt-cell>
+</ng-container>
+
+<ng-template #favoriteTemplate let-favorite="favorite" let-data="data">
+  <dt-icon
+    [name]="favorite ? 'favorite' : 'star-empty'"
+    [class.dt-favorite-column-filled-star]="favorite"
+    [class.dt-favorite-column-empty-star]="!favorite"
+    (click)="onChange(data)"
```
thomaspink: clicking on non button/anchor elements i kinda bad for a11y. Can we change it to an icon button?

components/table/src/simple-columns/favorite-column.ts
```diff
@@ -0,0 +1,65 @@
...
+   * Deliberatley set to Default because the embedded view that gets created for the
+   * dtColumDef can't handle onPush and results in ChangeAfterChecked error.
+   */
+  changeDetection: ChangeDetectionStrategy.Default,
+  providers: [{ provide: DtSimpleColumnBase, useExisting: DtFavoriteColumn }],
+})
+export class DtFavoriteColumn<T> extends DtSimpleColumnBase<T> {
+  @Output() readonly change = new EventEmitter<T>();
+
+  favorite: boolean = false;
```
thomaspink: * Should it be part of the public api - or can it be `internal`?
* Missing documentation. Maybe something like `Whether the column is marked as a favourite`.

components/table/src/simple-columns/favorite-column.ts
```diff
@@ -0,0 +1,65 @@
...
+   * displayAccessor function to get the simpleData for display.
+   */
+  // tslint:disable-next-line: no-any
+  _getData(data: T): any {
+    return this.displayAccessor
+      ? this.displayAccessor(data, this.name)
+      : (data as any)[this.name]; // tslint:disable-line:no-any
+  }
+
+  onChange(data: T): void {
```
thomaspink: * We try to avoid generic method names like these if possible. Maybe something like `toggleFavourite` or if it does not work `handleClick` is also possible in certain cases. See: https://barista.lab.dynatrace.org/components/coding-standards#methods
* Please also add a doc comment that further describes this method.
* I think this method does not need to be public. Can we make it internal?

components/table/src/simple-columns/favorite-column.ts
```diff
@@ -0,0 +1,65 @@
...
+  encapsulation: ViewEncapsulation.Emulated,
+  /*
+   * Deliberatley set to Default because the embedded view that gets created for the
+   * dtColumDef can't handle onPush and results in ChangeAfterChecked error.
+   */
+  changeDetection: ChangeDetectionStrategy.Default,
+  providers: [{ provide: DtSimpleColumnBase, useExisting: DtFavoriteColumn }],
+})
+export class DtFavoriteColumn<T> extends DtSimpleColumnBase<T> {
+  @Output() readonly change = new EventEmitter<T>();
```
thomaspink: * We try to avoid (generic) outputs that are like native events (such like `change`) on components (except the ones that replace native elements, like checkbox).
* As it is public api, please provide a doc comment.

thomaspink:
@yngrdyn I have added some minor feedback. After that i guess we are fine to merge it. thanks again for the work you have done.

components/table/src/simple-columns/favorite-column.ts
yngrdyn: Property was removed

components/table/src/simple-columns/favorite-column.ts
yngrdyn: Name was changed to `_toggleFavorite, comment was added describing the method and also was set as internal

components/table/src/simple-columns/favorite-column.ts
yngrdyn: Event was changed to `favoriteToggled` and the doc comment was added

components/table/src/simple-columns/favorite-column.ts
yngrdyn: I renamed it to `_isFavorite` and now returns a boolean

components/table/src/simple-columns/favorite-column.html
yngrdyn: I changed to a button but please check the scss to know if it's a better way to do the coloring of the star

components/table/src/simple-columns/favorite-column.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  constructor(@Optional() table: DtTable<T>) {
+    super(table);
+  }
+
+  /**
+   * @internal is Favorite returns whether the column specified by name in the datasource
+   * holds true or false.
+   */
+  // tslint:disable-next-line: no-any
```
tomheller: Nitpick: There is no `any` anymore, so we could remove the tslint disable.
```suggestion
```

yngrdyn: Thank you! I overlooked this

thomaspink:
I will take care of the rebasing after the review has been completed :)

ffriedl89:
We need to update the license header in this PR but we can do it after merging this. Thx a lot for the contribution 😄 very nicely done!

PR #385, opened 2020-01-07
Refactor/cleanup old dependencies

lukasholzer:
Cleanup legacy dependencies that are only bloating our install time on the ci.

We are not using gulp anymore so everything that is related to gulp can be deleted and some more dependencies.

## Dependency list

- **q** A promise library that is absolutely not needed!
- **puppeteer** We don't need puppeteer we are using testcafe
- **tsconfig-paths** Was needed for gulp to compile with typescript path aliases
- **undertaker-forward-reference** We don't use gulp anymore
- **watch** Wwe don't need a watcher anymore since we are using angular builders
- **web-animations-js** Is only needed if the [AnimationBuilder](https://angular.io/api/animations/AnimationBuilder) of angular is used but we don't use it.
- **gulp** We are using angular builders now
- **gulp-clean** We don't use gulp anymore
- **gulp-util** We don't use gulp anymore
- **jasmine-reporter** We use jest and jest reporters
- **resolve-bin** Was used by gulp
- **cross-spawn** Was used by gulp
- **copyfiles** Just not used anymore
- **rxjs-tslint** Just not used


thomaspink:
Gulp code has not been removed yet. We should either remove also all the gulp code or block this one until it is removed

tomheller:
I agree with @thomaspink here, there is still some gulp code in the codebase, which is currently being removed by others. I would also suggest blocking this one, until we have removed the code and are absolutely certain, that removing the dependencies does not break anything.

lukasholzer:
@tomheller @thomaspink  The gulp code is not working any more (as a point of reference I would leave the code till everything is in the builders) but this code cannot work anymore. Paths are wrong everything there is for the old workspace setup. So the dependencies only slow down our CI times in case we have to install a lot more.

We can remove the gulp code immediately and have it as a reference in the git history but I would do this in any case in a separate PR.

This PR should only speed up our CI times.

thomaspink:
Blocked - as we still got code in place that depends on gulp (e.g. breaking-changes)
> The gulp code is not working any more (as a point of reference I would leave the code till everything is in the builders) but this code cannot work anymore.
We need to first translate gulp code to node so we can safely remove gulp.

PR #400, opened 2020-01-08
ci: Speedup ci with parallel and xlarge resource.

lukasholzer:
Try to speed up our ci resources with parallelization and beeing wasteful with our credits 💯


ffriedl89:
Should we retrigger this? :D 💰

PR #416, opened 2020-01-10
build: Fix typings in secondary entrypoints

ffriedl89:
Works around the [ng-packagr issue ](https://github.com/ng-packagr/ng-packagr/issues/1239)
for secondary entrypoints


ffriedl89:
build should pass after #418 is merged

PR #428, opened 2020-01-14
feat(theming): Providing theming mixins in bundle

ffriedl89:
Provide theming mixins as part of the output, they are already heavily used in multiple apps to get palette colors. Expose them as part of the public api


tomheller:
Change is duplicated in the #431, will close this one.

PR #431, opened 2020-01-15
Rc findings

ffriedl89:
Changes relevant to findings in the rc.0 release

### DIFF
--- a/components/filter-field/src/filter-field-default-data-source.ts
+++ b/components/filter-field/src/filter-field-default-data-source.ts
@@ -30,31 +30,51 @@ import {
   dtRangeDef,
 } from './types';
 
+/** Shape of a base object to be usable as a option for various autocomplete option types. */
+export type DtFilterFieldDefaultDataSourceOptionBase = {
+  name: string;
+  id?: any;
+};
+
 /** Shape of an object to be usable as a option in an autocomplete */
-export type DtFilterFieldDefaultDataSourceOption = { name: string } | string;
+export type DtFilterFieldDefaultDataSourceOption =
+  | DtFilterFieldDefaultDataSourceOptionBase
+  | (DtFilterFieldDefaultDataSourceOptionBase &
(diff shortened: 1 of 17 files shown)


components/filter-field/src/filter-field-default-data-source.ts
```diff
@@ -31,12 +31,15 @@ import {
 } from './types';
 
 /** Shape of an object to be usable as a option in an autocomplete */
-export type DtFilterFieldDefaultDataSourceOption = { name: string } | string;
+export type DtFilterFieldDefaultDataSourceOption =
+  | { name: string; id?: string }
```
tomheller: This looks fine to me, a comment about why we are doing this would be cool.

ffriedl89: I am gonna go with any since the id might be an enum / number / whatever

libs/examples/src/filter-field/filter-field-distinct-example/filter-field-distinct-example.ts
```diff
@@ -29,6 +32,7 @@ export class DtExampleFilterFieldDistinct {
         name: 'AUT',
         autocomplete: ['Linz', 'Vienna', 'Graz'],
         distinct: true,
+        id: 'hello',
```
tomheller: remove dummy test info.

libs/examples/src/filter-field/filter-field-default-example/filter-field-default-example.ts
```diff
@@ -35,7 +38,7 @@ export class DtExampleFilterFieldDefault {
           'San Francisco',
           'Los Angeles',
           'New York',
-          { name: 'Custom', suggestions: [], validators: [] },
+          // { name: 'Custom', suggestions: [], validators: [] },
```
lukasholzer: can we remove this comment?

PR #437, opened 2020-01-16
fix(filter-field): Fixes an issue with inconsistent clearAll functionality

### LINKED ISSUE #435: Filter field: Clear all button requires 2 clicks
# Bug Report

## Expected Behavior

"Clear all" should clear all (editable) fields on a single click

## Current Behavior

I have to click "Clear all" twice to remove the fields
(In some cases the button never has a (visual) effect no matter how often I click it until I scroll, but I can't reproduce this consistantly, so hopefully it's the same problem)

## Steps to Reproduce

https://barista.dynatrace.com/components/filter-field/#readonly-non-deletable--non-editable-tags

1. Add an editable field to the filter-field
2. Click "Clear all"

tomheller:
The clearAll button rendering depended on the filterfield not being
focussed. When the clearAll button was rendered and was clicked, the
cdkFocusMonitor kicked in, set the focus and removed the clearAll button
again. As a result the clearAll button lost its focus, the document
focus was set back to the body and the clearAll was rendered again. The
click event on the clearAll was never fired.
We no longer remove the clearAll from the DOM, but hide it via css
classes. This makes it not lose focus, and results in a more consistent
behavior.

Fixes #435

<hr>

  often helpful).
  guidelines.

#### Checklist

### DIFF
--- a/apps/components-e2e/src/components/filter-field/filter-field.e2e.ts
+++ b/apps/components-e2e/src/components/filter-field/filter-field.e2e.ts
@@ -26,6 +26,7 @@ import {
   focusFilterFieldInput,
   getFilterfieldTags,
   tagOverlay,
+  setupSecondTestScenario,
 } from './filter-field.po';
 import { Selector } from 'testcafe';
 import { waitForAngular } from '../../utils';
@@ -255,4 +256,31 @@ test('should show the overlay on a tag because the tag value is ellipsed', async
     .hover(filterTags)
     .expect(tagOverlay.exists)
     .ok();
+});
+
+test('should remove all removable filters when clicking the clear-all button', async (testController: TestController) => {
+  // Setup the second datasource.
+  await testController
+    .click(setupSecondTestScenario)
+    // Wait for the filterfield to catch up.
+    .wait(500);
+
+  // Manually set an option, because this seems to break the
+  // clear all change detection.
+  await clickOption(1);
+  await clickOption(1)
+    .expect(filterTags.count)
+    .eql(2)
+    // Click somewhere outside so the clear-all button appears
+    .click(Selector('.outside'))
+    // Wait for the filterfield to catch up.
+    .wait(500)
+    .expect(clearAll.exists)
+    .ok()
+    // Click the clear all-button, the created filter should be removed
+    .click(clearAll)
+    // Wait for the filterfield to catch up.
+    .wait(500)
+    .expect(filterTags.count)
+    .eql(1);
 });
--- a/apps/components-e2e/src/components/filter-field/filter-field.ts
+++ b/apps/components-e2e/src/components/filter-field/filter-field.ts
@@ -17,15 +17,23 @@
 // tslint:disable no-lifecycle-call no-use-before-declare no-magic-numbers deprecation
 // tslint:disable no-any max-file-line-count no-unbound-method use-component-selector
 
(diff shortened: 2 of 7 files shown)


apps/components-e2e/src/components/filter-field/filter-field.e2e.ts
```diff
@@ -76,3 +77,30 @@ test('should remove all filters when clicking the clear-all button', async (test
...
+  // // Click somewhere outside so the clear-all button appears
+  await testController.click(Selector('.outside'));
+  // Wait for the filterfield to catch up.
+  await testController.wait(500);
+  await testController.expect(await clearAll.exists).ok();
+
+  // Click the clear all-button, the created filter should be removed
+  await testController.click(clearAll);
+  // Wait for the filterfield to catch up.
+  await testController.wait(500);
```
lukasholzer: is this related to animations?

Maybe we should consider using the `NoopAnimationsModule` instead of the `BrowseranimationsModule`?
@ffriedl89 Maybe this would reduce flakiness?

apps/components-e2e/src/components/filter-field/filter-field.ts
```diff
@@ -57,10 +58,75 @@ const TEST_DATA: DtFilterFieldDefaultDataSourceType = {
...
+
+  setupSecondTestScenario(): void {
+    this._dataSource = new DtFilterFieldDefaultDataSource(DATA[1]);
+    const linzFilter = [
+      DATA[1].autocomplete[0],
+      DATA[1].autocomplete[0].autocomplete![0],
+    ];
+
+    this._filterfield.filters = [linzFilter];
+    this._filterfield.currentTags.subscribe(() => {
```
lukasholzer: please unsubscribe in a `ngOnDestroy`

components/filter-field/src/filter-field.scss
```diff
@@ -110,6 +110,11 @@ button.dt-filter-field-clear-all-button {
   margin-left: 8px;
 }
 
+button.dt-filter-field-clear-all-button-hidden {
```
lukasholzer: why we have to hide it? Can we just `ngIf` it, please explain it to me :)

tomheller: As seen in the description of the PR, if we *ngIf the button, it will interfere with how the filterfield uses the focusmonitor to show the button or not. The button would be removed on mouseDown, would immediately loose focus (because it is being removed), and be shown again, without the click handler ever firing.

> The clearAll button rendering depended on the filterfield not being
focussed. When the clearAll button was rendered and was clicked, the
cdkFocusMonitor kicked in, set the focus and removed the clearAll button
again. As a result the clearAll button lost its focus, the document
focus was set back to the body and the clearAll was rendered again. The
click event on the clearAll was never fired.
We no longer remove the clearAll from the DOM, but hide it via css
classes. This makes it not lose focus, and results in a more consistent
behavior.

apps/components-e2e/src/components/filter-field/filter-field.e2e.ts
tomheller: It's not related to the animations module, but the way the filter field updates, it waits for the `zone` to get stable, before it will update it's state.

tomheller:
I have resolved all conflicts. PR should be ready now.

PR #454, opened 2020-01-20
feat(filter-field): Types

### LINKED ISSUE #507: Filter field: Default datasource types should be able to hold custom properties
# Bug Report

The current types that describe the data structure that need to be passed into the `DtFilterFieldDefaultDataSource` make it hard (nearly impossible) to describe custom properties that are added additionally.

## Expected Behavior

Custom data structures/properies on options for the default data source should be possible and the types should support it.

## Current Behavior

Currently you can set custom data/properties but the defined types on the default data source makes it hard to do so. A typesafe solution is not possible right now. The consumers have to workaround with `any` or `unkown`.

## Possible Solution

Add a Generic to the option type so the consumer can set a custom data structure via this generic.

## Steps to Reproduce

Apply any custom property to the options in the data structure and try to read out these property when the filter has changed (via the filterChanges event or using the filters property on the filter field)

## Context (Environment)

Some teams need to apply custom properties/data to the options and the current types make it hard to do so.

thomaspink:

The goal of this PR is to make the typings of the filter field more precise so it is easier for the consumer to work with the data it gets.

fixes #507 

#### Type of PR

#### Checklist

### DIFF
--- a/components/filter-field/src/filter-field-data-source.ts
+++ b/components/filter-field/src/filter-field-data-source.ts
@@ -17,88 +17,76 @@
 import { Observable } from 'rxjs';
 import { DtNodeDef } from './types';
 
-export abstract class DtFilterFieldDataSource {
+export abstract class DtFilterFieldDataSource<T> {
   /**
    * Used by the DtFilterFieldControl. Called when it connects to the data source.
    * Should return a stream of data that will be transformed, filtered and
    * displayed by the DtFilterField and the DtFilterFieldControl.
    */
-  abstract connect(): Observable<DtNodeDef | null>;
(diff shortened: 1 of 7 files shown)


ffriedl89:
LGTM, but ... why does this not break linting - no-any should be enabled  ¯\_(ツ)_/¯

ffriedl89:
Can we close this stale PR?

thomaspink:
Blocked. Should be finalised for version 7

tomheller:
@thomaspink As the release of 7.0.0 is around the corner, what does thes PR need to be finalized 
(as you stated in the last comment). 
Would you be able to prepare this PR for the release?

thomaspink:
Will make it ready for the release in the next days

PR #455, opened 2020-01-20
feat(ng-update): Added test for .spec file

rowa-audil:
Adds a test to migration, to check root import migration on specfiles.

### DIFF
--- a/components/schematics/src/ng-update/tests/root-imports.spec.ts
+++ b/components/schematics/src/ng-update/tests/root-imports.spec.ts
@@ -65,4 +65,33 @@ describe('v5 dynatrace angular components imports', () => {
 
     removeTempDir();
   });
+
+  it('should migrate root imports in .spec files', async () => {
+    const {
+      runFixers,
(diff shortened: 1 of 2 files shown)


components/schematics/src/ng-update/tests/root-imports.spec.ts
```diff
@@ -65,4 +65,33 @@ describe('v5 dynatrace angular components imports', () => {
...
+  it('should migrate root imports in .spec files', async () => {
+    const {
+      runFixers,
+      appTree,
+      writeFile,
+      removeTempDir,
+    } = await createTestCaseSetup('update-5.0.0', migrationCollection, []);
+
+    writeFile(
+      'projects/lib-testing/src/main.spec.ts',
```
lukasholzer: please check if the tsconfig inside the testing case setup is configured to work with the spec files otherwise the ts program that will be created cannot work.

Please check how material is testing this otherwise it is kind of useless.

PR #462, opened 2020-01-21
fix(autocomplete, filter-field): Fixes an issue with overlay positioning.

### LINKED ISSUE #451: Filter Field: Autocomplete appears above page header
# Bug Report

With an open autocomplete or range, the opened overlay appears above the sticky page-header (z-index problem). 

## Expected Behavior

I'm not entirely sure what should happen. Either the autocomplete or range box should close when scrolling or the z-index should be fixed. 

## Current Behavior

Autocomplete or range box stay open and overlay other components.

### Used Versions:

- **angular**: 8.2.12
- **@angular/cdk**: 8.2.3
- **@dynatrace/barista-components**: 4.15.1

tomheller:
Switched the autocomplete and filterfield overlay to use the
DtFlexibleConnectedPositionStrategy instead of the CDK
FlexibleConnectedPositionStrategy. The Dt version can deal with the
clusters page-ribbon component and in general deals with better with
overlaying elements.

Fixes #451

@thomaspink @ffriedl89 As we have to change the constructor signature here, should we consider this a breaking change?

#### Checklist

### DIFF
--- a/components/filter-field/src/filter-field-range/filter-field-range-trigger.ts
+++ b/components/filter-field/src/filter-field-range/filter-field-range-trigger.ts
@@ -17,11 +17,13 @@
 import { coerceBooleanProperty } from '@angular/cdk/coercion';
 import { ESCAPE, UP_ARROW } from '@angular/cdk/keycodes';
 import {
-  FlexibleConnectedPositionStrategy,
   Overlay,
   OverlayConfig,
   OverlayRef,
   PositionStrategy,
+  ViewportRuler,
+  OverlayContainer,
+  FlexibleConnectedPositionStrategy,
 } from '@angular/cdk/overlay';
 import {
   ChangeDetectorRef,
@@ -31,6 +33,8 @@ import {
   NgZone,
   OnDestroy,
   Renderer2,
+  Optional,
+  Inject,
 } from '@angular/core';
 import {
   EMPTY,
@@ -43,9 +47,14 @@ import {
 } from 'rxjs';
 import { filter, map, take } from 'rxjs/operators';
 
-import { readKeyCode } from '@dynatrace/barista-components/core';
+import {
+  readKeyCode,
+  DtFlexibleConnectedPositionStrategy,
+} from '@dynatrace/barista-components/core';
 
 import { DtFilterFieldRange } from './filter-field-range';
+import { Platform } from '@angular/cdk/platform';
+import { DOCUMENT } from '@angular/common';
 
 @Directive({
   selector: `input[dtFilterFieldRange]`,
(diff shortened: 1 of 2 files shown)


thomaspink:
@tomheller in the past we handled it as a breaking change (unless the additional params are optional)

tomheller:
@thomaspink What do you think, should we hold off on releasing this until we deliver v6.0.0 with Angular 9 support or schedule this for a minor version. The added constructor parameters are not optional, but are only in internal components like the autocomplete and the range component (part of the filter field). 

I don't have a strong opinion on how we should proceed.

thomaspink:
As it is only an internal component, I am fine with releasing it as non-breaking. I thing the benefit of releasing it earlier (via a patch release) outweighs the breaking part of it (as it is not used directly by the consumer).

tomheller:
@thomaspink @lukasholzer @ffriedl89 I have made the newly added Injections optional and have built a switch check into the creation of the new PositionStrategy. This should no longer be considered a breaking change. 
I have added breaking change comments for version 7 to remove the fallback to the cdk FlexibleConnectedPositionStrategy.

components/filter-field/src/filter-field-range/filter-field-range-trigger.ts
```diff
@@ -139,6 +153,11 @@ export class DtFilterFieldRangeTrigger implements OnDestroy {
     private _changeDetectorRef: ChangeDetectorRef,
     zone: NgZone,
     renderer: Renderer2,
+    @Optional() private _viewportRuler: ViewportRuler,
```
thomaspink: to make it non breaking pls make the prop itself also optional, not only via DI.
```suggestion
    @Optional() private _viewportRuler?: ViewportRuler,
```

Also annotate it as a breaking fo 7.0.0 to remove it again

components/filter-field/src/filter-field-range/filter-field-range-trigger.ts
```diff
@@ -139,6 +153,11 @@ export class DtFilterFieldRangeTrigger implements OnDestroy {
...
+    @Optional() private _platform: Platform,
```
thomaspink: Same here
```suggestion
    @Optional() private _platform?: Platform,
```

components/filter-field/src/filter-field-range/filter-field-range-trigger.ts
```diff
@@ -139,6 +153,11 @@ export class DtFilterFieldRangeTrigger implements OnDestroy {
...
+    @Optional() private _overlayContainer: OverlayContainer,
```
thomaspink:
```suggestion
    @Optional() private _overlayContainer?: OverlayContainer,
```

components/filter-field/src/filter-field-range/filter-field-range-trigger.ts
```diff
@@ -139,6 +153,11 @@ export class DtFilterFieldRangeTrigger implements OnDestroy {
...
+    // tslint:disable-next-line:no-any
+    @Optional() @Inject(DOCUMENT) private _document: any,
```
thomaspink:
```suggestion
    @Optional() @Inject(DOCUMENT) private _document?: any,
```

PR #464, opened 2020-01-21
chore: Update browserlist for caniuse-lite.

tomheller:
Updated the browserlist package from version 4.7.3 to 4.8.4.

Changelog for the browserlist updates can be found here:
https://github.com/browserslist/browserslist/blob/master/CHANGELOG.md#484

#### Type of PR


tomheller:
Conflicts in the package-lock.json were too severe, will reopen a new one.

PR #487, opened 2020-01-27
chore(filter-field): Fixes invalid example data.

thomaspink:

Fixes invalid example data.
We should add checks that prevents broken examples to be committed.


tomheller:
LGTM

PR #493, opened 2020-01-28
barista: Get UX snippet content from Strapi.

lara-aigmueller:

Added a Strapi snippet transformer that looks for `<ba-ux-snippet>` tags on content pages and replaces it with the snippet content coming from the CMS.

#### Type of PR

#### Checklist

### DIFF
--- a/libs/shared/barista-definitions/src/lib/barista-definitions.ts
+++ b/libs/shared/barista-definitions/src/lib/barista-definitions.ts
@@ -151,9 +151,10 @@ export interface BaStrapiPage extends BaStrapiBase {
 
 /** Strapi snippet */
 export interface BaStrapiSnippet extends BaStrapiBase {
-  slotID: string;
+  slotId: string;
   title: string;
   content: string;
+  public: boolean;
 }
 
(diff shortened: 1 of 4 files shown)


libs/shared/barista-definitions/src/lib/barista-definitions.ts
```diff
@@ -145,9 +145,10 @@ export interface BaStrapiPage extends BaStrapiBase {
 
 /** Strapi snippet */
 export interface BaStrapiSnippet extends BaStrapiBase {
-  slotID: string;
+  slotId: string;
```
tomheller: why do we call it `slot` once, and then `snippet` another time?

lara-aigmueller: In the beginning we talked about slots, now about snippets. I'll think about renaming it.

lara-aigmueller: I don't know. Naming things. I was thinking about having slots within the content where we then can place text/content snippets...

PR #499, opened 2020-01-28
feat(table): Expose currently rendered data from the datasource.

### LINKED ISSUE #497: Expose DtTableDataSource's _renderData
# Feature Request

## Summary

In order to lazy-load paginated table data, I need to know which rows are on the current page. `DtTableDataSource`'s api doesn't currently support that.

This information is currently hidden in the private field `_renderData`.

tomheller:

In order to give programmatic access to the data that is currently being rendered in the table, we should expose the renderedData BehaviourSubject.

Fixes #497

The question is if we want to expose the renderedData BehaviourSubject directly, or if we shoud introduce an additional member subject. I'm afraid that if we expose the Subject directly used in the table data source, that this could lead to consumers nexting the data in there themselves and therefore removing control from the DtTableDataSource.

#### Type of PR
Feature

#### Checklist

### DIFF
--- a/components/table/src/table-data-source.ts
+++ b/components/table/src/table-data-source.ts
@@ -113,6 +113,9 @@ export class DtTableDataSource<T> extends DataSource<T> {
   private _renderChangesSubscription = Subscription.EMPTY;
   private _searchChangeSubscription = Subscription.EMPTY;
 
+  /** Public stream emitting render data to the table */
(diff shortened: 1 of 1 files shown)


tomheller:
> Should we also add the change to the table's documentation?
We could but therefor we would need an entire section on the datasource, which is currently not there. 
Related #505

PR #520, opened 2020-02-03
feat(filter-field): Added highlighting of the currend input value to the options.

### LINKED ISSUE #203: Filter field: Add highlight to the options
# Feature Request

## Summary

I want to have the dt-highlight also on the options that are filtered when I type something.

## Feature Description

Since typing in the filter field filters the available options it makes sense to display the information to the user as well. Add the dt-highlight with the correct string to the options in the autocomplete

thomaspink:
fixes #203


tomheller:
Cool, thank you!
The changes are good, I'm just a bit skeptical about the visual style. As the currently selected option has the same background as the highlight. 

image

But thanks to the underline, you can at least guess what is happening.

PR #523, opened 2020-02-03
fix(legend): Fixes issue whether the legend item does not work with the new ivy engine.

### LINKED ISSUE #26: Legend: Fix incompatibility with Angular's ivy renderer
Since we upgraded to angular version 8.2.12, I tried to build app with enabled Ivy to see how much needs to be done.

I found it tries to compile Angular Components modules and breaks on `dt-legend` with message:
```
ERROR in Cannot combine @Input decorators with query decorators
```
Googling it a bit and it looks we cannot use `@Input()` and `@ContentChild()` on one class field at once. I found in codebase following place where it should be fixed:
legend-item.ts#56-58

Fixing this issue would help us further investigation on Ivy upgrade.

thomaspink:
fixes #26


tomheller:
Code looks good, please fix the typo in the commit message.

PR #536, opened 2020-02-04
Fix bounding calculation of tag-list component

rowa-audil:
Fixes / Improves evaluation of bounding properties of the tag-list wrapper and more-button


components/tag/src/tag-list/tag-list.ts
```diff
@@ -44,7 +44,7 @@ import { DtTag } from '../tag';
 import { DtTagAdd } from '../tag-add/tag-add';
 
 const DT_TAG_LIST_HEIGHT = 32;
-const DT_TAG_LIST_LAST_TAG_SPACING = 8;
+const DT_TAG_LIST_LAST_TAG_SPACING = 3;
```
ffriedl89: We have a 4-pixel grid, why do we have 3 here :D ?

components/tag/src/tag-list/tag-list.ts
```diff
@@ -166,13 +170,42 @@ export class DtTagList implements AfterContentInit, OnDestroy {
...
+      // Since the font-variant is set to tabular-nums we can use constants for the width
+      _hiddenTagCount < 10
+        ? this._setMoreBtnBoundingProperties(58)
+        : _hiddenTagCount < 100
+        ? this._setMoreBtnBoundingProperties(65.5)
+        : _hiddenTagCount < 1000
+        ? this._setMoreBtnBoundingProperties(73)
+        : _hiddenTagCount < 10000
+        ? this._setMoreBtnBoundingProperties(80.5)
+        : this._setMoreBtnBoundingProperties(88);
```
ffriedl89: Can we try to get rid of these magic numbers or describe why this is needed? Furthermore, .5 pixel values can't really be rendered so we should not set them. 

I've also tested the preview and I encountered some weird cases:
The first 2 were encountered after changing the width of the demo.
image
image
The next one is on initial load for me on Chrome 79
image

PR #549, opened 2020-02-05
feat(empty-state): Allow custom empty-states

### LINKED ISSUE #528: EmptyState directive
# Feature Request

## Summary

I want the be able to use a reusable empty state component inside the table. 

## Feature Description

I want to have a component that uses the dt-empty-state inside its own template. 

e.g.
```
<dt-table>
<my-custom-empty-state [dtEmptyState]></my-custom-empty-state>
</dt-table>
```
This way enables me to use a custom reusable empty state across multiple components

FYI: @myieye

myieye:
Enables the usage of custom empty states. This in turn makes empty states reusable.

Closes #528

#### Type of PR

Feature

#### Checklist

### DIFF
--- a/components/empty-state/README.md
+++ b/components/empty-state/README.md
@@ -36,3 +36,11 @@ This is mainly used in an empty entity view.
 ## Empty state in use
 
 <ba-ux-snippet name="empty-state-in-use"></ba-ux-snippet>
+
+## Custom empty state
+
+Custom/Reusable empty states are also supported using the `dtCustomEmptyState`
+directive. The custom empty state must both extend and provide itself as a
+`DtEmptyState`.
+
+<ba-live-example name="DtExampleCustomEmptyStateTable" background></ba-live-example>
--- a/components/empty-state/src/empty-state.ts
+++ b/components/empty-state/src/empty-state.ts
@@ -235,3 +235,12 @@ export class DtEmptyState
     }
   }
 }
+
+/**
+ * Marks a custom empty state component to be used within, for example, `<dt-table>`.
+ * The custom empty state must both extend and provide itself as a DtEmptyState.
+ */
+@Directive({
+  selector: '[dtCustomEmptyState]',
+})
+export class DtCustomEmptyState {}
--- a/components/table/src/table.spec.ts
+++ b/components/table/src/table.spec.ts
@@ -40,6 +40,7 @@ import {
 import {
   DtEmptyState,
   DtEmptyStateModule,
+  DtCustomEmptyState,
 } from '@dynatrace/barista-components/empty-state';
 import {
   DtLoadingDistractor,
@@ -75,6 +76,8 @@ describe('DtTable', () => {
         TestAppMultiExpandableTable,
         TestStickyHeader,
         TestIndicatorApp,
+        CustomEmptyState,
+        TestCustomEmptyStateApp,
       ],
     });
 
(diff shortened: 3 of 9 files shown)


components/table/src/table.spec.ts
```diff
@@ -827,3 +857,39 @@ class TestIndicatorApp {
...
+  template: `
+    <dt-table [dataSource]="dataSource">
+      <ng-container dtColumnDef="host" dtColumnAlign="text">
+        <dt-header-cell *dtHeaderCellDef>Host</dt-header-cell>
+        <dt-cell *dtCellDef="let row">{{ row.host }}</dt-cell>
+      </ng-container>
+
+      <dt-header-row *dtHeaderRowDef="['host']"></dt-header-row>
+      <dt-row *dtRowDef="let row; columns: ['host']"></dt-row>
+      <custom-empty-state dtEmptyState></custom-empty-state>
```
myieye: VSCode complains at this line that it can't find `custom-empty-state`. I'm not sure how to fix that, but everything seems to work...

tomheller:
Hi @myieye and thank you for contributing to the barista-components. 

Your pull request looks absolutely perfect 😍.
I will be here to help you to finalize the PR to get it merge ready. I will look into why the e2e tests are failing and why the `pull request preview` action does not pass. 

If you have questions at any time, please feel free to reach out.

tomheller:
It seems like the preview cannot be deployed from forks, so this is fine.
The failing e2e test is a flaky one that is already resolved in another pull request and can be ignored.
Everything else seems to be in order. Let's wait for the approval of at least 2 others and then I guess we will be able to merge this one.

thomaspink:
Hi Tim, thanks for your contribution.

components/table/src/table.spec.ts
```diff
@@ -827,3 +857,39 @@ class TestIndicatorApp {
...
+    },
+  ],
+  template: `
+    <dt-empty-state-item>
+      <dt-empty-state-item-title>No host</dt-empty-state-item-title>
+      Test message
+    </dt-empty-state-item>
+  `,
+})
+export class CustomEmptyState extends DtEmptyState {}
```
thomaspink: Is there a reason why you need to extend from the `DtEmptyState`? I think this is not necessary.

components/table/src/table.spec.ts
```diff
@@ -827,3 +857,39 @@ class TestIndicatorApp {
...
+})
+export class TestCustomEmptyStateApp {
+  dataSource: object[] = [{ host: 'host-1' }];
+}
+
+@Component({
+  selector: 'custom-empty-state',
+  providers: [
+    {
+      provide: DtEmptyState,
```
thomaspink: Same here, is there a reason why you need to provide it as the `DtEmptyState`? I think this is not needed

components/empty-state/src/empty-state.ts
```diff
@@ -235,3 +235,12 @@ export class DtEmptyState
...
 }
+
+/**
+ * Marks a custom empty state component to be used within, for example, `<dt-table>`.
+ * The custom empty state must both extend and provide itself as a DtEmptyState.
+ */
+@Directive({
+  selector: '[dtEmptyState]',
+})
+export class DtEmptyStateDirective {}
```
thomaspink: In the library we are not using the `Directive` suffix. 
See https://barista.dynatrace.com/components/coding-standards/#classes

Also pls rename the **class** and the **attribute selector** to something like `DtEmptyStateCustom`/`[dtEmptyStateCustom]` or `DtCustomEmptyState`/`[dtCustomEmptyState]` so it is clear that this is a different way of providing an empty-state and the consumer does not get confused with emty-state vs custom-emty-state.

components/table/src/table.spec.ts
myieye: [This QueryList](https://github.com/dynatrace-oss/barista/blob/0528c6610acd32b22f0b1d9a201f06a9fcc6dcf0/components/table/src/table.ts#L117) doesn't find it unless it provides a `DtEmptyState`.

myieye: @thomaspink I guess it would probably be better to adapt the table component to query for the directive as well?

thomaspink: @myieye i see your problem now. haven't checked the usage of the empty state in the table before but i now agree with you. we would either need the solution you already did, or adapt the table component to query differently as you have mentioned. both solutions are fine for me :)

components/table/src/table.spec.ts
thomaspink: This comment is obsolete: https://github.com/dynatrace-oss/barista/pull/549#discussion_r376270378

components/empty-state/README.md
```diff
@@ -35,4 +35,23 @@ This is mainly used in an empty entity view.
 
 ## Empty state in use
 
+<<<<<<< HEAD
 <ba-ux-snippet name="empty-state-in-use"></ba-ux-snippet>
```
myieye: I see that I totally botched a merge (or, rather, am not used to having to manually save each open file ;D), but where did the README content go? It seems to haven gotten sucked up by ux-snippet tags, but the content is no longer in the project anywhere?

tomheller: There are certain parts of the readme files, that are considered to be UX driven content. As the UX team should not be constrained to libraries release cycles or processes, a lot of contents in the readme have been replaced with `ba-ux-snippet` elements. They are being replaced during build times with more up-to-date content fetched from a CMS. 

The contents replacement with the `ba-ux-snippet` elements is correct.

PR #552, opened 2020-02-05
Feat e2e tests for overlay attribute propagation

rowa-audil:
Adds e2e tests to components that allow the propagation of attributes to its overlay

### DIFF
--- a/components/tag/src/tag-add/tag-add.ts
+++ b/components/tag/src/tag-add/tag-add.ts
@@ -134,10 +134,11 @@ export class DtTagAdd implements AfterViewInit, OnDestroy {
   constructor(
     private _changeDetectorRef: ChangeDetectorRef,
     private _zone: NgZone,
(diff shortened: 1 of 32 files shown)


components/tag/src/tag-add/tag-add.ts
```diff
@@ -127,6 +134,11 @@ export class DtTagAdd implements AfterViewInit, OnDestroy {
   constructor(
     private _changeDetectorRef: ChangeDetectorRef,
     private _zone: NgZone,
+    @Optional()
+    @Inject(DT_UI_TEST_CONFIG)
+    private _config?: DtUiTestConfiguration,
+    // @BREAKING-CHANGE: Optional will be removed with version 7.0.0
```
tomheller: Please use a JsDoc comment instead of a normal line comment
```suggestion
    /** @breaking-change: Optional will be removed with version 7.0.0 */
```

PR #554, opened 2020-02-05
Add sidenav example

rowa-audil:
Adds the sidnav component as an example

### DIFF
--- a/apps/dev/src/drawer/drawer-demo.component.scss
+++ b/apps/dev/src/drawer/drawer-demo.component.scss
@@ -5,17 +5,3 @@
 dt-drawer-container {
   margin-bottom: 20px;
 }
-
-.open-sidenav-button {
-  width: 44px;
-  height: 44px;
-  fill: #ffffff;
-  background: #191919;
-  border: 0;
-  outline: none;
-  cursor: pointer;
-
-  &:hover {
-    background: #353535;
-  }
-}
--- a/components/drawer/src/sidenav.scss
+++ b/components/drawer/src/sidenav.scss
@@ -4,17 +4,17 @@
   transform: translateX(-100%);
   will-change: transform;
   background-color: $gray-700;
+  color: #ffffff;
   min-width: 240px;
-  height: 100%;
+  height: fill-available;
   z-index: 100;
 }
 
 .dt-sidenav-header {
   display: flex;
-  max-height: 44px !important;
+  height: 44px;
   width: 240px;
   cursor: pointer;
-  color: #ffffff;
   background: #525252;
 }
 
@@ -40,9 +40,10 @@
 }
 
 .dt-sidenav-header-title {
-  max-height: 44px;
+  height: 44px;
   width: 196px;
   padding-left: 12px;
+  line-height: 44px;
   vertical-align: middle;
   text-align: left;
 }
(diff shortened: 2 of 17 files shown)


libs/examples/src/sidenav/sidenav-with-top-bar-navigation-example/sidenav-with-top-bar-navigation-example.html
```diff
@@ -0,0 +1,15 @@
+<dt-sidenav-container>
+  <dt-sidenav mode="side" opened #sidenav>
+    <dt-sidenav-header>
+      Sidenav title
+    </dt-sidenav-header>
+    Sidenav content
+  </dt-sidenav>
+  <dt-top-bar-navigation aria-label="Main navigation" class="nav">
+    <dt-top-bar-navigation-item *ngIf="!sidenav.opened" align="start">
+      <button class="open-sidenav-button" (click)="sidenav.open()">
```
lukasholzer: This should be implemented with the `dtTopBarAction` directive that provides the styling

```html
 <dt-top-bar-navigation-item align="start">
    <button (click)="sidenav.open()" dtTopBarAction>
      <dt-icon name="menu-hamburger"></dt-icon>
    </button>
  </dt-top-bar-navigation-item>
```

libs/examples/src/sidenav/sidenav-with-top-bar-navigation-example/sidenav-with-top-bar-navigation-example.scss
```diff
@@ -0,0 +1,13 @@
+.open-sidenav-button {
```
lukasholzer: should not be needed! Styling comes with the dtTopBarAction directive

libs/examples/src/sidenav/sidenav-default-example/sidenav-default-example.html
```diff
@@ -1 +1,9 @@
-<dt-sidenav></dt-sidenav>
+<dt-sidenav-container>
+  <dt-sidenav mode="side" opened>
+    <dt-sidenav-header>
+      Title
+    </dt-sidenav-header>
+    Sidenav content
```
lukasholzer: add something to reopen when it is closed

components/drawer/src/sidenav.scss
```diff
@@ -4,17 +4,17 @@
   transform: translateX(-100%);
   will-change: transform;
   background-color: $gray-700;
+  color: #ffffff;
   min-width: 240px;
-  height: 100%;
+  height: fill-available;
```
lukasholzer: Please use flexbox for that. That is no valid css property value for height

components/drawer/src/sidenav.scss
```diff
@@ -4,17 +4,17 @@
...
   z-index: 100;
 }
 
 .dt-sidenav-header {
   display: flex;
-  max-height: 44px !important;
+  height: 44px;
   width: 240px;
```
lukasholzer: why is the width set to 240? at least it should be a min width

apps/dev/src/drawer/drawer-demo.component.scss
```diff
@@ -5,3 +5,17 @@
 dt-drawer-container {
   margin-bottom: 20px;
 }
+
+.open-sidenav-button {
```
lukasholzer: this should not be part of the drawer

PR #563, opened 2020-02-06
Fix/aria label conflicts

### LINKED ISSUE #79: A11y: Aria labels input refactoring
We need to deprecate all the aria labels like `aria-label-close` that are violating our accessibility tests and rename the inputs to ariaLabelClose so when they are bound they are in the template.

### LINKED ISSUE #526: Fix a11y chart tests
# Bug Report

## Expected Behavior

The a11y tests should pass with our configuration.

## Current Behavior

the blacklisted examples don't pass.

```typescript
/** Blacklisted default examples */
const BLACKLIST = [
  'chart-selection-area-default-example',
  'chart-stream-example',
  'chart-loading-example',
];
```

## Steps to Reproduce

Go to the `apps/demos/src/barista-examples.a11y.ts` a11y tests file and uncomment the blacklist.
Then run `ng run demos:a11y --configuration=watch`

Use `--configurartion=ci` to execute it like on the ci

## Detailed Description

```
 ✖ chart-loading-example

   1) AssertionError: 1 violations found:
      1) ARIA attributes must conform to valid names
          "dt-loading-spinner"
      : expected false to be truthy

 ✖ chart-selection-area-default-example

   1) AssertionError: 1 violations found:
      1) ARIA attributes must conform to valid names
          "dt-chart-range"
      	"dt-chart-timestamp"
      : expected false to be truthy

 ✖ chart-stream-example

   1) AssertionError: 1 violations found:
      1) ARIA attributes must conform to valid names
          "dt-loading-spinner"
      : expected false to be truthy
```1

### PR DESCRIPTION
**Pull request by tomheller**

Specific inputs like `aria-label-button-close` are not allowed based on
the aria specification. ONly known aria attributes are allowed to be
used. We have renamed the inputs in question to their camelcased
version, but kept the old ones for compatibility for now.
With version 7.0. we sould remove the kebap-cased inputs.

Fixes #526
Fixes #79

#### Type of PR

Bugfix

#### Checklist

### CODE REVIEW

**Comment by tomheller** [comment]
There are still some errors in the tests, I will investigate.

File: components/chart/src/timestamp/timestamp.ts [hunk 1]
```diff
@@ -73,12 +73,55 @@ export class TimestampStateChangedEvent {
...
+  set ariaLabelClose(value: string) {
+    this._ariaLabelClose = value;
+  }
+  /** @internal Aria label value for the close button of the overlay */
+  _ariaLabelClose = ARIA_DEFAULT_CLOSE_LABEL;
 
-  /** Aria label for the selected moment */
+  /**
+   * Aria label for the close button in the overlay.
+   * @deprecated `aria-label-close` does not conform with accessibility standards.
```
lara-aigmueller: ARIA label for the selected button, instead of close button.

```suggestion
   * @deprecated `aria-label-selected` does not conform with accessibility standards.
```

components/pagination/src/pagination.ts
```diff
@@ -102,18 +103,131 @@ export class DtPagination implements OnInit {
...
+  get depAriaLabelEllipses(): string {
+    return this._ariaLabelEllipses;
+  }
+  set depAriaLabelEllipses(value: string) {
+    this._ariaLabelEllipses = value;
+  }
   /** Aria label for the ellipsis character. Defaults to "The next pages are ellipses" */
-  @Input('aria-label-ellipses') ariaLabelEllipses = ARIA_DEFAULT_ELLIPSES;
+  @Input()
+  get ariaLabelEllipses(): string {
```
lara-aigmueller: I think this should be named `ariaLabelEllipsis` instead (as you already wrote in the comments), as "ellipses" is plural.

documentation/linting.md
```diff
@@ -19,15 +19,15 @@ preventing those errors:
...
 | `dt-info-group-requires-icon`           | The `dt-info-group-icon` is a required child element of the `dt-info-group` and must be a direct child.                                                                                                                                             |
 | `dt-info-group-requires-title`          | The `dt-info-group-title` is a required child element of the `dt-info-group` and must be a direct child.                                                                                                                                            |
-| `dt-inline-editor-alt-text`             | An inline editor must provide alternative texts for the save and the cancel button using the following inputs: `aria-label-save` and `aria-label-cancel`.                                                                                           |
+| `dt-inline-editor-alt-text`             | An inline editor must provide alternative texts for the save and the cancel button using the following inputs: `ariaLabelCancel` and `aria-label-cancel`.                                                                                           |
```
lara-aigmueller:
```suggestion
| `dt-inline-editor-alt-text`             | An inline editor must provide alternative texts for the save and the cancel button using the following inputs: `ariaLabelSave` and `ariaLabelCancel`.                                                                                           |
```

libs/examples/src/empty-state/empty-state-multiple-items-example/empty-state-multiple-items-example.html
```diff
@@ -4,7 +4,7 @@
       <img src="/assets/cta-noagent.svg" alt="No agent" />
     </dt-empty-state-item-img>
 
-    <dt-empty-state-item-title>
+    <dt-empty-state-item-title aria-level="2">
```
lara-aigmueller: Do we also have to add `role="heading"` then?

I'm not sure if we should add the `aria-level` attribute here as we don't know where the code of this example is copied to. I think it can happen, that the code only gets copy-pasted without this value being changed (in case it needs to be changed because of the page hierarchy). Should we add a comment that this value needs to be changed if the empty-state-title is not a 2nd level headline?

libs/examples/src/table/table-empty-state-example/table-empty-state-example.html
```diff
@@ -23,16 +23,16 @@
...
         />
       </dt-empty-state-item-img>
 
-      <dt-empty-state-item-title>
+      <dt-empty-state-item-title aria-level="1">
```
lara-aigmueller: Same as I mentioned above. Do we also have to add `role="heading"`? And why is this one a level 1 heading?

libs/examples/src/empty-state/empty-state-multiple-items-example/empty-state-multiple-items-example.html
tomheller: The `role=heading`is set by the empty state title component. I think we should include it, otherwise the a11y tests for this component fails, and it is not clear that a level should be set at all.

components/chart/src/range/range.ts
```diff
@@ -84,20 +84,102 @@ export class RangeStateChangedEvent {
...
-  /** Aria label for the left handle */
+  /**
+   * Aria label for the left handle
+   * @deprecated `aria-label-left-handle` does not conform with accessibility standards.
+   * Please use `ariaLabelLeftHandle` input instead.
+   * @breaking-change Will be removed in version 7.0.0
+   */
   @Input('aria-label-left-handle')
-  ariaLabelLeftHandle = ARIA_DEFAULT_LEFT_HANDLE_LABEL;
+  get depAriaLabelLeftHandle(): string {
```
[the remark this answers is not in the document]
tomheller: They are needed, because the name of the input and the property change, so we need to keep consistency with the property and input and keep them working for at least 2 versions.

tomheller:
Worked into the changes requested, please review again @thomaspink @lukasholzer @lara-aigmueller

PR #572, opened 2020-02-07
barista(color-grid): Added colorlist input to the color-grid.

NinaKammerlander:

<hr>

Added another input to the color-grid component, to be able to use this component in the cms.

#### Checklist

### DIFF
--- a/apps/barista/src/components/color-grid/color-grid.ts
+++ b/apps/barista/src/components/color-grid/color-grid.ts
@@ -31,30 +31,40 @@ export class BaColorGrid implements AfterContentInit {
     this._colorInput = value;
   }
 
-  /** @internal name of the colorgroup */
-  _colorname: string;
+  /**
+   * list of all colors that should be displayed
+   * e.g. "yellow-500,blue-300,green-700"
+   */
+  @Input() colorList: string;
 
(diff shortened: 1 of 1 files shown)


apps/barista/src/components/color-grid/color-grid.ts
```diff
@@ -31,30 +31,44 @@ export class BaColorGrid implements AfterContentInit {
     this._colorInput = value;
   }
 
-  /** @internal name of the colorgroup */
-  _colorname: string;
+  @Input()
```
lara-aigmueller: Could you add a comment how this input should look like, i.e. how the colors have to be specified? :)

apps/barista/src/components/color-grid/color-grid.ts
```diff
@@ -31,30 +31,44 @@ export class BaColorGrid implements AfterContentInit {
...
+  get colorList(): string {
+    return this._colorList;
+  }
+  set colorList(value: string) {
+    this._colorList = value;
```
thomaspink: when you are not modifying the passed in value or calling any other methods, there is no need for a getter/setter.
A getter setter is only needed when you do additional stuff like the uppercasing in the color property.
Please change it to a normal property.

Issue #566, opened 2020-02-07
Angular 9 upgrade


ffriedl89:
# Feature Request

We want to upgrade our repository to Angular 9. 

For library authors, it's important to ship a view engine compiled library for now. 

see https://indepth.dev/the-angular-ivy-guide-for-library-authors/

ffriedl89:
This is also depending on the #537

PR #588, opened 2020-02-12
Update to angular 9

thomaspink:

**BREAKING CHANGE:** Update to Angular 9 requires us to update our `peerDependencies`.
Furthermore, our applications are compiled with Ivy. Only the component library is
compiled with the view engine to be backwards compatible.

Solves #566 

#### Type of PR

#### Checklist


.vscode/settings.json
```diff
@@ -2,5 +2,6 @@
   "editor.tabSize": 2,
   "files.eol": "\n",
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "explorer.compactFolders": false
```
tomheller: not really necessary to include within this PR.

angular.json
```diff
@@ -255,14 +269,19 @@
           },
           "configurations": {
             "production": {
+              "budgets": [
+                {
+                  "type": "anyComponentStyle",
+                  "maximumWarning": "6kb"
+                }
+              ],
```
tomheller: Is this necessary to include these budgets now, or is it an opt in. 
I'd opt for not introducing these budgets with this pull request and have a discussion on why we would need to add them in a separate issue.

apps/universal/src/app/kitchen-sink/kitchen-sink.html
```diff
@@ -301,6 +301,7 @@ <h1>Packets</h1>
   voluptua. At vero eos et accusam et justo duo dolores et ea rebum.
 </dt-expandable-text>
 
+<!--
```
tomheller: Please do not comment the event-chart here. We need to test it. If it does not work, please remove it and add a follow up issue.

components/autocomplete/src/autocomplete.spec.ts
```diff
@@ -1629,10 +1629,10 @@ class SimpleAutocomplete implements OnDestroy {
...
   closedSpy = jest.fn();
 
-  @ViewChild(DtAutocompleteTrigger, { static: false })
+  @ViewChild(DtAutocompleteTrigger)
   trigger: DtAutocompleteTrigger<any>;
-  @ViewChild(DtAutocomplete, { static: false }) panel: DtAutocomplete<any>;
-  @ViewChild(DtFormField, { static: false }) formField: DtFormField<any>;
+  @ViewChild(DtAutocomplete) panel: DtAutocomplete<any>;
+  @ViewChild(DtFormField) formField: DtFormField<any>;
   @ViewChildren(DtOption) options: QueryList<DtOption<any>>;
```
tomheller: Lots of public documentations missing. I know this is not part of the upgrade, just mentioning.

components/breadcrumbs/src/item/breadcrumbs-item.spec.ts
```diff
@@ -298,7 +298,7 @@ class TestBreadcrumbsItem {
   external = false;
 
   // tslint:disable-next-line:deprecation
```
tomheller: Do we still need the depraction disable?

components/event-chart/src/event-chart.ts
```diff
@@ -778,8 +780,12 @@ export class DtEventChart<T> implements AfterContentInit, OnInit, OnDestroy {
...
     const patternHost = this._document.createElement('div');
     // Move off-canvas so it does not overlap elements
-    patternHost.style =
-      'pointer-events: none; position: absolute; left: -999px;';
+
+    Object.defineProperty(patternHost, 'style', {
+      value: 'pointer-events: none; position: absolute; left: -999px;',
+    });
+
+    // patternHost.style = 'pointer-events: none; position: absolute; left: -999px;';
```
tomheller:
```suggestion
```

components/event-chart/src/event-chart.ts
```diff
@@ -474,6 +470,14 @@ export class DtEventChart<T> implements AfterContentInit, OnInit, OnDestroy {
     }
   }
 
+  /**
+   * Track by function for the renderEvents – speeds up performance
+   * and prevent deletion of mouseout
+   */
+  _renderEventTrackByFn(index: number, _item: RenderEvent<T>): number {
+    return index;
+  }
```
tomheller: You are tracking them by their array index, which does not work. If the values within the renderevents change, you will need to rerender them. Please check them based on either their originalIndex, or a combination of index and value.

components/expandable-section/src/expandable-section.ts
```diff
@@ -14,29 +14,27 @@
...
 
-@Component({
+@Directive({
   exportAs: 'dtExpandableSectionHeader',
   selector: 'dt-expandable-section-header',
-  template: '<ng-content></ng-content>',
-  encapsulation: ViewEncapsulation.Emulated,
-  preserveWhitespaces: false,
-  changeDetection: ChangeDetectionStrategy.OnPush,
 })
```
tomheller: Can we just change this from a component to a Directive? This immensely breaks behaviour, as there is no more `ng-content` outlet.

components/expandable-section/src/expandable-section.ts
```diff
@@ -55,38 +53,31 @@ export class DtExpandableSectionHeader {}
...
+    this._panel.expanded = coerceBooleanProperty(value);
     this._changeDetectorRef.markForCheck();
   }
 
-  /** Whether the expandable section is disabled. */
+  /** Whether the expandable section is expanded. */
   @Input()
   get disabled(): boolean {
-    return this._panel && this._panel.disabled;
+    return this._panel?.disabled;
```
tomheller: What does optional chaining return if _panel is undefined? Does it return `undefined` or `false`? I'm just curious.

components/table/src/cell.ts
```diff
@@ -32,17 +33,15 @@ import {
...
   ViewEncapsulation,
 } from '@angular/core';
-import { Subject, Subscription, merge } from 'rxjs';
-import { filter, startWith, switchMap, takeUntil } from 'rxjs/operators';
-
 import {
   DtIndicator,
-  _addCssClass,
   isDefined,
+  _addCssClass,
```
tomheller: Unnecessary sorting change.

components/table/src/cell.ts
```diff
@@ -32,17 +33,15 @@ import {
...
   _parseCssValue,
   _removeCssClass,
 } from '@dynatrace/barista-components/core';
-
+import { merge, Subject, Subscription } from 'rxjs';
+import { filter, startWith, switchMap, takeUntil } from 'rxjs/operators';
```
tomheller: Unnecessary sorting change.

components/table/src/cell.ts
```diff
@@ -185,24 +185,22 @@ export class DtCell implements AfterContentInit, OnDestroy {
   }
 
   ngAfterContentInit(): void {
-    this._indicators.changes.pipe(takeUntil(this._destroy)).subscribe(() => {
+    const indicatorChanges = this._dtIndicator
+      ? this._dtIndicator._stateChanges
+      : this._indicators.changes.pipe(
+          startWith(null),
+          filter(() => !!this._indicators.length),
+          switchMap(() => {
```
tomheller:
```typescript
switchMap(() => merge(
              ...this._indicators.map(indicator => indicator._stateChanges),
            )),
```

components/table/src/simple-columns/simple-column-base.ts
```diff
@@ -68,6 +67,7 @@ export type DtSimpleColumnFormatFunction = (
   displayValue: any, // tslint:disable-line:no-any
 ) => string | DtFormattedValue;
 
+@Directive()
```
tomheller: Does the selectorless directive work in this case, or will this handled by their sub-classes?

components/timeline-chart/src/timeline-chart-directives.ts
```diff
@@ -48,6 +48,7 @@ export class DtTimelineChartOverlayTitle {}
 export class DtTimelineChartOverlayText {}
 
 /** @internal Base class for sharing common logic between marker types. */
+@Directive()
```
tomheller: Does the selectorless directive work here?

angular.json
ffriedl89: I agree lets remove them for now and come up with a good number for us in another PR

components/table/src/cell.ts
lukasholzer: We have the tsconfig `"noUnusedParameters": true` in place that breaks all-time my build if there is an unused import. I don't fix them all time by hand –  I use the organize imports for that so sorry for the changes.

components/table/src/cell.ts
lukasholzer: We have the tsconfig `"noUnusedParameters": true` in place that breaks all-time my build if there is an unused import. I don't fix them all time by hand –  I use the organize imports for that so sorry for the changes.

angular.json
lukasholzer: This is done by the ng update schematic automatically because it is seen as go to. We should have this in place it is a warning nothing build breaking.

This is for the `demos app` and we should take this as measurement. because if we see  this warning we have done something wrong

lukasholzer: We have it already in place in all the other apps: `barista, components-e2e, dev`

components/autocomplete/src/autocomplete.spec.ts
lukasholzer: Yea but I won't document this in this pr. I have not even written this code.

components/timeline-chart/src/timeline-chart-directives.ts
lukasholzer: an abstract class that extends from a directive has to have the decorator according to the new Angular 9 this is correct.

components/table/src/simple-columns/simple-column-base.ts
lukasholzer: an abstract class that extends from a directive has to have the decorator according to the new Angular 9 this is correct.

components/breadcrumbs/src/item/breadcrumbs-item.spec.ts
lukasholzer: I think the deprecation of the breadcrumb should be handled in a separate PR:

```
/**
 * @deprecated Use {@link DtBreadcrumbsItem2} instead.
 * @breaking-change To be removed with 6.0.0
 */
```

components/expandable-section/src/expandable-section.ts
ffriedl89: Imo this should be fine - you can add any content inside a directive :)

.vscode/settings.json
lukasholzer: this file should be excluded from the git history. It only contains personal changes. I will remove my change.

components/expandable-section/src/expandable-section.ts
lukasholzer: undefined but I can cast it to a boolean if you want

libs/examples/src/filter-field/filter-field-unique-example/filter-field-unique-example.ts
```diff
@@ -49,8 +49,8 @@ export class DtExampleFilterFieldUnique {
             lessThanEqual: true,
           },
           unit: 's',
-          unique: true,
         },
+        unique: true,
```
ffriedl89: Thx for fixing this!

libs/testing/src/lib/event-objects.ts
```diff
@@ -54,7 +54,8 @@ export function createTouchEvent(
   const event = document.createEvent('UIEvent');
   const touchDetails = { pageX, pageY };
 
-  event.initUIEvent(type, true, true, window, 0);
+  // TS3.6 removes the initUIEvent method and suggests porting to "new UIEvent()".
```
ffriedl89: should we do that then? :D

components/expandable-section/src/expandable-section.ts
lukasholzer: we just used it for selecting the content. The `ng-content` was useless. Same public API after the changes as before.

image

apps/universal/src/app/kitchen-sink/kitchen-sink.html
lukasholzer: Just forgot to comment it in :)

.vscode/settings.json
tomheller: The only reason I see that it should stay in there is, that it locks your typescript version of vscode to the one installed in node_modules, and not the one shipped with the editor. 
This can be a huge hassle when developing (because ts is behaving very differently in in minor version changes). 
I would keep the file in there, but please remove your change.

angular.json
tomheller: > We have it already in place in all the other apps: barista, components-e2e, dev

We did not have this before, these were added in your changeset. I'd still consider at least creating an issue to talk about the numbers in there, instead of blindly applying the defaults it ships with.

components/breadcrumbs/src/item/breadcrumbs-item.spec.ts
tomheller: yes, you are right. I was just not sure if its not already resolved. Thank you for clarifying.

components/event-chart/src/event-chart.ts
lukasholzer: The index is only for identifying and it is working perfectly this way – take a look at this stackblitz where I demonstrate the behaviour:

https://stackblitz.com/edit/angular-prnpnp?file=src%2Fapp%2Fapp.component.ts

apps/components-e2e/src/components/expandable-section/expandable-section.e2e.ts
```diff
@@ -25,31 +25,31 @@ fixture('Expandable Section').page('http://localhost:4200/expandable-section');
...
+  await testController.expect(await open.textContent).eql('true');
   await testController.click(buttonDisable);
-  await testController.expect(await open.textContent).eql('0');
+  await testController.expect(await open.textContent).eql('false');
 });
 
 test('should open on label click', async (testController: TestController) => {
   await testController.click(labelHeader);
-  await testController.expect(await open.textContent).eql('1');
+  await testController.expect(await open.textContent).eql('true');
```
tomheller: BIT: could be chained to improve speed and flakyness.

tomheller:
Please resolve merge conflicts.

apps/components-e2e/src/components/expandable-section/expandable-section.e2e.ts
lukasholzer: Good point will do 💪

components/expandable-section/src/expandable-section.ts
```diff
@@ -55,38 +53,31 @@ export class DtExpandableSectionHeader {}
...
     return this._panel.expanded;
   }
   set expanded(value: boolean) {
-    this._panel.expanded = value;
+    this._panel.expanded = coerceBooleanProperty(value);
     this._changeDetectorRef.markForCheck();
   }
 
-  /** Whether the expandable section is disabled. */
+  /** Whether the expandable section is expanded. */
```
lara-aigmueller: Why change it to "expanded" when the input says "disabled"?

lukasholzer: The git diff is odd :D it is a copy pasta 🍝 error because I copied the input with doc from the expanded input some lines above :D

PR #591, opened 2020-02-12
fix(table): Fixed invalid fill color for favorite-column icon

tscherfler:
After consulting UX, the color of the favorite star on the table should be torquise-800.


ffriedl89:
FYI: @yngrdyn

ffriedl89:
Merging since flaky e2e test is already being fixed in a separate PR and is not affected by the change in this PR

Issue #570, opened 2020-02-07
Create Nx Linting rules to enforce module boundaries and correct imports


lukasholzer:
# Feature Request

Add the module boundaries linting rule and check for correct submodule imports:
I've seen a lot of imports like that `import { } from 'components/core/src/testing';` or `import { } from 'tools/shared/src/utils';`

So every import that imports not from the barrel via the path alias in the tsconfig or violates the module boundaries should be banned!

This is really a high prio task!

ffriedl89:
We need to adhere to the conventions of nx having everything inside apps or libs folders. 

See 

```
function isAbsoluteImportIntoAnotherProject(imp) {
    return (imp.startsWith('libs/') ||
        imp.startsWith('/libs/') ||
        imp.startsWith('apps/') ||
        imp.startsWith('/apps/'));
}
```

The check above only works if you follow this structure. Therefore will move components and tools inside libs folder

PR #602, opened 2020-02-17
Workspace: Added first constraints and resolved circular references

ffriedl89:
This PR is the first step towards resolving #570.

A couple of things to note here: 
* We ran into naming issues between the `examples` library in the libs folder and the `examples` library in tools folder the linting rules shipped by NX. 
* I moved the form-field tests into the input.spec file due to a cyclic dependency
* I will move the components folder into the libs folder in a separate PR. 
* We split up the testing library into 3 libs to control where to import from more granularly.


libs/testing/browser/README.md
```diff
@@ -0,0 +1,8 @@
+# testing-browser
+
+This library was generated with [Nx](https://nx.dev).
```
tomheller: Cool

tslint.json
```diff
@@ -61,6 +62,28 @@
...
+          },
+          {
+            "sourceTag": "scope:examples",
+            "onlyDependOnLibsWithTags": ["scope:components"]
+          }
+        ]
+      }
+    ],
+
+    // Custom Rules
```
NinaKammerlander: Does that work? I thought comments are not allowed in json files.

lukasholzer: in tslint it is allowed but in my opinion, they should be avoided.

lukasholzer:
please fix failing unit test of the event chart

components/filter-field/src/filter-field-tag/filter-field-tag.spec.ts
```diff
@@ -38,11 +38,10 @@ import { Platform } from '@angular/cdk/platform';
...
   createComponent,
   MockNgZone,
-} from '@dynatrace/barista-components/testing';
-import {
-  mockObjectProperty,
   mockGetComputedStyle,
-} from '@dynatrace/barista-components/testing/mock';
+} from '@dynatrace/barista-components/testing/browser';
+
+import { mockObjectProperty } from '@dynatrace/barista-components/testing/node';
```
lukasholzer: this is not ideal to import inside a component from testing/node...

ffriedl89: Its the spec file ;)

ffriedl89:
This PR is currently blocked by the upgrade to nx 9 
#588

PR #606, opened 2020-02-17
Barista/fix color grid and toc

lara-aigmueller:

This PR contains 2 small fixes for Barista:
1) The order of content components is relevant in case they are nested; this is the case with the color-grid / grid / grid-item components.
2) The check for the TOC property had to be fixed to set the toc-metadata entry correctly.

#### Type of PR

Barista

#### Checklist

### DIFF
--- a/apps/barista/src/components/components.module.ts
+++ b/apps/barista/src/components/components.module.ts
@@ -29,6 +29,11 @@ import { BaLayoutGrid } from './layout-grid/layout-grid';
 import { BaColor } from './color-component/color';
 import { BaLayoutGridItem } from './layout-grid/layout-grid-item';
 
+/**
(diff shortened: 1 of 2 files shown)


apps/barista/src/components/components.module.ts
```diff
@@ -29,16 +29,21 @@ import { BaLayoutGrid } from './layout-grid/layout-grid';
 import { BaColor } from './color-component/color';
 import { BaLayoutGridItem } from './layout-grid/layout-grid-item';
 
+/**
+ * The order of the following components is relevant in case they are nested.
+ * Inner components must be instantiated first. This is why the grid-item
+ * comes before the grid and the grid before the color-grid.
+ */
```
tomheller: NIT: Could also not be a JSdoc comment

PR #607, opened 2020-02-17
fix(tag-list): Fixes an issue where the taglist did not get updated within a tree table.

### LINKED ISSUE #603: Show more placeholder is shown in the tag list although there are no more tags available in the tag list
# Bug Report
## Expected Behavior

The tag list should not show the '0 More...' placeholder when there are no other tag components in the list to display.

## Current Behavior
The '0 More...' button placeholder is still displayed although there are no (more) tag entries. In some cases the '0 More...' disappears when clicking outside the tag list area and in some cases it is displayed for few moments and then it disappears on its own. 

## Steps to Reproduce

Please find attached a gif showing the behavior and the stackblitz for the issue.

## Context (Environment)
We are using the tag list component for displaying meaningful information when available and it misleading for the users to have a '0 More...' button when no (other) entry is available.

### Used Versions:

- **node**: 12.13.0
- **npm**: 
- **typescript**: 3.5.3
- **angular**: 8.0.0
- **@angular/cdk**: 8.0.0
- **@dynatrace/barista-components**: 5.2.0

## Attachments
tag_list_viewer

StackBlitz: [link](https://qgnirvwwa.github.stackblitz.io/)

tomheller:

When running inside of a tree table, the table row does only get updated
when the differ of the tree table triggers a repaint. As the tag-list
itself marks itself dirty, but change detection never runs for it
because no differ detectable changes happened.

Fixes #603

#### Type of PR
Bugfix

#### Checklist

### DIFF
--- a/apps/dev/src/tree-table/tree-table-demo.component.html
+++ b/apps/dev/src/tree-table/tree-table-demo.component.html
@@ -52,7 +52,13 @@
     <dt-tree-table-header-cell *dtHeaderCellDef>
       Running
     </dt-tree-table-header-cell>
-    <dt-cell *dtCellDef="let row">{{ row.running }}ms</dt-cell>
+    <dt-cell *dtCellDef="let row">
+      <dt-tag-list>
+        <dt-tag *ngFor="let tag of row.tags">
+          {{ tag }}
+        </dt-tag>
+      </dt-tag-list>
+    </dt-cell>
   </ng-container>
 
   <ng-container dtColumnDef="waiting">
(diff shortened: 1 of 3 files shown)


ffriedl89:
I feel we still need a followup issue for this since the tag-list does not seem to work on stackblitz - my best guess it does not work due to stackblitz not having the microTasks cleared.

@rowa-audil Did you already have time to check it out? It would be amazing if you could create a followup to fix this in stackblitz as well :)

PR #617, opened 2020-02-18
feat(linting): Set severity from error to warning for custom linting …

### LINKED ISSUE #367: Linting rules: Rethink severity
Since we are opensource our linting rules might not fit everybody - we should rethink the severity - is it a good idea to set all linting rules to warning - so we can educate devs but don't force them into a corner if they have a edge case

lara-aigmueller:
…rules.

Two questions that came up:

1.) What about the two rules (line 23, 24) that are not enabled at all?
2.) Is this a breaking change, as we break expected behavior with this PR?

closes #367


tomheller:
As discussed in #367, would it be possible to ship a _Base_ linting in addition, that sets the rules to error, where we think they should be on error?

lukasholzer:
I agree with @tomheller about the baseline that throws an error. Linting rules that only warn tend to be ignored. Furthermore in my opinion linting rules exist for writing opinionated code so the person that is installing those might want our opinion about how a good component is developed otherwise she or he would not install them.

Regarding the first point – I would activate them if they are making sense. If they are not making sense we could delete them.

Regarding the second point, I think this is a breaking change because we break a behaviour on which people are relying on.

Maybe a good way to make our linting rules more fittable to anybody else than our product we could try to make them a little bit of configurable instead of setting them to warning. Because this is the issue we are trying to fix.

like:

```
"dt-tile-needs-title": { true, "h2" },
```

so that you can provide further input than only true or warning

lara-aigmueller:
As our custom linting rules are not enabled by default (using ng-add), it's optional to activate/use them in a project. As discussed with @ffriedl89 we will close this issue for now. We can reopen it, once we have a use case to add more options to configure the linting rules.

PR #622, opened 2020-02-19
test(event-chart): Adjusted a path equality test to be less restrictive.

tomheller:

Event chart tested for path equality and had mismatches in 10<sup>-8</sup> area. Changed the expectation to be less restrictive.

#### Type of PR
Testcase fixed

#### Checklist

### DIFF
--- a/components/event-chart/src/event-chart.spec.ts
+++ b/components/event-chart/src/event-chart.spec.ts
@@ -194,16 +194,59 @@ describe('DtEventChart', () => {
 
     it('should render the path correctly', () => {
       const path = getRenderedEventsPath(fixture);
-      const expectedPath = [
-        'M13 73',
-        'L146.0666666666667 24',
-        'L212.60000000000002 73',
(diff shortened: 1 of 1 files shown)


components/event-chart/src/event-chart.spec.ts
```diff
@@ -194,16 +194,59 @@ describe('DtEventChart', () => {
...
-        'L345.6666666666667 73',
-        'L478.73333333333335 73',
-        'M678.3333333333334 73',
-        'L1011 24',
-      ].join(' ');
-      expect(path).toBe(expectedPath);
+
+      // Split the path into points and check
+      const splitPath = path
+        .replace(/ ([M,L,C,A])/gim, ' *$1')
```
lukasholzer: you only work with absolute coordinates?

tomheller: No, that's what the `i` is for, case insensitive.

PR #633, opened 2020-02-21
chore: Added concepts for design tokens and web components.

tomheller:
This pull request should be considered the start of the discussion about the introduction of design-tokens and web-components into barista. Please feel free to comment, contribute and discuss here. 
This is the beginning of this conversation.

### DIFF
--- a/concepts/design-tokens/README.md
+++ b/concepts/design-tokens/README.md
@@ -1,0 +1,234 @@
+# Design tokens
+
+> Design tokens are an agnostic way to store variables such as typography,
+> color, and spacing so that your design system can be shared across platforms
+> like iOS, Android, and regular ol' websites.
+
+-- <cite>Robin Rendle "[What are design tokens]", CSS-Tricks</cite>
+
+## What we want to achive with Design tokens
+
+In the current state of the Barista components library, we do share some basic
+color, typography and spacing definitions through scss mixins and variables
+within the angular ecosystem. Which works very well for color definitions, did
+not prove to be consistent for spacings and typography definitions. As our plans
+for Barista do not only target Angular, but a more framework agnostic component
+library, we are re-evaluating the way on how we deal with design tokens.
+
+Extracted design tokens should enable us to be extremely consistent within the
+system. With everything, that is relevant to the design-language baked into
+tokens, we should be able to enforce usage of these tokens with linting rules
+like [stylelint declaration strict value].
+
+## Naming of design tokens
+
+To prevent conflicts with css custom properties or other variables in output
+files, we should prefix all our provided design tokens, mixins or exported
+constants with a `dt-`/`dt` prefix.
+
+## Structure
+
+Tokens are usually stored in some sort of structual data (yml, json,
+database,...). Which format we are ultimately choosing is still up for
+discussion.
+
+Just like the format, the file structure is also up for discussion. There are
+multiple ways of structuring the design tokens into a folder structure, and a
+conversion tool like [theo] will let you customize
+[input and output formats quite easily](#output-and-formats).
+
+Here are some variations of design tokens structure in the wild:
+
+- [Salesforce design tokens]
+- [Open table design tokens]
+- [FirefoxUX design tokens]
+- [Stu Robson SCL Tokens]
+
+This is what we currently envision the structure of the design tokens to look
+like:
+
+```
+aliases
+ | border-radius.alias.yml
+ | breakpoints.alias.yml
+ | families.alias.yml
+ | palette.alias.yml
+ | typography.alias.yml
+
+global
+ | colors.yml
+ | font-family.yml
+ | spacing.yml
+
+patterns/components
+ | button.yml
+ | card.yml
+ | switch.yml
+
+theme ?
+```
+
+### Aliases
+
+Aliases should be considered variable definitions that assign a specific value
+to a variable name. These variable definitions are not considered design tokens
+yet, but they define values to be used within the global or pattern definitions.
+Examples could look like this:
+
+```yml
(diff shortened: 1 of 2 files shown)


ffriedl89:
Really like the summary so far :) Very nice! 🚀 🎨 
No objections whatsoever, I think these constraints should help us ship more consistent components. The layer to control each component individually is awesome, this way we can swap out a spacing value on the button only if we notice that the button spacing is off.

tomheller:
We should definitely consider some sort of visual impact or visual regression comparison for token changes. I have no idea what this could look like, but I assume this could be very helpful to see the impact of a change in a global fashion.

ffriedl89:
I've added the exact same idea to my list of unordered ideas in notion :D Have a tool that shows the impact of the change in the design tokens. List of affected components, maybe with a before and after view :D  ¯\\_(ツ)_/¯

thomaspink:
As far as i have understood design tokens (in combination with css variables), we also need to prefix all our css-variables because the cascade also applies to them.

concepts/design-tokens/README.md
```diff
@@ -0,0 +1,229 @@
...
+logic. For some of these usecases, exported typescript constants for the design
+tokens could help us keep things in sync between typescript and style.
+
+Potential output could look like this:
+
+```ts
+/** ts-constants/button.ts*/
+export const baseButtonBackgroundColor = '#123456';
+export const baseButtonBackgroundColorHover = '#123456';
+export const primaryPuttonBackgroundColor = '#123456';
```
lara-aigmueller:
```suggestion
export const primaryButtonBackgroundColor = '#123456';
```

concepts/design-tokens/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+## Structure
+
+Tokens are usually stored in some sort of structual data (yml, json,
+database,...). Which format we are ultimately choosing is still up for
+discussion.
+
+Just like the format, the file structure is also up for discussion. There are
+multiple ways of structuring the design tokens into a folder structure, and a
+convertion tool like [theo] will let you customize
```
NinaKammerlander: conversion

concepts/design-tokens/README.md
```diff
@@ -0,0 +1,229 @@
...
+> color, and spacing so that your design system can be shared across platforms
+> like iOS, Android, and regular ol' websites.
+
+-- <cite>Robin Rendle "[What are design tokens]", CSS-Tricks</cite>
+
+## What we want to achived with Design tokens
+
+In the current state of the barista components library, we do share some basic
+color, typography and spacing definitions through scss mixins and variables
+within the angular eco system. Which works very well for color definitions, did
```
NinaKammerlander: ecosystem

concepts/design-tokens/README.md
tomheller: but my `primaryPutton` :(

tomheller:
Thank you for spotting all my typos, they are resolved now.

tomheller:
> I'm wondering if it makes sense to scope some CSS custom properties and not making them all available at root-level. But I'm sure this depends on when the consumer should be able to override those values and where and why.

I believe that this makes a lot of sense to scope the pattern/component tokens to their respective component. But we will need to figure out how we can or want to achieve that.

Possible downside would be, if you would want to restyle your buttons, you would have to restyle all of them over and over again vs. redefining the new variable definitions in a wrapper and let the cascade handle the rest (could be very helpful for theming if we would need that). Both options seem valid, I think this is something we need to discuss further. Any input is very much appreciated here.

thomaspink:
Great work @tomheller !!

Just another very small input: 
Currently top level constants in typescript are (mostly - we forgot it sometimes) in all uppercase, with underscores between. Should we stick to this or switch to camelCase.

`export const dtBaseButtonBackgroundColor = '#123456';` 
vs 
`export const DT_BASE_BUTTON_BACKGROUND_COLOR = '#123456'`;

For consistency with the angular codebase i'd suggest to stick with the uppercase variant (as it is also more common in the TS community). But i also have no objections if we wanna change that. We should just agree on one way or the other for the whole codebase.

tomheller:
>Currently top level constants in typescript are (mostly - we forgot it sometimes) in all uppercase, with underscores between. Should we stick to this or switch to camelCase.

Good point, I have updated the typescript tokens section accordingly

concepts/design-tokens/README.md
```diff
@@ -0,0 +1,234 @@
...
+discussion.
+
+Just like the format, the file structure is also up for discussion. There are
+multiple ways of structuring the design tokens into a folder structure, and a
+conversion tool like [theo] will let you customize
+[input and output formats quite easily](#output-and-formats).
+
+Here are some variations of design tokens structure in the wild:
+
+- [Salesforce design tokens]
```
lukasholzer: should this be a link?

tomheller: Yes, and it is. This is a shorthand link syntax in markdown. 
Link targets are collected at the bottom of the md file: 
https://github.com/dynatrace-oss/barista/pull/633/files#diff-3f1f6d3a5d17588e9f3aa79c5b52262eR228

ffriedl89:
I think this is great content, should we merge it? And where do we wanna have the concept section popup in barista? Should we add an internal next section on top level?

tomheller:
> I think this is great content, should we merge it? And where do we wanna have the concept section popup in barista? Should we add an internal next section on top level?

I don't know? Does this make sense to include in the published barista? It's more about specs / guidelines within the component library.

tomheller:
Needed to reword the commit, because CI was not running on it. Review has been requested again.

PR #650, opened 2020-02-26
barista: Rename githubuser key in barista config files.

lara-aigmueller:

Sorry, i broke stuff. :(

#### Type of PR

Barista

#### Checklist

### DIFF
--- a/components/alert/barista.json
+++ b/components/alert/barista.json
@@ -9,17 +9,17 @@
     "dev": [
       {
         "name": "Lara Aigmueller",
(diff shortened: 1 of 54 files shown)


tomheller:
Please do not merge before 6.0 is out, because the cherry picker would do some weird stuff if we would

lara-aigmueller:
Manually cherry-picked into 5.x, but not into 5.2.x, because there were merge-conflicts (radial-chart component is not in 5.2.x). But this is fine as for the Barista build we only need the latest master commit. No action required.

PR #654, opened 2020-02-27
feat(empty-state): Progressive enhancements for relayout based on ResizeObserver

tomheller:

Uses the ResizeObserver to actively match the width of the emptyState component and adjust the layout accordingly. 

#### Type of PR

Feature

#### Checklist

### DIFF
--- a/libs/barista-components/empty-state/src/empty-state.ts
+++ b/libs/barista-components/empty-state/src/empty-state.ts
@@ -37,7 +37,7 @@ import {
   ViewChild,
 } from '@angular/core';
 import { Subject } from 'rxjs';
-import { startWith, takeUntil } from 'rxjs/operators';
+import { takeUntil, startWith } from 'rxjs/operators';
 
 import {
   DtViewportResizer,
@@ -49,6 +49,8 @@ const ITEMS_HORIZONTAL_BREAKPOINT = 540;
 
 /** The min-width from which the empty state items are aligned next to each other. */
 const LAYOUT_HORIZONTAL_BREAKPOINT = 760;
+
+declare const window: any;
 
 /**
  * An empty state item. An empty state card may contain one or more such items.
@@ -162,6 +164,8 @@ export class DtEmptyState
 
   private readonly _destroy$ = new Subject<void>();
 
+  private _containerSizeObserver: any;
+
   /**
    * @internal
    * Whether empty state items should have a horizontal layout
@@ -186,6 +190,10 @@ export class DtEmptyState
   constructor(
     private _changeDetectorRef: ChangeDetectorRef,
     private _elementRef: ElementRef<HTMLElement>,
+    /**
+     * @deprecated Remove the viewportResizer from the constructor as it is no longer needed.
+     * @breaking-change Remove the viewportResizer in version 8.0.0
+     */
     private _viewportResizer: DtViewportResizer,
     private _platform: Platform,
   ) {}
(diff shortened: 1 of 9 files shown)


components/empty-state/src/empty-state.ts
```diff
@@ -50,6 +50,8 @@ const ITEMS_HORIZONTAL_BREAKPOINT = 540;
 /** The min-width from which the empty state items are aligned next to each other. */
 const LAYOUT_HORIZONTAL_BREAKPOINT = 760;
 
+declare const window: any;
```
lukasholzer: NIT: you can declare your own typings till it is part of the `lib.dom` package: microsoft/TypeScript#28502


```typescript
/** Workaround until created: microsoft/TypeScript#28502 */
interface ResizeObserver {
...
}
```

```suggestion
declare const window: { ResizeObserver?: ResizeObserver };
```

components/empty-state/src/empty-state.ts
```diff
@@ -186,7 +190,11 @@ export class DtEmptyState
   constructor(
     private _changeDetectorRef: ChangeDetectorRef,
     private _elementRef: ElementRef<HTMLElement>,
-    private _viewportResizer: DtViewportResizer,
+    /**
+     * @deprecated Remove the viewportResizer from the constructor as it is no longer needed.
+     * @breaking-change Remove the viewportResizer in version 8.0.0
```
lukasholzer: can we do this with version 7? :D

components/empty-state/src/empty-state.ts
```diff
@@ -197,48 +205,60 @@ export class DtEmptyState
...
+      this._containerSizeObserver = new window.ResizeObserver(entries => {
+        // We need to wrap the call to the layout update into an additional
+        // requestAnimationFrame, because the resize observer would trow a
+        // javascript error when it is not able to process all entries within
+        // a single animation frame.
+        // From the specification:
+        // > This error means that ResizeObserver was not able to deliver all
+        // > observations within a single animation frame. It is benign (your
+        // > site will not break). - Aleksandar Totic
+        // https://stackoverflow.com/a/50387233
```
lukasholzer: Nice explanation!

components/empty-state/src/empty-state.ts
```diff
@@ -197,48 +205,60 @@ export class DtEmptyState
...
+        window.requestAnimationFrame(() => {
+          this._updateLayout(entries[0].contentRect.width);
```
lukasholzer: Is it possible that entries is empty or has no index 0?

tomheller: Not really, that would mean that the elementRef of the host element is empty. But I can add a check there, just to be sure.

components/empty-state/src/empty-state.ts
tomheller: I believe that at some point we have discussed this in the past and decided that we will announce breaking-changes 2 major versions in advance.

libs/barista-components/empty-state/src/empty-state.ts
```diff
@@ -197,61 +202,52 @@ export class DtEmptyState
   }
 
   ngAfterViewInit(): void {
-    this._viewportResizer
-      .change()
-      .pipe(startWith(null), takeUntil(this._destroy$))
-      .subscribe(() => {
-        this._updateLayout();
+    // Check if the broser supports the resizeObserver.
```
lara-aigmueller: nit: browser

tomheller:
This pull request removes the Base class for the dtEmptyState, which is considered a breaking change. Please add it back in, so we don't have a breaking version in there.

components/empty-state/src/empty-state.ts
tomheller: Nope, doesn't work. Tried to type it with 4 typescript specialists at my side... did not work. At some point of our nested tsconfig structure, either the types, files, typeroots or includes are being overwritten... 
this is sad 😢

tomheller:
Put back the deprecated class and functions... Will be removable in version 8

lukasholzer:
formatting and unit tests are failing

tomheller:
fixed formating and unit test. Once more for good measure @lukasholzer @lara-aigmueller

PR #663, opened 2020-02-28
feat(drawer-table): Added drawer-table component and example demo page

### LINKED ISSUE #587: DrawerTable component
# Feature Request

## Summary
Drawer table is a component that adds collapsible side content to table rows.

## Feature Description

This component uses table and drawer components. Table is the main content of the drawer and row click of the table opens the drawer content of the row. I think that this component should be a separate component since I could not find the way to include the drawer component inside table.  Columns which stay open when row is clicked can be defined like in example: 
 ``` 
<dt-drawer-table [openColumns]="['col1']">
          <ng-template let-row>
              <div class="row_col1">{{ row.col1 }}</div>
          </ng-template>
          <dt-table [dataSource]="dataSource" dtSort #sortable>
              <dt-simple-number-column
                      name="col1"
                      label="Column1"
              ></dt-simple-number-column>
              <dt-simple-number-column
                      name="col2"
                      label="Column2"
              ></dt-simple-number-column>
              <dt-simple-number-column
                      name="col3"
                      label="Column3"
              ></dt-simple-number-column>
              <dt-header-row
                      *dtHeaderRowDef="['col1', 'col2', 'col3']"
              ></dt-header-row>
              <dt-row [dtDrawerRowDef]="row"
                      *dtRowDef=" let row; columns: ['col1', 'col2', 'col3']"
              ></dt-row>
          </dt-table>
      </dt-drawer-table>
```

Here is the demo of the component: 

drawer-table

### PR DESCRIPTION
**Pull request by madina705**
Closes #587

<hr>
The README of this component is not created yet. The pull request was marked as draft to get early feedback on this component

Closes #587 

#### Type of PR

#### Checklist

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -1752,6 +1752,48 @@
       },
       "schematics": {}
     },
+    "drawer-table": {
+      "projectType": "library",
+      "root": "libs/barista-components/experimental/drawer-table",
+      "sourceRoot": "libs/barista-components/experimental/drawer-table/src",
+      "prefix": "dt",
+      "architect": {
+        "lint": {
+          "builder": "@angular-devkit/build-angular:tslint",
+          "options": {
+            "tsConfig": [
+              "libs/barista-components/experimental/drawer-table/tsconfig.lib.json",
+              "libs/barista-components/experimental/drawer-table/tsconfig.spec.json"
+            ],
+            "exclude": [
+              "**/node_modules/**",
+              "!libs/barista-components/experimental/drawer-table/**"
+            ]
+          }
+        },
+        "lint-styles": {
+          "builder": "./dist/libs/workspace:stylelint",
+          "options": {
+            "stylelintConfig": ".stylelintrc",
+            "reportFile": "dist/stylelint/report.xml",
+            "exclude": ["**/node_modules/**"],
+            "files": [
+              "libs/barista-components/experimental/drawer-table/**/*.scss"
+            ]
+          }
+        },
+        "test": {
+          "builder": "@nrwl/jest:jest",
+          "options": {
+            "jestConfig": "libs/barista-components/experimental/drawer-table/jest.config.js",
+            "tsConfig": "libs/barista-components/experimental/drawer-table/tsconfig.spec.json",
+            "setupFile": "libs/barista-components/experimental/drawer-table/src/test-setup.ts",
+            "passWithNoTests": true
+          }
+        }
+      },
+      "schematics": {}
+    },
     "empty-state": {
       "projectType": "library",
       "root": "libs/barista-components/empty-state",
--- a/libs/barista-components/experimental/drawer-table/README.md
+++ b/libs/barista-components/experimental/drawer-table/README.md
@@ -1,0 +1,39 @@
+# Drawer table (experimental)
+
+<ba-live-example name="DtExampleDrawerTableDefault" fullwidth></ba-live-example>
+
+The `DtDrawerTable` component uses table and drawer components. Table is the
+main content of the drawer and row click of the table opens the drawer content
+of the row.
+
+## Imports
+
+You have to import the `DtDrawerTableModule` to use the `dt-drawer-table`.
+
+```typescript
+import { NgModule } from '@angular/core';
+import { DtDrawerTableModule } from '@dynatrace/barista-components/experimental/drawer-table';
+
+@NgModule({
+  imports: [DtDrawerTableModule],
+})
+class MyModule {}
+```
+
+## Inputs
+
+The `DtDrawerTable` component supports the following input.
+
+| Name          | Type       | Default      | Description                                                    |
+| ------------- | ---------- | ------------ | -------------------------------------------------------------- |
+| `openColumns` | `string[]` | first column | Defines the columns that stay visible when the drawer is open. |
+
+#### Methods
+
+The following methods are on the `DtDrawerTable` class:
+
+| Name     | Description        | Return value |
+| -------- | ------------------ | ------------ |
+| `open`   | Opens the drawer   | `void`       |
+| `close`  | Closes the drawer  | `void`       |
+| `toggle` | Toggles the drawer | `void`       |
--- a/apps/dev/src/app.module.ts
+++ b/apps/dev/src/app.module.ts
@@ -45,6 +45,7 @@ import { ContextDialogDemo } from './context-dialog/context-dialog-demo.componen
 import { CopyToClipboardDemo } from './copy-to-clipboard/copy-to-clipboard-demo.component';
 import { CtaCardDemo } from './cta-card/cta-card-demo.component';
 import { DrawerDemo } from './drawer/drawer-demo.component';
+import { DrawerTableDemo } from './drawer-table/drawer-table-demo.component';
 import { EmptyStateDemo } from './empty-state/empty-state-demo';
 import { EventChartDemo } from './event-chart/event-chart-demo.component';
 import { ExpandablePanelDemo } from './expandable-panel/expandable-panel-demo.component';
@@ -122,6 +123,7 @@ export class NoopRouteComponent {}
     CopyToClipboardDemo,
     CtaCardDemo,
     DrawerDemo,
+    DrawerTableDemo,
     ExpandablePanelDemo,
     ExpandableSectionDemo,
     FilterFieldDemo,
--- a/apps/dev/src/devapp-routing.module.ts
+++ b/apps/dev/src/devapp-routing.module.ts
@@ -33,6 +33,7 @@ import { ContextDialogDemo } from './context-dialog/context-dialog-demo.componen
 import { CopyToClipboardDemo } from './copy-to-clipboard/copy-to-clipboard-demo.component';
 import { CtaCardDemo } from './cta-card/cta-card-demo.component';
 import { DrawerDemo } from './drawer/drawer-demo.component';
+import { DrawerTableDemo } from './drawer-table/drawer-table-demo.component';
 import { EmptyStateDemo } from './empty-state/empty-state-demo';
 import { EventChartDemo } from './event-chart/event-chart-demo.component';
 import { ExpandablePanelDemo } from './expandable-panel/expandable-panel-demo.component';
@@ -91,6 +92,7 @@ const routes: Routes = [
   { path: 'copy-to-clipboard', component: CopyToClipboardDemo },
   { path: 'cta-card', component: CtaCardDemo },
   { path: 'drawer', component: DrawerDemo },
+  { path: 'drawer-table', component: DrawerTableDemo },
   { path: 'empty-state', component: EmptyStateDemo },
   { path: 'expandable-panel', component: ExpandablePanelDemo },
   { path: 'expandable-section', component: ExpandableSectionDemo },
--- a/apps/dev/src/devapp.component.ts
+++ b/apps/dev/src/devapp.component.ts
@@ -59,6 +59,7 @@ export class DevApp implements AfterContentInit, OnDestroy {
     { name: 'Copy-to-clipboard', route: '/copy-to-clipboard' },
     { name: 'Cta-card', route: '/cta-card' },
     { name: 'Drawer', route: '/drawer' },
+    { name: 'Drawer-table', route: '/drawer-table' },
     { name: 'Empty-state', route: '/empty-state' },
     { name: 'Event-chart', route: '/event-chart' },
     { name: 'Expandable-panel', route: '/expandable-panel' },
--- a/apps/dev/src/drawer-table/drawer-table-demo.component.html
+++ b/apps/dev/src/drawer-table/drawer-table-demo.component.html
@@ -1,0 +1,24 @@
+<dt-drawer-table [openColumns]="['col2']">
+  <ng-template dtDrawerContent let-row>
+    <div class="row_col1">{{ row.col1 }}</div>
+  </ng-template>
+  <dt-table [dataSource]="dataSource" dtSort #sortable>
+    <dt-simple-number-column
+      name="col1"
+      label="Column1"
+    ></dt-simple-number-column>
+    <dt-simple-number-column
+      name="col2"
+      label="Column2"
+    ></dt-simple-number-column>
+    <dt-simple-number-column
+      name="col3"
+      label="Column3"
+    ></dt-simple-number-column>
+    <dt-header-row *dtHeaderRowDef="['col1', 'col2', 'col3']"></dt-header-row>
+    <dt-row
+      [dtDrawerRowDef]="row"
+      *dtRowDef="let row; columns: ['col1', 'col2', 'col3']"
+    ></dt-row>
+  </dt-table>
+</dt-drawer-table>
--- a/apps/dev/src/drawer-table/drawer-table-demo.component.ts
+++ b/apps/dev/src/drawer-table/drawer-table-demo.component.ts
@@ -1,0 +1,30 @@
+/**
+ * @license
+ * Copyright 2020 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'drawer-table-demo',
+  templateUrl: './drawer-table-demo.component.html',
+})
+export class DrawerTableDemo {
+  dataSource = [
+    { col1: 'test 1', col2: 'test 1', col3: 'test 1' },
+    { col1: 'test 2', col2: 'test 2', col3: 'test 2' },
+    { col1: 'test 3', col2: 'test 3', col3: 'test 3' },
+    { col1: 'test 4', col2: 'test 4', col3: 'test 4' },
+  ];
+}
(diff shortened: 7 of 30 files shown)

### CODE REVIEW

**Comment by lukasholzer** [comment]
@madina705 would you be so kind and rebase your changes with the latest master.
With version 6.0.0 we changed the location of the components folder and refactored some internal workspace structure.

Some steps you have to adapt:

1. You would have to move your `drawer-table` component to `/libs/barista-components/drawer-table`.
2. Adapt the paths in the `angular.json` from `components/` to `libs/barista-components/`.
3. Adapt the path in the root `tsconfig.json`
4. Adapt the `../../../` paths inside the following files to point correctly to the root:
  - `components/drawer-table/tslint.json`
  - `components/drawer-table/tsconfig.spec.json`
  - `components/drawer-table/tsconfig.lib.json`
  - `components/drawer-table/tsconfig.json`
  - `components/drawer-table/jest.config.js`
5. In the `components/drawer-table/jest.config.js` update the `snapshotSerializer` to the following content:

    ```
      snapshotSerializers: [
      'jest-preset-angular/build/AngularNoNgAttributesSnapshotSerializer.js',
      'jest-preset-angular/build/AngularSnapshotSerializer.js',
      'jest-preset-angular/build/HTMLCommentSerializer.js',
      ],
    ```
6. Try with `ng test drawer-table` and `ng lint drawer-table` if everything is working.

Sorry for the inconvenience. If you need some help by moving the files and adapting the paths please reach out to me via DM

File: angular.json [hunk 1]
```diff
@@ -3042,6 +3042,37 @@
...
+        "lint": {
+          "builder": "@angular-devkit/build-angular:tslint",
+          "options": {
+            "tsConfig": [
+              "components/drawer-table/tsconfig.lib.json",
+              "components/drawer-table/tsconfig.spec.json"
+            ],
+            "exclude": ["**/node_modules/**", "!components/drawer-table/**"]
+          }
+        },
```
lukasholzer: please add the stylelint target as well:


```json

        "lint-styles": {
          "builder": "@dynatrace/barista-builders:stylelint",
          "options": {
            "stylelintConfig": ".stylelintrc",
            "reportFile": "dist/stylelint/report.xml",
            "exclude": ["**/node_modules/**"],
            "files": ["libs/barista-components/drawer-table/**/*.scss"]
          }
        },
```

ffriedl89:
@madina705 Let us know once the PR is ready for review - so we don't delay it :)

libs/barista-components/drawer-table/src/drawer-table.scss
```diff
@@ -0,0 +1,18 @@
+$turquoise-600: #00a1b2;
```
tomheller: You can import our colorpalette from `@import '../../core/src/style/variables';`
It should not be necessary to redefine variables (especially color ones).

libs/barista-components/drawer-table/src/drawer-table.scss
```diff
@@ -0,0 +1,18 @@
...
+  width: 100%;
+  border: none;
+}
+
+:host ::ng-deep .dt-drawer-row-def {
+  cursor: pointer;
+}
+
+:host ::ng-deep .dt-drawer-row-def-selected {
+  border: 1px solid $turquoise-600 !important;
```
tomheller: I do hope that we can do this without an important. `!important` is impossible to override. I would rather opt for a specificity increase with a selector like this: 
```scss
:host ::ng-deep .dt-drawer-row-def-selected.dt-drawer-row-def-selected  {
 border: 1px solid $turquoise-600;
}
```
This increases the rules specificity from 0020 to 0030 (important would increase it to 1020).

Depending on what you actually want to change about the selected row, it might be sufficient to change the `border-color`, instead of redefining the whole border. If we ever change the border styling of the selected row, the drawer-table would not benefit from that change and it would lead to increasing inconsistency.

libs/barista-components/drawer-table/README.md
```diff
@@ -0,0 +1,71 @@
...
+| Name     | Description        | Return value |
+| -------- | ------------------ | ------------ |
+| `open`   | Opens the drawer   | `void`       |
+| `close`  | Closes the drawer  | `void`       |
+| `toggle` | Toggles the drawer | `void`       |
+
+## Basic example
+
+**An ng-template that defines the drawer's content has to be the first child of
+the component**
```
tomheller: it doesn't necessarily need to be the first child of the component. It's defined as a content child for the drawer-table, which requires it to be a direct child of the drawer table, but the order should be irrelevant.

libs/barista-components/drawer-table/src/drawer-table.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  /** Reference to the drawer element  */
+  @ViewChild(DtDrawer, { static: true }) private drawer: DtDrawer;
+
+  /** Reference to the drawer container  */
+  @ViewChild('container', { read: ViewContainerRef, static: false })
+  private container: ViewContainerRef;
+
+  /** Template reference for the drawer content. */
+  @ContentChild(TemplateRef, { static: false }) drawerContentRef: TemplateRef<{
```
tomheller: Can be marked as internal and prefixed with `_` or even private

libs/barista-components/drawer-table/src/drawer-table.ts
```diff
@@ -0,0 +1,188 @@
...
+    $implicit: T;
+  }>;
+
+  /** Reference to the table that is used to make the rows interactive */
+  @ContentChild(DtTable, { static: false }) table: DtTable<T>;
```
tomheller: Can be marked as internal and prefixed with `_` or even private

libs/barista-components/drawer-table/src/drawer-table.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  /** Reference to the header row definition which is used to set the columns  */
+  @ContentChild(DtHeaderRowDef, { static: false }) rowHeaderDef: DtHeaderRowDef;
```
tomheller: Can be marked as internal and prefixed with `_` or even private

libs/barista-components/drawer-table/src/drawer-table.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  /** Reference to the row definition which is used to get and set the columns  */
+  @ContentChild(DtRowDef, { static: false }) rowDef: DtRowDef<T>;
+  private initialColumns: Iterable<string>;
```
tomheller: Is there a specific reason why you are using `Iterable` typing here instead of an array? Is the rowDef columns from the cdk also typed as `Iterable`

libs/barista-components/drawer-table/src/drawer-table.ts
```diff
@@ -0,0 +1,188 @@
...
+  /** Sets the drawer content */
+  setDrawerContent(row: T): void {
+    this.container.clear();
+    this.container.createEmbeddedView(this.drawerContentRef, {
+      $implicit: row,
+    });
+  }
+
+  /** Row click opens the drawer and sets its content or closes if the drawer is already open */
+  onRowClicked(row: T): void {
```
tomheller: can be an internal.

libs/barista-components/drawer-table/src/drawer-table.ts
```diff
@@ -0,0 +1,188 @@
...
+  row: T;
+
+  constructor(private readonly drawerTable: DtDrawerTable<T>) {}
+
+  get borderStyle(): boolean {
+    if (this.drawerTable.currentRow === this.row && this.drawerTable.isOpen()) {
+      return true;
+    } else {
+      return false;
+    }
```
tomheller: I believe this would be a little simpler. No need for the extra if here.
```suggestion
    return this.drawerTable.currentRow === this.row && this.drawerTable.isOpen();
```

tomheller:
@madina705 Great job so far. You are making good progress. I have commented on a couple of things in your status quo to keep the feedback cycles shorter. 
If you do not agree with one of my comments, I'm very open to discussions here. Please leave a comment in the discussion if there is something I've missed.

libs/barista-components/drawer-table/README.md
madina705: Yes it is strange, but it doesn't work when it isn't the first child. I have checked once more now to be sure and it couldn't find the template when it is defined after the table.

tomheller: This is definitely weird. This should work and probably requires some investigation. I guess you have found some weird behavior on this one. I will have a look on monday to see, why it will not find the template if it's not the first child.

libs/barista-components/drawer-table/src/drawer-table.ts
tomheller: If you declare your own directive for the templateRef and read the templateRef, it will no longer be necessary for the ngTemplate to be the first child:

pseudo code example: 
```typescript
@ContentChild(DtDrawerContentRef, { static: false, read: TemplateRef }) drawerContentRef: TemplateRef<{ $implicit: T; }>;
```
and in the template: 
```html
...
<ng-template DtDrawerContentRef>...</ng-templae>
...
```

libs/barista-components/drawer-table/README.md
tomheller: If you declare your own directive for the templateRef it should be fine, wherever it is. See https://github.com/dynatrace-oss/barista/pull/663/files#r392887831

The ContentChild will look for the first templateRef match, and since some directives from the table do extend from it, it is quite possible to get a wrong match. 
to be safe here, you can define your own directive and read the templateRef.

tomheller: If you declare your own directive for the templateRef it should be fine, wherever it is. See https://github.com/dynatrace-oss/barista/pull/663/files#r392887831

The ContentChild will look for the first templateRef match, and since some directives from the table do extend from it, it is quite possible to get a wrong match. 
to be safe here, you can define your own directive and read the templateRef.

libs/barista-components/drawer-table/src/drawer-table.ts
madina705: Yes, the reason why I used Iterable was that columns from cdk also types as Iterable. Do you recommend using array instead and converting the columns to array when assigning?

libs/barista-components/drawer-table/src/drawer-table.ts
madina705: I tried this and it still returns undefined, couldn't find the template again.

libs/barista-components/drawer-table/src/drawer-table.ts
tomheller: No this is fine. I'd rather not convert if the cdk table uses `Iterable`. In the end, the cdk needs to deal with the iteration. Keep it as Iterable please.

libs/barista-components/drawer-table/src/drawer-table.ts
```diff
@@ -0,0 +1,188 @@
...
+    this._setColumns(this.getOpenColumns());
+  }
+
+  /**
+   * Closes the drawer if it is not already closed.
+   */
+  close(): void {
+    this._changeDetectorRef.markForCheck();
+    this.drawer.close();
+    this._setColumns(this.initialColumns);
```
tomheller: Although it should not matter too much, but maybe call `markForCheck` after you have done the changes: 
```suggestion
    this.drawer.close();
    this._setColumns(this.initialColumns);
    this._changeDetectorRef.markForCheck();
```

libs/barista-components/drawer-table/src/drawer-table.ts
```diff
@@ -0,0 +1,188 @@
...
+    this.rowDef.columns = columns;
+  }
+
+  /**
+   * Opens the drawer if it is not already opened.
+   */
+  open(): void {
+    this._changeDetectorRef.markForCheck();
+    this.drawer.open();
+    this._setColumns(this.getOpenColumns());
```
tomheller: Although it should not matter too much, but maybe call `markForCheck` after you have done the changes: 
```suggestion
    this.drawer.open();
    this._setColumns(this.getOpenColumns());
    this._changeDetectorRef.markForCheck();
```

libs/barista-components/drawer-table/src/drawer-table.ts
tomheller: Hi @madina705!

It worked for me when I tested it, here is what I changed and in which files: 

drawer-table.ts added: 
```typescript
@Directive({
  selector: '[dtDrawerContentRef]',
})
export class DtDrawerContentRef {}
```

drawer-table.ts L63-67 changed to:
```typescript
/** @internal Template reference for the drawer content. */
  @ContentChild(DtDrawerContentRef, { static: false, read: TemplateRef })
  private _drawerContentRef: TemplateRef<{
    $implicit: T;
  }>;
```

drawer-table-module.ts L17 changed to:
```typescript
import { DtDrawerRowDef, DtDrawerTable, DtDrawerContentRef } from './drawer-table';
```

drawer-table-module.ts L25-26 changed to:
```typescript
  declarations: [DtDrawerTable, DtDrawerRowDef, DtDrawerContentRef],
  exports: [DtDrawerTable, DtDrawerRowDef, DtDrawerContentRef],
```

drawer-table-demo.component.html, moved ng-template div to the end of the dt-drawer-table and changed to:
```html
<ng-template dtDrawerContentRef let-row>
   <div class="row_col1">{{ row.col1 }} AAA</div>
</ng-template>
```

You may have forgotten to add the new directive to the declarations in the module, which sadly does not throw any error...

madina705: Hi @tomheller,

Yes it worked, you are right I forgot to add the new directive to the declarations in the module. Thank you very much :)

libs/barista-components/drawer-table/src/drawer-table.ts
```diff
@@ -0,0 +1,193 @@
...
+  /** @internal Reference to the table that is used to make the rows interactive */
+  @ContentChild(DtTable, { static: false }) private _table: DtTable<T>;
+
+  /** @internal Reference to the header row definition which is used to set the columns  */
+  @ContentChild(DtHeaderRowDef, { static: false })
+  private _rowHeaderDef: DtHeaderRowDef;
+
+  /** Reference to the row definition which is used to get and set the columns  */
+  @ContentChild(DtRowDef, { static: false }) rowDef: DtRowDef<T>;
+  private initialColumns: Iterable<string>;
```
ffriedl89: Prefix as internal with  _
```suggestion
  private _initialColumns: Iterable<string>;
```

libs/barista-components/drawer-table/src/drawer-table.ts
```diff
@@ -0,0 +1,193 @@
...
+    }
+    this._currentRow = row;
+  }
+
+  getOpenColumns(): string[] {
+    return isEmpty(this.openColumns)
+      ? [this.initialColumns[0]]
+      : this.openColumns;
+  }
+
```
ffriedl89: we should use the openColumns input getter to return the correct values. 
So we can remove this function here, but we need to make sure to set the values for openColumns correctly. 
We might need to use Promise.resolve().then in the ngAfterContentInit lifecycle hook to set the values for the openColumns input if none are given to the initialColumns. 

If you need help on that one we can sync any time :)

libs/barista-components/drawer-table/src/drawer-table.ts
```diff
@@ -0,0 +1,193 @@
...
+    '(click)': '_handleRowClick()',
+  },
+})
+export class DtDrawerRowDef<T> {
+  @Input('dtDrawerRowDef')
+  row: T;
+
+  constructor(private readonly drawerTable: DtDrawerTable<T>) {}
+
+  get borderStyle(): boolean {
```
ffriedl89: If i understood it correctly this is the open state - could we rename this to `_isOpen` and make it internal

libs/barista-components/drawer-table/src/drawer-table.ts
```diff
@@ -0,0 +1,193 @@
...
+
+/**
+ * Data row definition for the dt-table of the drawer-table component.
+ * Used to handle click event and apply style.
+ */
+@Directive({
+  selector: '[dtDrawerRowDef]',
+  exportAs: 'dtDrawerRowDef',
+  host: {
+    class: 'dt-drawer-row-def-sse',
```
ffriedl89: I was not able to figure out what see means here :D What is the meaning of see? Can we remove it?

libs/barista-components/drawer-table/src/drawer-table.ts
```diff
@@ -0,0 +1,193 @@
...
+    return !isEmpty(this.drawer) && this.drawer.opened;
+  }
+
+  /** Sets the drawer content */
+  setDrawerContent(row: T): void {
+    this.container.clear();
+    this.container.createEmbeddedView(this._drawerContentRef, {
+      $implicit: row,
+    });
+  }
```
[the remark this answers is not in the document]
madina705: I think it is not needed as part of API, made it internal

libs/barista-components/drawer-table/src/drawer-table.ts
madina705: I think I was testing somethings and forgot to remove that :smile:

ffriedl89:
I've done a little cleanup, once you had a chance to checkout my commit we can squash everything together and work on the experimental flagging and finalize it

libs/barista-components/experimental/drawer-table/README.md
```diff
@@ -0,0 +1,39 @@
+# Drawer table
+
```
tomheller, line 2: @ffriedl89 should we add a experimental notice here as well?

ffriedl89: Thanks for spotting this - done!

ffriedl89:
@tomheller & @lukasholzer I think we can have this as ready and put the label to cherry pick on it.

tomheller:
Cherry picked into 6.x, new head is cfd7766f9ad32f100a9f4933ae3a817994f07890

tomheller:
Great job madina, congratulations on your first merged contribution 🎉

PR #682, opened 2020-03-02
chore: Fix path to shareable examples lib

tomheller:

Fixed the path to the shareable lib tooling. 

#### Type of PR

Chore

#### Checklist

### DIFF
--- a/environments/barista-environment.prod.ts
+++ b/environments/barista-environment.prod.ts
@@ -45,6 +45,7 @@ export const environment: BaEnvironment = {
   examplesLibDir: join(ROOT_DIR, 'libs', 'examples', 'src'),
   shareableExamplesToolsDir: join(
     ROOT_DIR,
+    'libs',
     'tools',
     'shareable-examples',
     'src',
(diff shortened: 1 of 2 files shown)


tomheller:
>May we change the ROOT_DIR to the process.cwd() to avoid errors by moving folders around? I Bet the location of this file will change in the future.

Would not have solved this issue to be honest. I'm not entirely happy to move to `process.cwd()`, because it then depends on where you start the process from, which is not the case now.

lukasholzer:
> > May we change the ROOT_DIR to the process.cwd() to avoid errors by moving folders around? I Bet the location of this file will change in the future.
> 
> Would not have solved this issue to be honest. I'm not entirely happy to move to `process.cwd()`, because it then depends on where you start the process from, which is not the case now.

You are right it does not solve this issue but would prevent future issues. In a monorepo you should execute every task from the root (all the ng/nx commands only work in the root), because  you can have dependencies to other libraries.

tomheller:
>  In a monorepo you should execute every task from the root 

Yes, I agree that you should. But you cannot enforce that people do exactly that. To rely on a relative structure, that you can control (you know where you start with a `__dirname` based on a file) you can more securely resolve the correct path.

> all the ng/nx commands only work in the root

This is not true, they work in sub paths of the project as well. because they go up the folder structure just as node does, to look for their respective .json files.

image

PR #697, opened 2020-03-05
fix(chart-heatfield): Chart HeatField is shown when start and end are…

sherifElhefnawyDynatrace:

- Button/heatfield is now hidden if both start and end are undefined
- Added test to check for start and end being both undefined

#### Checklist

### DIFF
--- a/libs/barista-components/chart/src/heatfield/chart-heatfield.html
+++ b/libs/barista-components/chart/src/heatfield/chart-heatfield.html
@@ -1,6 +1,7 @@
 <button
   class="dt-chart-heatfield-marker"
   [class.dt-chart-heatfield-active]="active"
+  [style.visibility]="_isValidStartEndRange ? 'visible' : 'hidden'"
   #marker="cdkOverlayOrigin"
   cdk-overlay-origin
   (click)="_toggleActive()"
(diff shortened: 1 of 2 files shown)


ffriedl89:
Could you maybe reword the commit to "fix(chart-heatfield): Heatfield is now hidden when start and end are undefined" - because this will end up in the changelog and might be confusing for people, because the fix does exactly the opposite of what you mention in the commit message ;)

sherifElhefnawyDynatrace:
Reworded commit :)

tomheller:
When the PR is merged, could the merging person please cherry pick the commits into the relevant branches (auto cherry picker does not work for forks)

libs/barista-components/chart/src/heatfield/chart-heatfield.html
```diff
@@ -1,6 +1,7 @@
 <button
   class="dt-chart-heatfield-marker"
   [class.dt-chart-heatfield-active]="active"
+  [style.visibility]="_isValidStartEndRange ? 'visible' : 'hidden'"
```
lukasholzer: Is it possible to click the invisible? Please check that :)
Maybe add an e2e test scenario for that

sherifElhefnawyDynatrace: Have already checked that, and here is a stackoverflow thread: https://stackoverflow.com/questions/5659563/css-is-a-hidden-object-clickable

sherifElhefnawyDynatrace: As for the e2e test, might add it later ya but not on this pr

PR #716, opened 2020-03-10
barista: Modifying overlay rule for avoiding conflicts with CDK's.

### LINKED ISSUE #715: [Event chart] Merged items aren't shown vertically in the overlay
# Bug Report

## Expected Behavior

In the **[Event chart](https://github.com/dynatrace-oss/barista/blob/master/libs/barista-components/event-chart)**, when there're 2 (or more) events merged, the overlay should show the items aligned vertically.

## Current Behavior

Currently the overlay is showing the events items horizontally.

## Steps to Reproduce

1. Go to User Sessions
2. Select a user tag which has merged events.
3. Hover , or click, over a merged event in the **event chart**

### Used Versions:

- **node**: v13.9.0
- **npm**: 6.13.7
- **typescript**: 3.7.5
- **angular**: 9.0.4
- **@angular/cdk**: 9.1.0
- **@dynatrace/barista-components**: 6.0.0

## Detailed Description

The current implementation of the css rule for the overlay component is getting a conflict with the Angular CDK one, so changing the way we defined the rule will make it have more priority.

## Possible Implementation

[Current Implementation](https://github.com/dynatrace-oss/barista/blob/master/libs/barista-components/event-chart/src/event-chart.scss#L165)

```scss
::ng-deep .cdk-overlay-pane {
  &.dt-event-chart-overlay-panel {
    @include dt-overlay-container();
  }
}
```

## Attachments

**_Current:_**
image

**_Expected:_**
image

pedromosquerat:

Fixes #715

#### Type of PR

#### Checklist

### DIFF
--- a/libs/barista-components/event-chart/src/event-chart.scss
+++ b/libs/barista-components/event-chart/src/event-chart.scss
@@ -162,6 +162,9 @@
 }
 
 /** Event chart overlay */
-::ng-deep .dt-event-chart-overlay-panel {
+/** The selector is combined with the cdk-overlay-pane to
+increase the specificity to avoid edge cases where the cdk styles
+are loaded after the component's styles in the angular.json */
+::ng-deep .cdk-overlay-pane.dt-event-chart-overlay-panel {
   @include dt-overlay-container();
 }


ffriedl89:
Hey @pedromosquerat thank you for the contribution, can you explain a little bit, which properties are problematic between what the cdk and our mixin provides? 
You should be able to provide your own grid inside the overlay to handle create the desired layout.

pedromosquerat:
> Hey @pedromosquerat thank you for the contribution, can you explain a little bit, which properties are problematic between what the cdk and our mixin provides?

Hi @ffriedl89,

The one that is causing the issue is the `display: flex;` coming from cdk. As we have it in ours (`display: inline-block`), works as we want

ffriedl89:
Ok, so I've done a little bit of digging and found out that the order in which the styles are defined in the angular.json file makes a difference. So if you load the styles from the components before the apps root scss file everything is fine, if you specify it the other way round - the cdk's property win. 

I guess this change makes sense and since it is experimental we don't have to do a breaking change here.

libs/barista-components/event-chart/src/event-chart.scss
```diff
@@ -162,6 +162,8 @@
 }
 
 /** Event chart overlay */
-::ng-deep .dt-event-chart-overlay-panel {
-  @include dt-overlay-container();
+::ng-deep .cdk-overlay-pane {
+  &.dt-event-chart-overlay-panel {
+    @include dt-overlay-container();
+  }
```
ffriedl89:
```suggestion
/** The selector is combined with the cdk-overlay-pane to increase the specificity to avoid edge cases where the cdk styles are loaded after the component's styles in the angular.json */
::ng-deep .cdk-overlay-pane.dt-event-chart-overlay-panel  {
    @include dt-overlay-container();
```
No need for nesting here - it makes it harder to read and parse selectors for developers :) Please see our coding guidelines https://github.com/dynatrace-oss/barista/blob/master/CODING_STANDARDS.md#scss-specifics

ffriedl89:
Additionally please sign the CLA so we can proceed after you've done the changes :) And thanks a lot for your contribution!

pedromosquerat:
Done! Thanks for your help and suggestions @ffriedl89

tomheller:
Dear PR merger, please make sure that you cherry pick the pull request to its respective targets, as the auto cherry picker sadly does not yet work for forked contributions.

PR #721, opened 2020-03-11
ci: Added a release promotion reminder

ffriedl89:
Added an action that triggers a webhook to post a message on slack after a release.

Since we don't want to promote barista instantly to production - this action is a reminder to do this after a release is done


ffriedl89:
Blocked by #722 resolving the security vulnerability

PR #724, opened 2020-03-11
barista: Remove flat white from the icon color wheel.

lara-aigmueller:

The newly introduced color-value `FLAT_WHITE` broke the layout of our icons color wheel in Barista. This filter should fix it.

#### Type of PR

Barista bugfix

#### Checklist

### DIFF
--- a/apps/barista-design-system/src/components/icon-color-wheel/icon-color-wheel.ts
+++ b/apps/barista-design-system/src/components/icon-color-wheel/icon-color-wheel.ts
@@ -68,6 +68,8 @@ export class BaIconColorWheel {
 
   constructor(private _sanitizer: DomSanitizer) {
     const groupedBlobs = Object.keys(DtColors)
+      /** breaking-change Can eb removed when flat_white is removed from the color list. 7.0.0 */
+      .filter(key => key !== 'FLAT_WHITE') // no flat white for the color wheel
       .map((key): BaColorWheelBlob | undefined => {
         /* tslint:disable no-magic-numbers */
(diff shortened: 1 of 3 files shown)


tomheller:
Security checks fail, because of the minimist dependency. This should be fixed by #722

apps/barista-design-system/src/components/icon-color-wheel/icon-color-wheel.ts
```diff
@@ -68,6 +68,7 @@ export class BaIconColorWheel {
 
   constructor(private _sanitizer: DomSanitizer) {
     const groupedBlobs = Object.keys(DtColors)
+      .filter(key => key !== 'FLAT_WHITE') // no flat white for the color wheel
```
lukasholzer: To be honest, I would remove the colour from the scss file :D

lukasholzer: And add a comment to the scss file that every added colour will be displayed in the colour wheel otherwise we will forget in the future.

ffriedl89: I agree with @lukasholzer here that this is an unnecessary risk. It was fun while it lasted :D Let's cleanup the color, remove the filter and add a comment in the colors to check the color wheel once something changes.

tomheller: Need to keep it in, because removing the color would be a breaking change.

Issue #465, opened 2020-01-21
Remove the RouterLink from the secondary-nav-section


NinaKammerlander:
# Feature Request

The secondary-nav-section should work without a router.

Currently, the `dt-secondary-nav-section` contains a switch within its template where it defines whether the generated link should be either an external link (which uses the href attribute on the rendered html), or an internal link (which uses the `routerLink` directive). 

Template reference from https://github.com/dynatrace-oss/barista/blob/master/components/secondary-nav/src/section/secondary-nav-section.html#L13-L22: 
```html
<a [href]="href" class="dt-secondary-nav-section-header" *ngIf="external">
    <ng-container *ngTemplateOutlet="content"></ng-container>
  </a>
  <a
    *ngIf="!external"
    [routerLink]="href"
    class="dt-secondary-nav-section-header"
  >
    <ng-container *ngTemplateOutlet="content"></ng-container>
  </a>
```

In general we are trying to avoid a hard dependency to the Angular Router, which is why we should think about removing this switch and let the consumer handle what linking mechanism they would like to choose.

NinaKammerlander:
@ffriedl89 can you please add more to the description? 😄

tomheller:
I have updated the description to be a bit more telling about what is supposed to be the outcome here.

thomaspink:
To resolve this we have to change the way the section works to something like:

````
<a dt-secondary-nav-section routerLink="/something">
    <dt-secondary-nav-section-title>
      Monitoring
    </dt-secondary-nav-section-title>
    <dt-secondary-nav-section-description>
      Setup and overview
    </dt-secondary-nav-section-description>
</a>
```

Internally the html of the section (expect for the template) has to move to the secondary nav itself (or an internal component).

thomaspink:
We have to change the api and the way how the section work for that. Please hit me up if someone starts working on this issue.

ffriedl89:
@areknow We talked about this a while ago, and we would like to get this going again :) Would you be able to help us out here? 
Any help is very much appreciated

areknow:
@ffriedl89 sorry for the delay, I will get the PR ready this week

tomheller:
Issue is still relevant and should not be closed. @areknow would you still be interested to provide the fix for this one?

areknow:
Hey @tomheller, I started the work and should be able to complete it next sprint 👍

PR #734, opened 2020-03-12
feat(secondary-nav): Remove routerlink from secondary nav section

areknow:

This is the requested modification for the DtSecondaryNav as described here: [https://github.com/dynatrace-oss/barista/issues/465](https://github.com/dynatrace-oss/barista/issues/465)

This is breaking change. The component will no longer rely on the router module. The section will be decoupled from the link entirely. 

Sections can be defined with routerlink or href like so: 
```
<dt-secondary-nav>
  <dt-secondary-nav-section>
    <dt-secondary-nav-section-title>Monitoring</dt-secondary-nav-section-title>
    <dt-secondary-nav-section-description>Setup and overview</dt-secondary-nav-section-description>
    <a dtSecondaryNavLink dtSecondaryNavLinkActive routerLink="/">Monitored technologies</a>
    <a dtSecondaryNavLink routerLink="/">Monitoring overview</a>
    <a dtSecondaryNavLink href="/">Host naming</a>
  </dt-secondary-nav-section>
</dt-secondary-nav>
```

And external links can be easily added to the root of the menu like so:
```
<dt-secondary-nav>
  <a dtSecondaryNavLink dtSecondaryNavLinkActive href="https://google.com">
    <dt-secondary-nav-section-title>Processes and containers</dt-secondary-nav-section-title>
    <dt-secondary-nav-section-description>Detection and naming</dt-secondary-nav-section-description>
  </a>
</dt-secondary-nav>
```

Because of this simplification, there were plenty of inputs removed from the `dt-secondary-nav-section`. Changes have been updated in the examples and readme.


thomaspink:
Hi @areknow , thanks for the pull request.

Please be aware that these changes should be considered breaking.

Please annotate your commit as BREAKING CHANGE and add some guidance what has changed:
Secondary nav does not depend on the router module any more. A new directive for setting the link as active has therefor been introduced.

Example:
```
feat: Removes Renderer2 to be compatible with ivy. 

Closes #518

BREAKING CHANGE: Components that are using the outdated Renderer2, which is removed in this commit, do this by injecting it via DI in their constructor. When removing the renderer constructor param the api changes which leads to a breaking change.
```

thomaspink:
I am wondering if you can also provide a description (maybe als a link to an external example as we still wanna avoid using router links in the examples lib) on how to set links active based on the router/url with the new directive. 
I guess as the dependency for the routermodule has been removed (which is great, thanks :) ) the consumer has to somehow bind the active directory to the router/routerlinkactive?

libs/barista-components/secondary-nav/src/secondary-nav.scss
```diff
@@ -207,5 +236,5 @@ $dt-secondarynav-header-padding-left: 20px;
 .dt-secondary-nav-section
   .dt-secondary-nav-section-header.dt-secondary-nav-section-expanded:hover
   .dt-secondary-nav-section-title {
-  color: $gray-700;
+  color: $gray-700; // ask ryan
```
tomheller: Let's ask Ryan or @dynatrace-oss/ux-core and figure out if this is the correct value.

libs/barista-components/secondary-nav/src/secondary-nav.scss
```diff
@@ -64,6 +93,18 @@ $dt-secondarynav-header-padding-left: 20px;
...
+.dt-secondary-nav-section-header:hover {
+  background: $gray-200;
+  text-decoration: none;
+}
+
+.dt-secondary-nav-section-header:hover .dt-secondary-nav-section-title {
+  color: $turquoise-700;
+}
+.dt-secondary-nav-section-header .dt-secondary-nav-section-title {
+  width: calc(100% - 52px);
```
tomheller: Could we quickly comment here, how the 52px come into existence. This is a very magic number.
It would be great if we could calculate this based on some sass variables (if this is calculatable). If not a comment will suffice.

libs/barista-components/secondary-nav/src/secondary-nav.scss
```diff
@@ -108,84 +149,72 @@ $dt-secondarynav-header-padding-left: 20px;
...
+  justify-content: center;
   text-decoration: none;
+  outline: none;
+  height: 56px;
+  font-size: 15px;
+  padding: 8px 0 8px 20px;
+  border-width: 1px 0 0;
+  border-style: solid;
+  border-color: $gray-200;
+  font-family: inherit;
```
tomheller: I guess you don't need to set this to inherit. What is changing your font familiy in the cascade?

libs/barista-components/secondary-nav/src/secondary-nav.scss
```diff
@@ -108,84 +149,72 @@ $dt-secondarynav-header-padding-left: 20px;
...
+  color: inherit;
```
tomheller: I guess there is no need to set this to inherit.

libs/barista-components/secondary-nav/src/secondary-nav.scss
```diff
@@ -108,84 +149,72 @@ $dt-secondarynav-header-padding-left: 20px;
...
-  font-size: 14px;
-  line-height: 30px;
-  display: block;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
   text-decoration: none;
+  outline: none;
+  height: 56px;
+  font-size: 15px;
```
tomheller: This seems like an oddly specific size. Could we think about streamlining the font size for this one, so it fits the other components as well. Maybe the default 14px will also be alright.

areknow:
@thomaspink @tomheller thank you for the feedback! 

When I started the work on this refactor, I tried to write it in order to keep the existing API working along side this new refactor in order to correctly deprecate the old way of using the component. Unfortunately I was not able to get the existing API and new API to exist simultaneously. Maybe this is something that someone else is able to solve, but I was not able to.

I will work on the SCSS issues identified 👍 
I will add the commit message annotation 👍

@thomaspink should I replace all `routerLink` references with `href`? And how should I create an external example before the modified component is published? Consumers can easily use routerLinkActive in order to set the dtSecondaryNavLinkActive to true :)

libs/barista-components/secondary-nav/src/secondary-nav.scss
areknow: fixed, wrong font size was being used

libs/barista-components/secondary-nav/src/secondary-nav.scss
areknow: fixed, rule was not needed

libs/barista-components/secondary-nav/src/secondary-nav.scss
areknow: fixed, rule was not needed

libs/barista-components/secondary-nav/src/secondary-nav.scss
areknow: Screen Shot 2020-03-13 at 2 37 21 PM

The 52px = icon width + padding. I will add a comment unless you know of a sass var for the icon width?

libs/barista-components/secondary-nav/src/secondary-nav.scss
areknow: Removed comment, verified color matches value in invision mockup

thomaspink:
> @thomaspink should I replace all routerLink references with href? 

Yes, i would suggest so

> And how should I create an external example before the modified component is published? Consumers can easily use routerLinkActive in order to set the dtSecondaryNavLinkActive to true :)

I am currently thinking of two ways we could show the consumer how the integration with the router can work.

1. Leave all examples with just `href` attributes and add a section to the documentation on how to use `routerLink` and `routerLinkActive` in conjunction with the secondary nav.
2. Create one example that uses the `routerLink` and `routerLinkActive`. But instead of using these directives form the angular/router module we just apply them as plain html attributes (they won't get parsed via angular on barista).

areknow:
Commenting to keep the ticket from going stale. Work is still being done here. Thanks

tomheller:
Thank you @areknow, we have increased the stale timeout of our issues to 90 days, since 30 felt a little short after the initial trial period.

lukasholzer:
Please rebase your branch with the latest master, then perform and install and run `npm run format --all` to get the latest formatting changes that were introduced with the version upgrade of prettier to 2.0. This update was necessary to be compatible with the latest typescript version that we are using. Furthermore, we improved the angular template formatting!

areknow:
Hey team, I have hit a small road block and I would appreciate some assistance.
The component refactor is complete and the new API works well. I have tested it in 2 applications and have had feedback from another dev that they think the new API is a big improvement. 

The issue I'm facing is regarding `routerLinkActive` directive. When the RLA property is used to trigger an active state, the component does not react to it without using a setTimeout to defer the the cycle. https://github.com/dynatrace-oss/barista/pull/734/commits/b3d1af0950012371b128a8faf9019d733c580579

You can test the behavior by pulling the branch (recently rebased), and commenting out or enabling the setTimeout on line 113 of the secondary-nav-section. I have also set up an example in the secondary-nav-demo-component.html that uses the RLA property to demonstrate the issue. https://github.com/dynatrace-oss/barista/pull/734/commits/696fe98c11181f5933c120d3b020867a0f7185d3#diff-a84e50f4e759c9e3a9a7b406c003de89R90

With the setTimeout enabled, you can see that the section reacts correctly to the RLA active property.
Without the setTimeout enabled, you will see that the section does not react to RLA active and will not set the section to the active+expanded state.

Is it good practice to leave this setTimeout zero? I have seen in other components the use of a `Promise.resolve().then(() => {}` to solve the same issue, but this solution does not work in my case and I am not sure why.

Thanks!

tomheller:
Hi @areknow!
I will have a look at this as soon as I can. I hope we can solve this problem without the timeout or promise resolve, tbh.

areknow:
That’s great news @tomheller!

Please feel free to merge your changes into the branch 👍

areknow:
@tomheller I just tested the change you included and unfortunately they problem has not been resolved. 

The section that has the RLA property set to true is not being expanded and set to active:
Screen Shot 2020-05-06 at 8 14 59 AM

After clicking on it:
Screen Shot 2020-05-06 at 8 15 09 AM

What I would expect to see when the component reacts properly to the RLA variable (as seen when using setTimeout zero):
Screen Shot 2020-05-06 at 8 16 54 AM

Let me know if you would like to meet and discuss further! Thank you for your help 😄

tomheller:
@areknow, I've added a switchMap to the pipe, which waits for the next zoneStable. This should wait until all macroTasks are cleared (also wait for the routerLink evaluation in the next tick). 

I added two tests that check this behavior and took  the liberty of simplifying your test setup a little bit to remove code duplication. 

The question is if the component should also listen for changes in the activeLinks and react to these, as currently it is only evaluated at creation and if the queryList changes. And it can only expand, not close itself back down. But I guess that is something that we can figure out later as well.

areknow:
@tomheller this works great! Thanks for the commit.
Would it be equivalent to: 
```
  ngAfterContentInit(): void {
    this._ngZone.onStable
      .pipe(startWith(null), takeUntil(this._destroy$))
      .subscribe(() => {
        // Check if section has any active links, and activate + expand the entire section
        if (
          this._activeLinks.some(
            (activeLink) => activeLink.dtSecondaryNavLinkActive,
          )
        ) {
          this._activateAndExpandSection();
        }
        // Check if section has any groups set to active, and activate + expand the entire section
        if (this._groups.some((group) => group.active)) {
          this._activateAndExpandSection();
        }
      });
  }
```
Is it necessary to use the switch map in conjunction with the changes from this_activeLinks ?
Thanks!

areknow:
Unfortunately I found another issue 😅 
If you use the global menu to navigate to another component demo page, then navigate back to the secondary-nav page, the RLA property will not activate+expand the menu as it does when the initialization happens

tomheller:
> Subsequent clicks on the page will eventually trigger the ngZone subscription to fire and the component will realize that the section should be open.

This is a tough one. We might need to add an output to the secondaryNavLink that fires when it's active state changes. We would need to group these outputs in the querylist.changes subscription and react to them there. 

Because it seems the input change on the secondaryNavLink does not trigger the changes in the query list?

areknow:
@tomheller I finished updating the files per our discussion from today. I have also updated the documentation and the examples

libs/barista-components/secondary-nav/src/section/secondary-nav-group.ts
```diff
@@ -35,7 +39,20 @@ import {
...
   encapsulation: ViewEncapsulation.Emulated,
 })
-export class DtSecondaryNavGroup {
-  /** The text value of the group header */
+export class DtSecondaryNavGroup implements AfterContentInit {
+  /** The text value of the group header. */
   @Input() label: string;
+
+  /** Whether or not the group is active. */
+  active = false;
```
tomheller, line 47: I think this could be an internal property.

libs/barista-components/secondary-nav/src/section/secondary-nav-group.ts
```diff
@@ -35,7 +39,20 @@ import {
...
+
+  /** @internal Active link directive within the group. */
+  @ContentChildren(DtSecondaryNavLinkActive)
+  private _links!: QueryList<DtSecondaryNavLinkActive>;
```
tomheller, line 51: I don't think we will need this.
```suggestion
  private _links: QueryList<DtSecondaryNavLinkActive>;
```

libs/barista-components/secondary-nav/src/section/secondary-nav-group.ts
```diff
@@ -35,7 +39,20 @@ import {
...
+
+  ngAfterContentInit(): void {
+    if (this._links.find((link) => link.dtSecondaryNavLinkActive)) {
+      this.active = true;
+    }
+  }
```
tomheller, line 57: We may need to think about reacting to changes of the links here, but for now this will be fine. 
I guess there will be a follow up story, where some tests will be added that will trip us up here, if the group does not get destroyed, but the links are dynamically added or removed.

libs/barista-components/secondary-nav/src/section/secondary-nav-section.ts
```diff
@@ -78,40 +78,21 @@ export class DtSecondaryNavSection {
...
-    return this._active;
-  }
-  set active(value: boolean) {
-    this._active = coerceBooleanProperty(value);
-    if (this._active && this._expandable) {
-      this.expanded = true;
-    }
-  }
-  private _active = false;
+  /** @internal List of all groups within the section. */
```
tomheller, line 90: If the property is private there is no need to annotate it as internal.
```suggestion
  /** List of all groups within the section. */
```

libs/barista-components/secondary-nav/src/section/secondary-nav-section.ts
```diff
@@ -131,6 +112,42 @@ export class DtSecondaryNavSection {
...
+    this._destroy$.next();
+    this._destroy$.complete();
+  }
+
+  /** @internal Check if sections has any active links or groups with active links within them */
+  _checkForActiveStates(): void {
+    // Check if section has any active links, and activate + expand the entire section
+    // Includes a setTimeout((), 0) because the RouterLinkActive directive uses Promise.resolve().then(())
+    // https://github.com/angular/angular/blob/master/packages/router/src/directives/router_link_active.ts#L125
+    setTimeout(() => {
```
tomheller, line 129: Nice, thanks for the extra comment here. This will make it easier to understand in the future, why this was necessary.

areknow:
FYI I noticed that when running `npm run barista`, two files get modified which gets the source files out of sync with what the check-formatting stage is expecting. This then requires an additional `nx format:write`. If this has already been reported, please ignore 🙂

Screen Shot 2020-05-12 at 11 19 36 AM

areknow:
Everything looks good except for the ZEIT deploy

```
 ZEIT Now Deployment
##[error]The process 'npx' failed with exit code 1
Run amondnet/now-deployment@v1
/usr/bin/docker exec  cefe41df38e7e6411563d02a1162ce8b11c802b721292c5eba362efcbf8098d7 sh -c "cat /etc/*release | grep ^ID"
Container distribution is alpine. Running JavaScript Action with external tool: node12_alpine
npx now -t  -m githubCommitSha=f61132524ef2eb9fea36dca6344f32fa78550e15 -m githubCommitAuthorName=amond -m githubCommitAuthorLogin=amondnet -m githubDeployment=1 -m githubOrg=dynatrace-oss -m githubRepo=barista -m githubCommitOrg=dynatrace-oss -m githubCommitRepo=barista -m githubCommitMessage=Merge 1be462c23fda423a97ae1c8d2a93533d09a3005e into f7ed0baea6bd46b0e0341d7b3c1a5d605570a925
Now CLI 19.0.1

Now CLI 19.0.1
Error! No existing credentials found. Please run `now login` or pass "--token"
> More details: https://err.sh/now/no-credentials-found

Error! No existing credentials found. Please run `now login` or pass "--token"
> More details: https://err.sh/now/no-credentials-found
##[error]The process 'npx' failed with exit code 1

lukasholzer:
> Everything looks good except for the ZEIT deploy
> 
> ```
>  ZEIT Now Deployment
> ##[error]The process 'npx' failed with exit code 1
> Run amondnet/now-deployment@v1
> /usr/bin/docker exec  cefe41df38e7e6411563d02a1162ce8b11c802b721292c5eba362efcbf8098d7 sh -c "cat /etc/*release | grep ^ID"
> Container distribution is alpine. Running JavaScript Action with external tool: node12_alpine
> npx now -t  -m githubCommitSha=f61132524ef2eb9fea36dca6344f32fa78550e15 -m githubCommitAuthorName=amond -m githubCommitAuthorLogin=amondnet -m githubDeployment=1 -m githubOrg=dynatrace-oss -m githubRepo=barista -m githubCommitOrg=dynatrace-oss -m githubCommitRepo=barista -m githubCommitMessage=Merge 1be462c23fda423a97ae1c8d2a93533d09a3005e into f7ed0baea6bd46b0e0341d7b3c1a5d605570a925
> Now CLI 19.0.1
> 
> Now CLI 19.0.1
> Error! No existing credentials found. Please run `now login` or pass "--token"
> > More details: https://err.sh/now/no-credentials-found
> 
> Error! No existing credentials found. Please run `now login` or pass "--token"
> > More details: https://err.sh/now/no-credentials-found
> ##[error]The process 'npx' failed with exit code 1
> ```

Yea that is fine currently our preview builds are only working for maintainer in the library. So on a fork pull request, GitHub does not allow to pass the credentials to this workflow and that is the reason why our zeit deployment does not work.

As we cannot leak our credential with hardcoding it, we don't have previews currently for forked pull requests.

This is a known problem. Sorry for the inconvenience.

tomheller:
> FYI I noticed that when running `npm run barista`, two files get modified which gets the source files out of sync with what the check-formatting stage is expecting. This then requires an additional `nx format:write`. If this has already been reported, please ignore 🙂

@areknow  This is fine, they get generated if examples are being generated or changed. These should be checked in. It states it in the files itself that these are generated and should be checked in.

areknow:
@ffriedl89 thank you for simplifying the code logic around setting a section to active 👍

ffriedl89:
FYI: I've removed the RouterModule from the secondary nav module and force pushed so i can merge the PR right away :)

areknow:
@ffriedl89 I forgot to remove the router module 😂  thanks!

Issue #733, opened 2020-03-12
Sidenav height bug when content is smaller than the view


rowa-audil:
The sidenav component doesn't fill the whole page when the content inside is smaller than the view.
Possible solution: height: 1vh or 100% should solve the issue.

PR #735, opened 2020-03-13
Fixes SideNav height bug when its content is smaller than view

rowa-audil:
Changes SideNav Height property from 100% to 100vh

### DIFF
--- a/libs/barista-components/drawer/src/sidenav.scss
+++ b/libs/barista-components/drawer/src/sidenav.scss
@@ -5,7 +5,7 @@
   will-change: transform;
   background-color: $gray-700;
   min-width: 240px;
-  height: 100%;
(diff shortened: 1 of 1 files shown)


lukasholzer:
Is there an issue connected with this PR?

lukasholzer:
Maybe you can reword your commit to something more descriptive for the changelog:

The scope would be drawer in case you import it from `@dynatrace/barista-components/drawer`

`fix(drawer): Fixes an issue where the side navigation drawer was not expanding to the full-screen height.`

rowa-audil:
Yes: https://github.com/dynatrace-oss/barista/issues/733

lukasholzer:
please add issue number in commit body.

Issue #77, opened 2019-11-12
Copy to clipboard (Components): Remove own version of clipboard api access move to cdk


ffriedl89:
Remove copy to clipboard API from components and move to CDK

ffriedl89:
Needs upgrade to version 9 before #566

PR #746, opened 2020-03-16
Copy to clipboard: move to CDK instead of current API

rowa-audil:
Moves copy to clipboard component to Angular's CDK.

Closes. #77

### DIFF
--- a/libs/barista-components/copy-to-clipboard/src/copy-to-clipboard.ts
+++ b/libs/barista-components/copy-to-clipboard/src/copy-to-clipboard.ts
@@ -28,6 +28,7 @@ import {
   ViewEncapsulation,
   Input,
 } from '@angular/core';
+import { Clipboard } from '@angular/cdk/clipboard';
 import { Subscription, timer } from 'rxjs';
 
 import {
@@ -53,7 +54,10 @@ const DT_COPY_TO_CLIPBOARD_SUCCESSFUL = 'dt-copy-to-clipboard-successful';
   encapsulation: ViewEncapsulation.Emulated,
 })
 export class DtCopyToClipboard implements AfterContentInit, OnDestroy {
-  constructor(private _cd: ChangeDetectorRef) {}
+  constructor(
+    private _changeDetectorRef: ChangeDetectorRef,
+    private _cdkClipboard: Clipboard,
+  ) {}
 
   /** Defines the button variant of the copy button. */
   @Input() variant: ButtonVariant = 'primary';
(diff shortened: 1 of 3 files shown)


libs/barista-components/copy-to-clipboard/src/copy-to-clipboard.ts
```diff
@@ -53,7 +54,10 @@ const DT_COPY_TO_CLIPBOARD_SUCCESSFUL = 'dt-copy-to-clipboard-successful';
   encapsulation: ViewEncapsulation.Emulated,
 })
 export class DtCopyToClipboard implements AfterContentInit, OnDestroy {
-  constructor(private _cd: ChangeDetectorRef) {}
+  constructor(
+    private _cd: ChangeDetectorRef,
```
tomheller:
```suggestion
    private _changeDetectorRef: ChangeDetectorRef,
```

I know that you did not change this one, but as long as we are touching this file, I feel like we could align this one to our coding guidelines.

libs/barista-components/copy-to-clipboard/src/copy-to-clipboard.ts
```diff
@@ -89,26 +93,25 @@ export class DtCopyToClipboard implements AfterContentInit, OnDestroy {
...
-        );
-        this._copyButton.nativeElement.focus();
-      }
+      this._input.nativeElement.textContent !== null
+        ? this.cdkClipboard.copy(this._input.nativeElement.textContent)
+        : this.copyFailed.emit();
+    }
+  }
+
+  copiedToClipboard(): void {
```
tomheller: does this one need to be public?

libs/barista-components/copy-to-clipboard/src/copy-to-clipboard.ts
```diff
@@ -89,26 +93,25 @@ export class DtCopyToClipboard implements AfterContentInit, OnDestroy {
   /** Copies the provided content to the clipboard. */
   copyToClipboard(): void {
     if (this._input) {
```
lara-aigmueller: Should we also emit the `copyFailed` event when there is no `_input`, i.e. in the else-branch?

rowa-audil: Yes, thanks for pointing it out.

libs/barista-components/copy-to-clipboard/src/copy-to-clipboard.ts
rowa-audil: Its called by the CDK when it finished copying, therefor I'll annotate it with @internal in the new commit.

tomheller:
Can you please squash the two commits into one?

rowa-audil:
Squashed.

lukasholzer:
please rebase to get the security fixes

libs/barista-components/copy-to-clipboard/src/copy-to-clipboard.ts
```diff
@@ -88,27 +92,27 @@ export class DtCopyToClipboard implements AfterContentInit, OnDestroy {
...
-      this._showIcon = true;
-      _addCssClass(this._input.nativeElement, DT_COPY_TO_CLIPBOARD_SUCCESSFUL);
-      if (this._copyButton) {
-        _addCssClass(
-          this._copyButton.nativeElement,
-          DT_COPY_TO_CLIPBOARD_SUCCESSFUL,
-        );
-        this._copyButton.nativeElement.focus();
-      }
+    if (this._input && this._input.nativeElement.textContent !== null) {
```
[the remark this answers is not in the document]
tomheller: I would consider this a NIT, to be honest.

ffriedl89:
Is this really a feature? I'd argue this is more a refactor since the consumer does not care if we use the copy to clipboard internally or do it by ourselves. @dynatrace-oss/barista  What do you think? 

I'd suggest rewording the commit to refactor instead of feat and merge it afterwards

PR #761, opened 2020-03-19
barista: Use images instead of inline svgs for icons.

### LINKED ISSUE #750: Barista: Extract inline SVGs from the template to assets
# Feature Request

Currently, there are a lot of SVGs for example in the footer inlined in the template. We should extract them because the browser can cache them. Otherwise, we have to download it on every page download.

Furthermore, it would reduce our main bundle size, because the ivy renderer is creating javascript code that generates the HTML and this code is way larger than the original HTML that is needed for the SVG.

Last but not least it would keep our code more readable.

lara-aigmueller:

closes #750 

Updated the GitHub link to dynatrace-oss

#### Type of PR

Barista improvement


lukasholzer:
thx great but id does not close the issue to 100% it is a part of it there are so many more svgs like the miniature icons on the tiles etc.

lara-aigmueller:
> thx great but id does not close the issue to 100% it is a part of it there are so many more svgs like the miniature icons on the tiles etc.

Alright, I added a wip-label and will search for all the missing svgs and add it to this PR.

lara-aigmueller:
I did not move the `ba-tile-status-icon-triangle` to an svg-asset, because as inlined svg it's much easier to theme it according to the page-theme (and it's used on all overview-pages).

ffriedl89:
took care of the cherry pick

PR #778, opened 2020-03-23
Build/create docker build image

lukasholzer:

Create a workspace build image that can be used for faster CI build times on Jenkins for barista.

In a follow up it can be used on the circleci to skip installing on every build, and use our custom image for building like:

```yml
  build-test:
    working_directory: /dynatrace
    docker:
      - image: designops/workspace-base:latest
    steps:
      - attach_workspace:
          at: /dynatrace
      - run: ./node_modules/.bin/nx workspace-lint
      - run: yarn affected:lint --all --parallel
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,20 @@
-node_modules
-.git
-.github
-.circleci
-.vscode
+#--------------------------
+# disallow anything
+#--------------------------
+**
+
+#--------------------------
+# allow explicit
+#--------------------------
+!package.json
+!package-lock.json
+
+# Used by multiple projects:
+# - /Dockerfile
+!tsconfig.json
+!angular.json
+
+
+# Used by: /Dockerfile
+!/libs/tools/builders
+!/entrypoint.sh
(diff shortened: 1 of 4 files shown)


.dockerignore
```diff
@@ -1,5 +1,20 @@
-node_modules
-.git
-.github
-.circleci
-.vscode
+#--------------------------
```
tomheller: this change might be problematic for these github actions, like this one: 
https://github.com/dynatrace-oss/barista/blob/master/.github/workflows/cherry-picker.yml
https://github.com/dynatrace-oss/barista/tree/master/tools/cherry-picker

Can we check if they work first?

lukasholzer: they have their own `.dockerignore` files. So I thought it should be safe but how can we test them?

lukasholzer: I will add some merge-ready label and wait if the pull request labeller is working. 

Consider the label `merge-ready` as **not merge ready** :D

lukasholzer: image

The pull request labeller worked fine and it has the same configuration as the cherry-picked.  So everything is working! 👍

PR #791, opened 2020-03-25
barista(absolute-link-transformer): The added transformer converts in…

### LINKED ISSUE #752: Barista: Replace internal links* with router links
Currently links that are part of the page content (coming from Markdown files) are `a`-tags with `href`-attributes. For Barista-internal links, `routerLink` should be used instead of `href`.

*it's not about "internal links" to internal pages/resources but links that go from one Barista page to another ;-)

rowa-audil:
This PR adds a transformer that converts internal 'href' links to 'routerLink' links.

Fixes #752

### DIFF
--- a/apps/barista-design-system/src/components/components.module.ts
+++ b/apps/barista-design-system/src/components/components.module.ts
@@ -29,6 +29,7 @@ import { BaColorGrid } from './color-grid/color-grid';
 import { BaLayoutGrid } from './layout-grid/layout-grid';
 import { BaColor } from './color-component/color';
 import { BaLayoutGridItem } from './layout-grid/layout-grid-item';
+import { BaContentLink } from './content-link/content-link';
 
 /**
  * The order of the following components is relevant in case they are nested.
@@ -40,6 +41,7 @@ export const BA_CONTENT_COMPONENTS: any[] = [
   BaLiveExample,
   BaHeadlineLink,
   BaColor,
+  BaContentLink,
   BaLayoutGridItem,
   BaLayoutGrid,
   BaColorGrid,
--- a/apps/barista-design-system/src/components/content-link/content-link.ts
+++ b/apps/barista-design-system/src/components/content-link/content-link.ts
@@ -1,0 +1,63 @@
+/**
+ * @license
+ * Copyright 2020 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Component, Input } from '@angular/core';
+import { Router } from '@angular/router';
+
+/**
+ * The ba-content link component is used because we need to dynamically
+ * instanciate a router link and this is not possible, because it is a directive.
+ */
+@Component({
+  selector: 'a[contentLink]',
+  template: '<ng-content></ng-content>',
+  styles: ['a:hover { cursor: pointer; }'],
+  host: {
+    '[href]': 'contentLink',
+    '[attr.contentlink]': 'contentLink',
+    '(click)': 'onClick($event)',
+  },
+})
+export class BaContentLink {
+  /** Absolute url for navigation on the page. For example /components/button */
+  @Input() contentLink: string;
+
+  /** QueryParams of absolute url */
+  @Input() queryParams: { [key: string]: string };
+
+  /** Fragment of absolute url */
+  @Input() fragment: string;
+
+  constructor(private _router: Router) {}
+
+  onClick(event: MouseEvent): boolean {
+    // Check if the click happens on a button or with some control keys pressed and let the
+    // click event bubble to let consumers or the browser handle the event.
+    if (
+      event.button !== 0 ||
+      event.ctrlKey ||
+      event.metaKey ||
+      event.shiftKey
+    ) {
+      return true;
+    }
+    this._router.navigate([this.contentLink], {
+      fragment: this.fragment,
+      queryParams: this.queryParams,
+    });
+    return false;
+  }
+}
--- a/apps/barista-design-system/src/pages/single-page/page-content.ts
+++ b/apps/barista-design-system/src/pages/single-page/page-content.ts
@@ -72,9 +72,13 @@ export class BaPageContent implements OnDestroy {
       const placeholderElements: HTMLElement[] = Array.from(
         this._elementRef.nativeElement.querySelectorAll(factory.selector),
       );
+
       for (const el of placeholderElements) {
+        const children = [].slice.call(el.childNodes);
         this._componentRefs.push(
-          createComponent(factory, this._viewContainerRef, this._injector, el),
+          createComponent(factory, this._viewContainerRef, this._injector, el, [
+            ...children,
+          ]),
         );
       }
     }
--- a/libs/tools/barista/src/transform.spec.ts
+++ b/libs/tools/barista/src/transform.spec.ts
@@ -15,12 +15,14 @@
  */
 
 import { BaPageLayoutType } from '@dynatrace/shared/barista-definitions';
+
 import {
   componentTagsTransformer,
   extractH1ToTitleTransformer,
   internalLinksTransformerFactory,
   headingIdTransformer,
   internalContentTransformerFactory,
+  relativeUrlTransformer,
 } from './transform';
 
 describe('Barista transformers', () => {
@@ -183,6 +185,52 @@ describe('Barista transformers', () => {
       });
       expect(transformed.content).toBe(
         '<h1>Fonts</h1>\n\n<p>\n\nThis is just a simple internal string\n\n</p>\n\n<p>Some more public content.</p>',
+      );
+    });
+  });
+
+  describe('internalLinkReplacer', () => {
+    it('should replace not fully qualified links with content-link component', async () => {
+      const content = `<a href="http://www.dynatrace.com">Dynatrace</a>
+        <a href="ftpt://resources/guides">Resource</a>
+        <a href="smtp://resources/guides">Resource</a>
+        <a href="smthn://resources/guides">Resource</a>
+        <a href="lorem://resources/guides">Resource</a>
+        <a href="//resources/guides">Resource</a>
+        <a href="resources/guides">Resource</a>
+        <a href="/resources/bundle">Resource</a>
+        <a href="/brand/guides#headline1">Resource</a>
+        <a href="#headline1">Resource</a>
+        <a href="#submitting-a-pull-request">submit a pull request</a>
+        <a href="/patterns/button-alignment?sort=ASC">Resource</a>
+        <a href="/patterns/button-alignment?sort-order=ASC">Resource</a>
+        <a href="/components/button?sort=ASC#headline1">Resource</a>
+        <a href="/patterns/button-alignment?sort=ASC&id=2">Resource</a>
+        <a href="../">Back</a>
+      `;
+      const transformed = await relativeUrlTransformer({
+        title: 'Links',
+        layout: BaPageLayoutType.Default,
+        content,
+      });
+      expect(transformed.content).toBe(
+        `<a href="http://www.dynatrace.com">Dynatrace</a>
+        <a href="ftpt://resources/guides">Resource</a>
+        <a href="smtp://resources/guides">Resource</a>
+        <a href="smthn://resources/guides">Resource</a>
+        <a href="lorem://resources/guides">Resource</a>
+        <a href="//resources/guides">Resource</a>
+        <a contentLink="resources/guides">Resource</a>
+        <a contentLink="/resources/bundle">Resource</a>
+        <a contentLink="/brand/guides" fragment="headline1">Resource</a>
+        <a contentLink="/" fragment="headline1">Resource</a>
+        <a contentLink="/" fragment="submitting-a-pull-request">submit a pull request</a>
+        <a contentLink="/patterns/button-alignment" queryParams="{sort: ASC}">Resource</a>
+        <a contentLink="/patterns/button-alignment" queryParams="{sort-order: ASC}">Resource</a>
+        <a contentLink="/components/button" fragment="headline1" queryParams="{sort: ASC}">Resource</a>
+        <a contentLink="/patterns/button-alignment" queryParams="{sort: ASC,id: 2}">Resource</a>
+        <a contentLink="../">Back</a>
+      `,
       );
     });
   });
--- a/libs/tools/barista/src/transform.ts
+++ b/libs/tools/barista/src/transform.ts
@@ -25,6 +25,7 @@ import { isPublicBuild } from '@dynatrace/tools/shared';
 import { baElementBlockIgnore } from './markdown-custom-elements-ignore';
 import { fetchContentList } from './utils/fetch-strapi-content';
 import { runWithCheerio } from './utils/run-with-cheerio';
+import { parse } from 'url';
 
 import {
   BaStrapiContentType,
@@ -202,6 +203,41 @@ export const copyHeadlineTransformer: BaPageTransformer = async source => {
   return transformed;
 };
 
+/** Replaces absolute href with contentLink */
+export const relativeUrlTransformer: BaPageTransformer = async source => {
+  const transformed = { ...source };
+  if (source.content && source.content.length) {
+    transformed.content = runWithCheerio(source.content, $ => {
+      const links = $('a');
+      links.each((_, link) => {
+        const linkValue = $(link).attr('href');
+        if (linkValue && !isQualifiedLink(linkValue)) {
+          let url = parse(linkValue);
+          // Link Value
+          $(link.attribs).append(
+            $(link)
+              .removeAttr('href')
+              .attr('contentLink', url.pathname || '/'),
+          );
+          // Fragment
+          if (url.hash) {
+            $(link.attribs).append(
+              $(link).attr('fragment', url.hash.replace('#', '')),
+            );
+          }
+          // QueryParam
+          if (url.query) {
+            $(link.attribs).append(
+              $(link).attr('queryParams', toQueryParamValue(url.query)),
+            );
+          }
+        }
+      });
+    });
+  }
+  return transformed;
+};
+
 /** Removes internal links from the content on public build. */
 export function internalLinksTransformerFactory(
   isPublic: boolean,
@@ -299,4 +335,23 @@ export function internalContentTransformerFactory(
 
     return transformed;
   };
+}
+
+/**
+ * Checks whether a URL is a fully qualified link
+ */
+function isQualifiedLink(href: string): boolean {
+  // Matches every character case insensitive until a colon is coming. Has to have at least one character.
+  // https://regex101.com/r/nzwuR4/1
+  return !!href.match(/^(?:[a-z]+:)?\/\//i);
+}
+
+/** Converts query parameters to a json object */
+function toQueryParamValue(query: string): string {
+  const params = new URLSearchParams(query);
+  const queryParams: string[] = [];
+  params.forEach((value, key) => {
+    queryParams.push(`${key}: ${value}`);
+  });
+  return `{${queryParams.join(',')}}`;
 }
--- a/apps/barista-design-system/src/shared/services/toc.service.ts
+++ b/apps/barista-design-system/src/shared/services/toc.service.ts
@@ -114,7 +114,6 @@ export class BaTocService implements OnDestroy {
 
     for (const headline of headlines) {
       const item = this._createEntry(headline, docId, idMap, []);
-
       if (item.level === 'h2') {
         toc.push(item);
       } else if (item.level === 'h3') {
--- a/apps/barista-design-system/src/utils/create-component.ts
+++ b/apps/barista-design-system/src/utils/create-component.ts
@@ -27,17 +27,18 @@ import {
  * Per default the placeholder element is then removed.
  */
 export function createComponent(
-  // tslint:disable-next-line:no-any
   factory: ComponentFactory<any>,
   viewContainerRef: ViewContainerRef,
   injector: Injector,
   placeholderElement: Element,
+  projectableNodes: any[][] = [],
   shouldRemovePlaceholder: boolean = true,
 ): ComponentRef<unknown> {
   const componentRef = viewContainerRef.createComponent(
     factory,
     viewContainerRef.length,
     injector,
+    [projectableNodes],
   );
 
   // At this point the component has been instantiated,
--- a/libs/tools/barista/src/builder/components.ts
+++ b/libs/tools/barista/src/builder/components.ts
@@ -35,6 +35,7 @@ import {
   uxSnippetTransformer,
   headingIdTransformer,
   copyHeadlineTransformer,
+  relativeUrlTransformer,
 } from '../transform';
 
 import { slugify } from '../utils/slugify';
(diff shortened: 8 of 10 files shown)


libs/tools/barista/src/transform.ts
```diff
@@ -202,6 +202,27 @@ export const copyHeadlineTransformer: BaPageTransformer = async source => {
...
 
+export const internalLinksToRouterLinksTransformer: BaPageTransformer = async source => {
+  const transformed = { ...source };
+  if (source.content && source.content.length) {
+    transformed.content = runWithCheerio(source.content, $ => {
+      const links = $('a');
+      if (links.length) {
+        links.each((_, link) => {
+          const linkValue = $(link).attr('href');
+          if (linkValue && !linkValue.includes('http')) {
```
lara-aigmueller: I would update the check here. An internal link could always include `http` somewhere. Either change it to `.startsWith('http')` or check if the link starts with `/` followed by a letter, then it should be an internal link. Or maybe check if the linkValue contains `://`, then it's also an absolute URL and probably not linking to another internal page.

rowa-audil: Exactly what I needed to know. Thank you

libs/tools/barista/src/transform.ts
```diff
@@ -202,6 +202,31 @@ export const copyHeadlineTransformer: BaPageTransformer = async source => {
...
+  const transformed = { ...source };
+  if (source.content && source.content.length) {
+    transformed.content = runWithCheerio(source.content, $ => {
+      const links = $('a');
+      if (links.length) {
+        links.each((_, link) => {
+          const linkValue = $(link).attr('href');
+          if (
+            linkValue &&
+            (!linkValue.startsWith('http') || !linkValue.includes('//:'))
```
lara-aigmueller:
```suggestion
            (!linkValue.startsWith('http') || !linkValue.includes('://'))
```

libs/tools/barista/src/transform.spec.ts
```diff
@@ -186,4 +187,38 @@ describe('Barista transformers', () => {
...
+        <a href="http://www.facebook.at">Facebook</a>
+        <a href="http://resources/guides">Guides</a>
+      </div>
+      `;
+      const transformed = await internalLinksToRouterLinksTransformer({
+        title: 'Links',
+        layout: BaPageLayoutType.Default,
+        content,
+      });
+      // console.log(transformContent)
```
lara-aigmueller: please remove

libs/tools/barista/src/transform.ts
ffriedl89: This check is rather complicated to read - I think what we want to do is check whether a link is relative or an absolute URL. 

I'd create a function that is called: isAbsoluteUrl(url: string): boolean

see https://stackblitz.com/edit/typescript-em8xdt

And we can then use this function in this expression

libs/tools/barista/src/transform.ts
```diff
@@ -300,3 +323,16 @@ export function internalContentTransformerFactory(
     return transformed;
   };
 }
+
+// Matches links to check if link is an internallink
+function checkInternalLink($: CheerioStatic, link: CheerioElement): boolean {
```
ffriedl89: rename 
```suggestion
function isInternalLink(href: string): boolean {
```

libs/tools/barista/src/transform.ts
```diff
@@ -202,6 +202,29 @@ export const copyHeadlineTransformer: BaPageTransformer = async source => {
...
 
+/** Replaces internal href links with routerLink links */
+export const internalLinksToRouterLinksTransformer: BaPageTransformer = async source => {
+  const transformed = { ...source };
+  if (source.content && source.content.length) {
+    transformed.content = runWithCheerio(source.content, $ => {
+      const links = $('a');
+      if (links.length) {
+        links.each((_, link) => {
+          const linkValue = $(link).attr('href');
```
ffriedl89: You already have a string here :)

libs/tools/barista/src/transform.ts
```diff
@@ -300,3 +323,16 @@ export function internalContentTransformerFactory(
...
+function checkInternalLink($: CheerioStatic, link: CheerioElement): boolean {
+  if (
+    $(link).attr('href') &&
+    !$(link)
+      .attr('href')!
+      .match(/^(?!\/\/)$^\/[a-z]|^[a-z]/g)
+  ) {
+    return true;
+  }
+  return false;
```
ffriedl89:
```suggestion
    return !href.match(/^(?!\/\/)$^\/[a-z]|^[a-z]/g)
```

ffriedl89: If i am not wrong this should work as well :)

libs/tools/barista/src/transform.spec.ts
```diff
@@ -186,4 +188,35 @@ describe('Barista transformers', () => {
...
+      });
+      expect(transformed.content).toBe(
+        `<a href="http://www.dynatrace.com">Dynatrace</a>
+        <a href="ftpt://resource/guides">Resource</a>
+        <a href="smtp://resource/guides">Resource</a>
+        <a href="smthn://resource/guides">Resource</a>
+        <a href="lorem://resource/guides">Resource</a>
+        <a href="resources/bundle">Bundle</a>
+        <a routerLink="/resources/bundle">Bundle</a>
+        <a routerLink="//resource/guides">Resource</a>
```
lara-aigmueller: Shouldn't the link in line 215 become a routerLink, but the one in 217 not? 🤔

libs/tools/barista/src/transform.ts
```diff
@@ -212,11 +213,29 @@ export const relativeUrlTransformer: BaPageTransformer = async source => {
...
-            );
+            let u = url.parse(linkValue);
+            console.log(u);
+            // if (u.pathname) {
+            if (u.pathname === null) {
+              $(link.attribs).append(
+                $(link)
+                  .removeAttr('href')
+                  .attr('routerLink', '/')
+                  .attr('fragments', u.hash!),
```
rowa-audil: removed already

libs/tools/barista/src/transform.ts
```diff
@@ -300,3 +342,17 @@ export function internalContentTransformerFactory(
     return transformed;
   };
 }
+
+// Checks whether a URL is relative
+function isRelativeUrl(href: string): boolean {
+  return !href.match(/^(?:[a-z]+:)?\/\//i);
```
lukasholzer, line 348: Please provide a URL to a sample regex101 or something similar in a comment where we have a reference for this regex.

for example:

```suggestion
  // Matches every character case insensitive until a double point is coming. Has to have at least one character.
  // https://regex101.com/r/quBotu/3/
  return !href.match(/^(?:[a-z]+:)?\/\//i);
```

libs/tools/barista/src/transform.ts
```diff
@@ -300,3 +342,17 @@ export function internalContentTransformerFactory(
     return transformed;
   };
 }
+
+// Checks whether a URL is relative
```
lukasholzer, line 346: Method comments should be JsDoc. Because JsDoc is part of the function declaration and therefore it can be extracted for type information.

```suggestion
/** Checks whether a URL is relative */
```

libs/tools/barista/src/transform.ts
```diff
@@ -300,3 +342,17 @@ export function internalContentTransformerFactory(
...
+function isRelativeUrl(href: string): boolean {
+  return !href.match(/^(?:[a-z]+:)?\/\//i);
+}
+
+function toQueryParamValue(query: string): string {
```
lukasholzer, line 351: Please add function description

libs/tools/barista/src/transform.ts
```diff
@@ -300,3 +342,17 @@ export function internalContentTransformerFactory(
...
+  const params = new URLSearchParams(query);
+  const queryParams: string[] = [];
+  params.forEach((v, k) => {
```
lukasholzer, line 354: Please use more descriptive names like `value` and `key`

apps/barista-design-system/src/components/content-link/content-link.ts
```diff
@@ -0,0 +1,67 @@
...
+import { Router } from '@angular/router';
+
+/**
+ * The ba-content link component is used because we need to dynamically
+ * instanciate a router link and this is not possible, because it is a directive.
+ * So we create a component that handles the navigation.
+ */
+@Component({
+  selector: 'a[contentlink]',
+  template: '{{linkValue}}',
```
lukasholzer, line 27: please use an `ng-content` instead of the value content of a link can be more than text

apps/barista-design-system/src/pages/single-page/page-content.ts
```diff
@@ -72,6 +72,9 @@ export class BaPageContent implements OnDestroy {
       const placeholderElements: HTMLElement[] = Array.from(
         this._elementRef.nativeElement.querySelectorAll(factory.selector),
       );
+      // console.log(factory);
```
lukasholzer: remove this.

apps/barista-design-system/src/components/content-link/content-link.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  /** Fragement of absolute url */
+  @Input() fragment: string;
+
+  /** link value of absolute url */
+  @Input() linkValue: string;
+
+  constructor(private _router: Router) {}
+
+  ngOnInit(): void {}
```
lukasholzer, line 44: If you don't need the ngOnInit remove it.

apps/barista-design-system/src/components/content-link/content-link.ts
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * The ba-content link component is used because we need to dynamically
+ * instanciate a router link and this is not possible, because it is a directive.
+ * So we create a component that handles the navigation.
+ */
+@Component({
+  selector: 'a[contentlink]',
+  template: '{{linkValue}}',
+})
+export class BaContentLink implements OnInit {
```
lukasholzer, line 29: remove OnInit

apps/barista-design-system/src/components/content-link/content-link.ts
```diff
@@ -0,0 +1,67 @@
...
+  @Input() contentlink: string;
+
+  /** QueryParams of absolute url */
+  @Input() queryParams: { [k: string]: string };
+
+  /** Fragement of absolute url */
+  @Input() fragment: string;
+
+  /** link value of absolute url */
+  @Input() linkValue: string;
```
lukasholzer, line 40:
```suggestion
```

libs/barista-components/button/README.md
```diff
@@ -6,6 +6,10 @@
 
 ## Imports
 
+// TODO: REMOVE THIS!!! [input](/components/input) <br> // TODO: REMOVE THIS!!!
```
lukasholzer: Please remove

libs/tools/barista/src/transform.spec.ts
```diff
@@ -186,4 +188,48 @@ describe('Barista transformers', () => {
       );
     });
   });
+
+  describe('internalLinkReplacer', () => {
+    it('should replace internal links to routerLinks', async () => {
```
tomheller, line 193:
```suggestion
    it('should replace not fully qualified links with content-link component', async () => {
```

libs/tools/barista/src/transform.ts
```diff
@@ -202,6 +203,51 @@ export const copyHeadlineTransformer: BaPageTransformer = async source => {
...
+                  .removeAttr('href')
+                  .attr('contentLink', '/'),
+              );
+            } else {
+              $(link.attribs).append(
+                $(link)
+                  .removeAttr('href')
+                  .attr('contentLink', url.pathname),
+              );
+            }
```
tomheller, line 230: This could probably be simplified to keep the code a little bit more dry
```suggestion
              $(link.attribs).append(
                $(link)
                  .removeAttr('href')
                  .attr('contentLink', url.pathname || '/'),
              );
```

libs/tools/barista/src/transform.ts
```diff
@@ -300,3 +346,20 @@ export function internalContentTransformerFactory(
     return transformed;
   };
 }
+
+/** Checks whether a URL is relative  */
+function isRelativeUrl(href: string): boolean {
```
tomheller, line 351: I would like to add a quick sidenote on terminonlogy here.

_Relative Links_
Urls that either start without a slash or even dots in their href. They start resolving the URL from the pathname they are currently on: i.e. On the page `/components/button/` a relative link like `../` would be completely valid, and point to `/components/`;

_Absolute links_ 
are links that start with a slash, they start their navigation at the hostname level. That is why a link on the page `/components/button/` that looks like this `/patterns/guidelines` will end up on the barista.dynatrace.com/patterns/guidelines. 

_Fully qualified links_
Also include their hostname, i.e. `https://barista.dynatrace.com/components/button`.
Something like this, would also be a fully qualified link, but without a protocol specification (which means it will take the protocol of the page that you are currently on) `//barista.dynatrace.com/components/button`

--- 

The regex in this function checks if the url is fully qualified, but does not check if it is a relative path or not. 
If you expand the matcher in the demo to global and multiline (which will make it highlight the other test matches, you will see that it does not differentiate between relative and absolute links.
https://regex101.com/r/nzwuR4/3

libs/tools/barista/src/transform.ts
```diff
@@ -202,6 +203,51 @@ export const copyHeadlineTransformer: BaPageTransformer = async source => {
   return transformed;
 };
 
+/** Replaces absolute href with contentLink */
+export const relativeUrlTransformer: BaPageTransformer = async source => {
+  const transformed = { ...source };
+  if (source.content && source.content.length) {
+    transformed.content = runWithCheerio(source.content, $ => {
+      const links = $('a');
+      if (links.length) {
```
tomheller, line 212: I don't think you need to check for length in an each statement. It will just not iterate and it will make your nesting less deep.

apps/barista-design-system/src/components/content-link/content-link.ts
```diff
@@ -0,0 +1,93 @@
...
+  HostBinding,
+  OnChanges,
+  OnInit,
+} from '@angular/core';
+import { Router, Event, NavigationEnd } from '@angular/router';
+
+/**
+ * The ba-content link component is used because we need to dynamically
+ * instanciate a router link and this is not possible, because it is a directive.
+ * So we create a component that handles the navigation.
```
tomheller, line 30:
```suggestion
```
Feels a bit redundant.

apps/barista-design-system/src/components/content-link/content-link.ts
```diff
@@ -0,0 +1,93 @@
...
+  ngOnChanges(): void {
+    this._updateTargetUrlAndHref();
+  }
+
+  @HostListener('click', [
+    '$event.button',
+    '$event.ctrlKey',
+    '$event.metaKey',
+    '$event.shiftKey',
+  ])
```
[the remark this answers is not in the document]
rowa-audil: I couldn't find a way to get the arguments ($event.button, etc.) when using the decorator.
Is there a way to do that?

apps/barista-design-system/src/components/content-link/content-link.ts
```diff
@@ -0,0 +1,93 @@
...
+  @HostBinding() href: string;
+
+  /** QueryParams of absolute url */
+  @Input() queryParams: { [k: string]: string };
+
+  /** Fragement of absolute url */
+  @Input() fragment: string;
+
+  constructor(private _router: Router) {
+    //todo unsubscribe
```
lukasholzer, line 51: I think you forgot that :D

apps/barista-design-system/src/components/content-link/content-link.ts
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * The ba-content link component is used because we need to dynamically
+ * instanciate a router link and this is not possible, because it is a directive.
+ * So we create a component that handles the navigation.
+ */
+@Component({
+  selector: 'a[contentLink]',
+  template: '<ng-content></ng-content>',
+})
+//TODO: Implement HTML sanitizer in on init for linkValue if <ng-content> doesn't work.
```
lukasholzer, line 36: I think you can remove that.

apps/barista-design-system/src/components/content-link/content-link.ts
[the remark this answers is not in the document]
lukasholzer: @tomheller there is only one argument why you should define it in the decorator. This is when you have static property, that does not change. Then you would have to create a variable to bind it with a `@HostBinding` only to accomplish that. For that, the host property in the decorator is cleaner because you don't have to add additional logic to your component.
We said for the library it makes sense to have this convention because every component has a class on the host that is static. 

For apps I don't see any benefit - especially in this case I would prefer the decorator as it sticks to the class and makes the code more readable. Furthermore, it is originally the recommended way from angular to accomplish that.

Maybe we can take an action item out of that a discussion in the CODING_STANDARDS that mark this.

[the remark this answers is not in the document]
tomheller: @lukasholzer the onClick is a static function that is not changed. You can just as easily do that in the component decorator and pass the event in, there are no additional bindings needed. 

It is in the CODING_STANDARDS within the repository, this makes it apply to everything that is in the repository. 
If you want to discuss these things you can open a discussion issue. 

It would be the first introduction of a `@HostBinding` in the whole codebase and I am arguing that it could be done within the set guidelines. But for this pull request, I don't care...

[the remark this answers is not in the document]
ffriedl89: I agree with @tomheller currently it is in our coding standards and we should stick to it until we have a discussion about it.

libs/tools/barista/src/transform.ts
tomheller: Unresolved this comment, because it is not fixed.

apps/barista-design-system/src/components/content-link/content-link.ts
```diff
@@ -0,0 +1,102 @@
...
+ * instanciate a router link and this is not possible, because it is a directive.
+ */
+@Component({
+  selector: 'a[contentLink]',
+  template: '<ng-content></ng-content>',
+  styles: ['a:hover { cursor: pointer; }'],
+  host: {
+    '[href]': 'contentLink',
+    '(click)':
+      'onClick($event.button, $event.ctrlKey, $event.metaKey, $event.shiftKey)',
```
ffriedl89, line 40: I'd suggest to only pass the $event object here and use the properties in the function :) 
```suggestion
      'onClick($event)',
```

apps/barista-design-system/src/components/content-link/content-link.ts
```diff
@@ -0,0 +1,102 @@
...
+  /** QueryParams of absolute url */
+  @Input() queryParams: { [k: string]: string };
+
+  /** Fragement of absolute url */
+  @Input() fragment: string;
+
+  /** Href link to display */
+  href: string;
+
+  destroy$ = new Subject<void>();
```
ffriedl89, line 56:
```suggestion
  private _destroy$ = new Subject<void>();
```

apps/barista-design-system/src/components/content-link/content-link.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  constructor(private _router: Router) {
+    this._router.events
+      .pipe(takeUntil(this.destroy$))
```
ffriedl89, line 60:
```suggestion
      .pipe(takeUntil(this._destroy$))
```

apps/barista-design-system/src/components/content-link/content-link.ts
```diff
@@ -0,0 +1,102 @@
...
+    this._updateTargetUrlAndHref();
+  }
+
+  ngOnChanges(): void {
+    this._updateTargetUrlAndHref();
+  }
+
+  ngOnDestroy(): void {
+    this.destroy$.next();
+    this.destroy$.complete();
```
ffriedl89, line 78:
```suggestion
    this._destroy$.next();
    this._destroy$.complete();
```

apps/barista-design-system/src/components/content-link/content-link.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  onClick(
+    button: number,
+    ctrlKey: boolean,
+    metaKey: boolean,
+    shiftKey: boolean,
```
ffriedl89, line 85:
```suggestion
    event: Event
```

apps/barista-design-system/src/components/content-link/content-link.ts
```diff
@@ -0,0 +1,102 @@
...
+  ): boolean {
+    // Check if the click happens on a button or with some control keys pressed and let the
+    // click event bubble to let consumers or the browser handle the event.
+    if (button !== 0 || ctrlKey || metaKey || shiftKey) {
```
ffriedl89, line 89:
```suggestion
    if (event.button !== 0 || event.ctrlKey || event.metaKey || event.shiftKey) {
```

libs/tools/barista/src/transform.ts
```diff
@@ -300,3 +336,22 @@ export function internalContentTransformerFactory(
     return transformed;
   };
 }
+
+/**
+ * Checks whether a URL is a fully qualified link
+ */
+function isQualifiedLink(href: string): boolean {
+  // Matches every character case insensitive until a double point is coming. Has to have at least one character.
```
ffriedl89, line 344:
```suggestion
  // Matches every character case insensitive until a colon is coming. Has to have at least one character.
```

apps/barista-design-system/src/components/content-link/content-link.ts
```diff
@@ -0,0 +1,95 @@
...
+  },
+})
+export class BaContentLink implements OnChanges, OnInit, OnDestroy {
+  /** Absolute url for navigation on the page. For example /components/button */
+  @Input() contentLink: string;
+
+  /** QueryParams of absolute url */
+  @Input() queryParams: { [k: string]: string };
+
+  /** Fragement of absolute url */
```
tomheller, line 43: nit
```suggestion
  /** Fragment of absolute url */
```

libs/barista-components/button/README.md
```diff
@@ -1,5 +1,7 @@
 # Button
 
+Remove: [dummy](/components/input#invalid-input-types)
+
```
tomheller, line 4:
```suggestion
```

tomheller:
Well done, this was a complex one, but your provided solution seems good. 👍 
After others had a chance to review, you can squash the commits.

apps/barista-design-system/src/components/components.module.ts
```diff
@@ -49,6 +51,7 @@ export const BA_CONTENT_COMPONENTS: any[] = [
   imports: [
     CommonModule,
     RouterModule,
+
```
tomheller, line 54: NIT
```suggestion
```

apps/barista-design-system/src/components/content-link/content-link.ts
```diff
@@ -0,0 +1,95 @@
...
+    '[attr.contentlink]': 'contentLink',
+    '(click)': 'onClick($event)',
+  },
+})
+export class BaContentLink implements OnChanges, OnInit, OnDestroy {
+  /** Absolute url for navigation on the page. For example /components/button */
+  @Input() contentLink: string;
+
+  /** QueryParams of absolute url */
+  @Input() queryParams: { [k: string]: string };
```
tomheller, line 41: NIT: Coding guidelines suggest to not abbreviate variable names

```suggestion
  @Input() queryParams: { [key: string]: string };
```

apps/barista-design-system/src/components/content-link/content-link.ts
```diff
@@ -0,0 +1,95 @@
...
+  @Input() contentLink: string;
+
+  /** QueryParams of absolute url */
+  @Input() queryParams: { [key: string]: string };
+
+  /** Fragment of absolute url */
+  @Input() fragment: string;
+
+  /** Href link to display */
+  href: string;
```
ffriedl89, line 47: this is unused right? I might have missed this in the previous review

lara-aigmueller:
yeah, finally! 🎉 😉

PR #793, opened 2020-03-25
ci(slack-release-notifier): Send automatic Slack notifications on release

### LINKED ISSUE #785: Release bot action - Post changelog on slack
# Feature Request

## Summary

I want to have a bot that posts a message into our channel once a release is done on github containing the changelog.

## Feature Description

We can leverage the same webhook we already have for the reminder in our team channel on slack. See the https://github.com/dynatrace-oss/barista/blob/master/.github/workflows/slack-barista-production-promotion.yml action. There is also a possibility to add more complex styling for the slack messages to add the changelog in a nice way. 

See our help components channel for previous announcements for references :)

SkaillZ:

Adds a GitHub Action for automatically sending Slack messages when a new release is published (closes #785)

#### Type of PR

#### Checklist

### DIFF
--- a/.github/workflows/slack-release-notifier.yml
+++ b/.github/workflows/slack-release-notifier.yml
@@ -1,0 +1,15 @@
+name: "Slack release notifications"
+on:
+  release:
+    types: [published]
+jobs:
+  slack-release-notifier:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - uses: ./tools/slack-release-notifier/
+        with:
+          repo-token: "${{ secrets.GITHUB_TOKEN }}"
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_RELEASE_BOT_TEST }}
--- a/tools/slack-release-notifier/Dockerfile
+++ b/tools/slack-release-notifier/Dockerfile
@@ -1,0 +1,20 @@
+FROM node:12-alpine
+
+COPY \
+  ./package.json  \
+  ./package-lock.json \
+  ./tsconfig.json \
+  /slack-release-notifier/
+COPY ./entrypoint.sh /entrypoint.sh
+
+COPY ./src/ /slack-release-notifier/src/
+
+# We need git for an NPM package
+RUN apk update && apk upgrade && \
+    apk add --no-cache bash git openssh
+
+RUN cd /slack-release-notifier && npm ci
+
+RUN chmod +x /entrypoint.sh
+
+ENTRYPOINT ["/entrypoint.sh"]
--- a/tools/slack-release-notifier/package.json
+++ b/tools/slack-release-notifier/package.json
@@ -1,0 +1,24 @@
+{
+  "name": "@dynatrace/slack-release-notifier",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "author": "",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@actions/core": "^1.2.3",
+    "@actions/github": "^1.1.0",
+    "@slack/webhook": "^5.0.3",
+    "github-cherry-pick": "^1.1.0",
+    "html-to-mrkdwn": "github:github-slack/html-to-mrkdwn",
+    "marked": "^0.8.2",
+    "semver": "^7.1.3",
+    "ts-node": "^8.8.1",
+    "tslib": "^1.11.1",
+    "typescript": "^3.8.3"
+  },
+  "devDependencies": {
+    "@types/marked": "^0.7.3",
+    "@types/semver": "^7.1.0"
+  }
+}
--- a/tools/slack-release-notifier/src/main.ts
+++ b/tools/slack-release-notifier/src/main.ts
@@ -1,0 +1,117 @@
+/**
+ * @license
+ * Copyright 2019 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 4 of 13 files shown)


tools/slack-release-notifier/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN cd /slack-release-notifier && npm ci
+COPY ./src/ /slack-release-notifier/src/
+
+# We need git for an NPM package
+RUN apk update && apk upgrade && \
+    apk add --no-cache bash git openssh
+
+RUN cd /slack-release-notifier && npm ci
+
+RUN chmod +x /entrypoint.sh
```
SkaillZ: The flag could be set on the file itself but I couldn't do this on Windows

.github/workflows/slack-release-notifier.yml
```diff
@@ -0,0 +1,15 @@
...
+  slack-release-notifier:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - uses: ./tools/slack-release-notifier/
+        with:
+          repo-token: "${{ secrets.GITHUB_TOKEN }}"
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_RELEASE_BOT_TEST }}
```
SkaillZ: @ffriedl89 Please update to send to the right channel

tools/slack-release-notifier/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@dynatrace/barista-pull-request-labeler",
```
ffriedl89:
```suggestion
  "name": "@dynatrace/slack-release-notifier",
```

tools/slack-release-notifier/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "cherry-picker": "ts-node --project ./tsconfig.json src/main.ts"
```
ffriedl89: I think we can remove this script here

tools/slack-release-notifier/src/main.ts
```diff
@@ -0,0 +1,118 @@
...
+    ? release.name
+    : `${release.name} (${release.tag_name})`;
+}
+
+async function run(): Promise<void> {
+  const slackWebhookUrl = process.env.SLACK_WEBHOOK_URL;
+  const githubToken = getInput('repo-token', { required: true });
+
+  if (!slackWebhookUrl)
+    throw new Error('Required environment variables are not set.');
```
ffriedl89: NIT: I'd add that the slack webhook env variable is missing instead of a generic error message.

tools/slack-release-notifier/src/main.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  // The ref in the context is formatted like 'ref/tags/1.0.0'.
+  // The API expects only the tag string e.g. 1.0.0
+  const splitTag = context.ref.split('/');
+  const tag = splitTag[splitTag.length - 1];
+
+  const client = new GitHub(githubToken);
+
+  const release = (
+    await client.repos.getReleaseByTag({
```
ffriedl89: What happens if this errors ? What is the payload in that case?

tools/slack-release-notifier/src/main.ts
```diff
@@ -0,0 +1,118 @@
...
+      owner: context.repo.owner,
+      repo: context.repo.repo,
+      tag: tag,
+    })
+  ).data;
+
+  // To find out if we should announce a major, minor or patch release,
+  // we need to compare with the previous release version.
+  let previousRelease = await findPreviousRelease(client, tag);
```
ffriedl89: We already did these checks before the release - at that point the version is already released. If it the patch version is greater than 0 we have a patch version if the minor version is greater than 0 we have a minor version and otherwise we have a major version. Special cases are rc versions but they can be easily identified by the rc at the end :)

ffriedl89: You can use the semver package for that :) We already have it in the library - you can search for it

tools/slack-release-notifier/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM node:12-alpine
+
+COPY ./package.json /slack-release-notifier/package.json
+COPY ./package-lock.json /slack-release-notifier/package-lock.json
+COPY ./tsconfig.json /slack-release-notifier/tsconfig.json
```
lukasholzer: You can combine the copy into one layer makes the docker image smaller because it has fewer layers

```suggestion
COPY \
  ./package.json  \
  ./package-lock.json \
  ./tsconfig.json \
  /slack-release-notifier
```

tools/slack-release-notifier/src/main.ts
SkaillZ: I tested it with a 404 case and it threw an error which was caught and logged.

.github/workflows/slack-release-notifier.yml
ffriedl89: I'd rather test it on one release first and change the URL afterwards :)

tools/slack-release-notifier/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM node:12-alpine
```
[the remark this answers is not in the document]
ffriedl89: It is the github action and it has about 5 dependencies, why would we use the designops image here that does not even have the same dependencies installed as the action needs?

Issue #86, opened 2019-11-12
Overlay: Rethink overlay


ffriedl89:
currently we are solving the issue that the cdk overlay does not have a solution for overlays that need to follow the mouse
Rethink API and what we want to do with this service/directives

ffriedl89:
After discussing several approaches that would cover all usecases we have internally and for our consumers we decided it would be best to wrap the cdk and provide the needed styles / css classes on the panel. 

Additionally the DtOverlay as is should be deprecated once we have the alternative in place. 

Current working name for the new implementation would be `DtTooltip`. We think this will remove some of the confusion between UX and development since a tooltip and the overlay as we called it in code looked the same but where completely different components.

We also want to use the new version in all components internally to battle test the new approach.

We still need to keep functionality for our edge cases like pinning and the `DtMouseFollowPositionStrategy`. 

Usecases that we see:
Adding a tooltip to any element
```html
<button dtTooltipOrigin #origin="dtTooltipOrigin"></button>
<ng-template dtTooltip [dtTooltipOrigin]="origin"></ng-template>
```
Service usage
```typescript
DtTooltip.create(config?): DtTooltipRef
```
Reusing the same template for multiple tooltips
This  is currently not planned to be possible inside the template, but using the service in typescript this is possible

Special wrapper for common chart tooltip usecases `dt-chart-tooltip` has a default template if none is provided or can accept a ng-template as a content child
```html
<dt-chart>
	<dt-chart-tooltip></dt-chart-tooltip>
</dt-chart>

<dt-chart>
	<dt-chart-tooltip>
		<ng-template>
			my awesome template
		</ng-template>
	</dt-chart-tooltip>
</dt-chart>
```

ffriedl89:
As a more general discussion, is it worth the effort to wrap the directive/components provided by the CDK just so they are prefixed with `dt`.
What do we gain by doing that? 
For the overlay the API is quite straightforward and not that big, but thinking about #561 and drag & drop this needs way more effort.

I am seriously considering not wrapping the cdk related stuff, there is no real reason for me to do it. 

What's your take on it @dynatrace-oss/barista?

tomheller:
I would be totally fine to communicate to users of our library that overlays should be handled via the cdk and provide styling for the base of the overlay within our library. 
The possibilities of what cdk already provides with their overlay api would just be limited if we wrap only selected parts of it.

lukasholzer:
[...]

PR #803, opened 2020-03-27
Overlay: Fix mousemove

### LINKED ISSUE #220: DtOverlay: when triggering the overlay near the screen border it appears with a wrong width
# Bug Report

## Expected Behavior

Overlay should appears with the width of the content
## Current Behavior

Overlay appears with a wrong width when triggering the overlay near the screen border 

## Steps to Reproduce

1. Go to verlay-demo.component.scss
2. Add the following style
```
span {
  display: flex;
  width: 100%;
}
```
3. Hover the overlay near the border
image

1. 

### Used Versions:

- **node**: 12.1.0 
- **npm**:  6.9.0 
- **typescript**: 3.5.3
- **angular**: 8.2.14
- **@angular/cdk**: 8.2.3 
- **@dynatrace/barista-components**:  4.15.0

### LINKED ISSUE #772: Overlay is misplaced for trigger element consisting of mutiple elements
As discussed with Lukas:

See example:
https://github-slwyzc.stackblitz.io

When you hover over the first element everything is fine, hovering over "=" or the last word you see that the overlay is misplaced.

ffriedl89:
This PR fixes some long standing issues that keep popping up.

Fixes #772
Fixes #220

The discussion on what to do with the overlay is already quite long. This is the intermediate step of fixing what we have right now the best way I could. 

Plan for the future: Provide default config that can be used with the cdk overlay service and directives. 
And create a "tooltip" (working title) component that would cover 95% percent of the use cases in the template. More will follow in the #86

### DIFF
--- a/libs/barista-components/overlay/src/overlay-trigger.ts
+++ b/libs/barista-components/overlay/src/overlay-trigger.ts
@@ -18,7 +18,6 @@ import { FocusMonitor } from '@angular/cdk/a11y';
 import { ENTER, SPACE } from '@angular/cdk/keycodes';
 import { Platform } from '@angular/cdk/platform';
 import {
-  AfterViewInit,
   Attribute,
   Directive,
   ElementRef,
@@ -60,14 +59,11 @@ export const _DtOverlayTriggerMixin = mixinTabIndex(
   inputs: ['disabled', 'tabIndex'],
 })
 export class DtOverlayTrigger<T> extends _DtOverlayTriggerMixin
-  implements CanDisable, HasTabIndex, OnDestroy, AfterViewInit {
+  implements CanDisable, HasTabIndex, OnDestroy {
   private _content: TemplateRef<T>;
   private _config: DtOverlayConfig = new DtOverlayConfig();
   private _dtOverlayRef: DtOverlayRef<T> | null = null;
   private _moveSub = Subscription.EMPTY;
-
-  private _hostRect: ClientRect;
-  private _svgRect: ClientRect | null = null;
 
   /** Overlay pane containing the content */
   @Input('dtOverlay')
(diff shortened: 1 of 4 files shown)


ffriedl89:
FYI @deeleman

libs/barista-components/overlay/src/overlay-trigger.ts
```diff
@@ -93,16 +89,13 @@ export class DtOverlayTrigger<T> extends _DtOverlayTriggerMixin
     private _ngZone: NgZone,
     private _focusMonitor: FocusMonitor,
     @Attribute('tabindex') tabIndex: string,
+    // @ts-ignore
```
lukasholzer: May you explain why typescript failed here on this line?

Or can we remove that?

libs/barista-components/overlay/src/overlay.ts
```diff
@@ -93,11 +92,16 @@ const DT_OVERLAY_POSITIONS: ConnectedPosition[] = [
  */
 export const DT_OVERLAY_NO_POINTER_CLASS = 'dt-no-pointer';
 
+// TODO: Deprecate this once we have a tooltip component in place
```
lukasholzer: Have you created a follow-up issue for this?

libs/barista-components/overlay/src/overlay.ts
```diff
@@ -107,10 +111,14 @@ export class DtOverlay implements OnDestroy {
...
     private _injector: Injector,
     private _overlay: Overlay,
+    // @ts-ignore unused variable
     private _viewportRuler: ViewportRuler,
     // tslint:disable-next-line:no-any
+    // @ts-ignore unused variable
     @Inject(DOCUMENT) private _document: any,
+    // @ts-ignore unused variable
     private _platform: Platform,
+    // @ts-ignore unused variable
```
lukasholzer: If they are unused can we deprecate them?

libs/barista-components/overlay/src/overlay-trigger.ts
ffriedl89: Oh good catch - this was a mistake! 👍

libs/barista-components/overlay/src/overlay.ts
ffriedl89: We already have #86  - i'll add more information in the next days there.

libs/barista-components/overlay/src/overlay.ts
ffriedl89: Plan is to deprecated the entire component. That's why I didn't add a deprecation notice per argument.

libs/barista-components/overlay/src/overlay-trigger.ts
ffriedl89: Or not :D it is also unused. :D also did not check the file I am in while reading the comment. 

So it is unused since we don't need the platform anymore.

PR #806, opened 2020-03-27
fix(copy-to-clipboard): Fixes an issue where copied text is undefined

rowa-audil:
The copied text was undefined when an `input` element was provided.
Fixed by getting the `value` out of the element instead of `textContent`.


libs/barista-components/copy-to-clipboard/src/copy-to-clipboard.ts
```diff
@@ -93,7 +93,7 @@ export class DtCopyToClipboard implements AfterContentInit, OnDestroy {
   /** Copies the provided content to the clipboard. */
   copyToClipboard(): void {
     if (this._input && this._input.nativeElement.textContent !== null) {
```
ffriedl89: we should check here whether the value is defined with `isDefined`

PR #817, opened 2020-04-02
chore: Added design tokens to the library

tomheller:

First draft of adding design tokens to the library.
Included in this pull request: 
* Minimal design tokens.yml in a shared lib
* A build and package target for the design tokens to output scss and ts files
* A minimal copy of adobe/leonard-contrast-colors as we are currently not able to used their published library, because it ships a `type: module` package.json which node cannot deal with. 

#### Type of PR
Other

#### Checklist


package.json
```diff
@@ -62,6 +62,7 @@
     }
   },
   "dependencies": {
+    "@adobe/leonardo-contrast-colors": "^1.0.0-alpha.4",
```
ffriedl89, line 65: I think all of these should be dev dependencies right?

ffriedl89: By all of these i mean all added to the dependencies :D

libs/shared/design-tokens/generated/patterns/button.ts
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+/*
+ * THIS FILE IS GENERATED BASED ON THE DESIGN TOKENS DEFINED IN THE src/**.yml
+ * FILES, DO NOT CHANGE MANUALLY.
+ * TO GENERATE THESE FILES RUN `ng build shared-design-tokens`
+ */
+
+export const FLUID_BUTTON_PADDING = '0 16px';
```
ffriedl89, line 23: Just as a comment, would it make sense to transform these padding / spacing tokens into an object maybe like 
```
{ top: { value: 0} , right: { value: 16, unit: 'px' } ...
```

tomheller: I guess it could make sense to split them, but let's think about the use case of the design tokens in typescript here. In 99% of the cases, you are going to mirror the value of these tokens into a style attribute or something similar. If we split the values, a consumer would have to join these together manually again. 
I don't think we need to split these values up.

Also the parsing of multi-value css or other shorthands can be quite complicated.

tomheller:
Resolved conflicts again!

PR #818, opened 2020-04-02
ci: Increase the time for stale issues

ffriedl89:
I think the time of 30 days was too short especially if issues should be taken by contributors outside the team. I've extended the period to 90 days. Let's see how we cope with this change and watch our issue count over time

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -10,7 +10,7 @@ jobs:
     - uses: actions/stale@v1
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -12,5 +12,5 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         stale-issue-message: 'This issue is stale, because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days'
```
tomheller, line 13:
```suggestion
        stale-issue-message: 'This issue is stale, because it has been open 90 days with no activity. Remove stale label or comment or this will be closed in 5 days'
```

PR #820, opened 2020-04-02
Expandables Added aria-controls and aria-expanded

### LINKED ISSUE #788: A11y: Expandables (section, panel, trigger) lack aria properties
# Bug Report

The expandable trigger lacks the correct aria attributes.

## Expected Behavior

Expandable trigger: 
The trigger should have the `aria-controls` property set to the id of the expandable panel it controls.
The trigger should also have the `aria-expanded` set depending on the expand state of the panel. 

This is also expected to behave correctly in the expandable panel, but this should be covered since we use the trigger internally.

for references and a11y best practices please see https://www.w3.org/TR/wai-aria-practices-1.1/#accordion
https://www.w3.org/TR/wai-aria-1.1/#aria-expanded

## Current Behavior

Currently there is no aria-expanded or aria-controls property set on the trigger.

## Possible Solution

Add those properties to the host metadata. We might also have to introduce an id input for the expandable panel with a generated id defaulted similar to the id we have in the radio.ts for example to be

tomheller:
Fixed an issue where the correct aria attributes where not present on accordion like elements like
expandable-section, expandable-text and expandable-panel.

Fixes #788

#### Type of PR

Bugfix

#### Checklist

### DIFF
--- a/libs/barista-components/core/src/common-behaviours/id.ts
+++ b/libs/barista-components/core/src/common-behaviours/id.ts
@@ -1,0 +1,55 @@
+/**
+ * @license
+ * Copyright 2020 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Constructor } from './constructor';
+import { isDefined } from '../util';
+
+/**
+ * UniqueId counter, will be incremented with every
+ * instatiation of the ExpandablePanel class
+ */
+let uniqueId = 0;
+
+export interface HasId {
+  /** Represents the unique id of the component. */
+  id: string;
+}
+
+/** Mixin to augment a directive with a `id` property. */
+export function mixinId<T extends Constructor<{}>>(
+  base: T,
+  idPreset: string,
+): Constructor<HasId> & T {
+  return class extends base {
+    /** Sets a unique id for the expandable section. */
+    get id(): string {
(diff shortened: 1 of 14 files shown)


libs/barista-components/expandable-panel/src/expandable-panel.ts
```diff
@@ -71,6 +77,9 @@ import { Observable } from 'rxjs';
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class DtExpandablePanel {
+  /** @internal Unique identifier on the page for the expandable panel */
+  _uniqueId = `dt-expandable-panel-${uniqueId++}`;
+
   /** Whether the panel is expanded. */
```
ffriedl89, line 83: We should add an Input for the id on the component here and use the id if given otherwise fallback to the uniqueid

libs/barista-components/expandable-text/src/expandable-text.ts
```diff
@@ -45,6 +51,9 @@ import { Observable } from 'rxjs';
   encapsulation: ViewEncapsulation.Emulated,
 })
 export class DtExpandableText {
+  /** @internal Unique identifier on the page for the expandable text. */
```
ffriedl89, line 54: Same here i think for the id - maybe a mixin makes sense that takes a string for the uniqueid generation?

tomheller: I'm not sure how I would be able to do this in a mixin, as the element should get an ID that is referring somewhat to the component name or the selector.

But I can give it a try. One thing I can already see is, that it will be far less readable.

libs/barista-components/expandable-panel/src/expandable-panel.ts
tomheller: I agree on the input part, I will take care of this.

libs/barista-components/core/src/common-behaviours/id.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+let uniqueId = 0;
+
+export interface HasId {
+  /** Represents the unique id of the component. */
+  id: string;
+}
+
+/** Mixin to augment a directive with a `id` property. */
+export function mixinId<T extends Constructor<{}>>(
```
lukasholzer, line 32: @bardosmisi FYI

Issue #621, opened 2020-02-19
Move Custom Builders to Nx Plugins


lukasholzer:
# Feature Request

Move our custom builders to Nx plugins:

https://github.com/nrwl/nx/commit/fe98e29#diff-9e66bea35c8c76309609c9218bc259c4R30

Plugins are the official way how we should deal with our builders and whole tooling.

tomheller:
I'm questioning why we will need to do this. Is there a pain that we are solving with this one? As far as I can see, the tooling as we have it now works. What benefits do we gain when rewriting all our tooling to nrwl plugins?

lukasholzer:
not rewriting – it is more how it is integrated in the nx workspace. It should be more a convert library to plugin. Then we can get rid of the "hacky" way how our own builders are getting built.
with a `tsc --outdir /node_modules/dynatrace/barista-builders` for example

We should always follow the guidelines of nx because it is an opinionated folder structure and otherwise the tooling does not work as expected

tomheller:
Please correct me if I"m wrong here, but currently the tooling does work as expected, does it not?

lukasholzer:
does not. We encounter problems with files do not getting linted and imported from tooling where we crush our dependency graph. Our tools are no libraries but no tools in the meaning of nrwl/nx. They are plugins so we should treat them as one. This should be tackled after @ffriedl89 workspace refactorings.

tomheller:
Okay then. As I understood the explanation that @ffriedl89 gave me: 
* We had to move all our tools into the libs folder for #570 because nx has hard coded paths of `apps` and `libs` in their boundary linting rule
* Because we have moved tools into libs now, other linting rules of nx fail, because they disallow additional files in their libraryRoot (i.e. Dockerfile, ect. _files we depend on in some tool occations_)
* Nx provides with version 9 a new solution for these, because they apparently noticed, that there is a need for additional tooling. And their way of doing this is to add the tooling (which should not be considered a library), into a plugins folder where linting is less strict.

lukasholzer:
@tomheller perfect summary :D

PR #844, opened 2020-04-08
Moved the existing builders from the tools/builders to the workspace.

### LINKED ISSUE #414: Restructure Workspace builders
# Feature Request

## Summary

Currently, we cannot share common logic between builders and other libraries. So we had to duplicate this logic.

## Feature Description

Currently, we cannot share logic among other libraries and parts of our monorepo. The `@dynatrace/barista-builders` are getting built in a post-install step with a `tsc` typescript compile. The `outDir` is set to `node_modules/@dynatrace/barista-builders` so that our angular cli can use the compiled files directly without publishing it to npm. 

The problem with `tsc` is that it does not bundle files. So we need a module bundling to add files that are not located under the package root dir with typescript path aliases.

The current solution was more built to be a starting point that needs further improvements.

lukasholzer:
Now it is possible to share logic between libraries and the builders. Because we use for the workspace a module bundler (webpack), where we can specify multiple entry points. All workspace dependencies that are not an entry file are getting bundled into them. This is a part of issue #621.

Fixes #414

<hr>

  often helpful).
  guidelines.

#### Checklist


lukasholzer:
blocked until this issue #851 is merged because paths depend on each other. Might need adaption then.

Furthermore double check the `entrypoint.sh` for the barista preview.

libs/tools/shared/src/interfaces/package-json.ts
```diff
@@ -16,10 +16,15 @@
 
 export interface PackageJson {
   version?: string;
-  peerDependencies?: {
+  peerDependencies: {
```
ffriedl89, line 19: not all packageJson files have a peerDependency field
same for dev or normal dependencies - those fields should be optional

lukasholzer: Good point just copied it.

lukasholzer:
> @dynatrace-oss/barista, the auto cherry picker failed. Please have a look at the failure and cherry pick by hand.

Successfully resolved merge conflict by cherry picking by hand.

PR #846, opened 2020-04-08
perf: Switched to lodash-es as a peerDependency.

### LINKED ISSUE #4: Replace lodash dependency with lodash-es
**Description**
Currently, we have a dependency to `lodash` but we are not exposing it as peer dependency in case that angular ships `lodash`.
Lodash is not tree-shakeable so we are always forcing the consumer to ship the whole `lodash`. That is the reason why we should change to a tree-shakeable version of `lodash` – `lodash-es`.

**Expected behaviour 🌈**
Using all lodash methods like before only from the tree-shakeable version. 

**Additional context**

When using `lodash-es` it will work with the library but not with our unit tests that are written in jest. In case that jest cannot scope with es-modules.

ffriedl89:
BREAKING CHANGE: Install lodash-es instead of lodash now. Previously lodash was installed with angular automatically.

This PR finally works around the jest issue that we could not run lodash-es within jest. 
Fixes #4
Amazing :D 

This reduces the bundle size of the chart module by 120kb :O


lukasholzer:
@ffriedl89 please rebase then the preview will pass

PR #849, opened 2020-04-08
fix(filter-field): Fixes an issue where two filterfield flaps were open at the same time.

### LINKED ISSUE #840: Filter field: autocomplete stays open when clicking into another filter field
# Bug Report

## Expected Behavior

The autocomplete in a filter field should close when clicking into another filter field, just like it would for anywhere else in the page.

## Current Behavior

When clicking directly into two filter fields, the autocomplete menus for both will be open.

## Possible Solution

This likely sources from the filter field swallowing the click event in the [_handleHostClick](https://github.com/dynatrace-oss/barista/blob/master/libs/barista-components/filter-field/src/filter-field.ts#L572) method.

## Steps to Reproduce

StackBlitz: https://stackblitz.com/edit/github-dixtrt

1. Click into top filter field (the autocomplete will open)
2. Without clicking anywhere else on the page, click into the bottom filter field
3. Both filter fields will have their autocomplete menus open now.

## Context (Environment)

We are currently presenting the user with an option to manage two sets of filters on a single set of tag data. The filter fields are close to each other, so can be activated at the same time.

### Used Versions:

- **node**: v12.13.0
- **npm**: 6.14.1
- **typescript**: 3.8.3
- **angular**: 9.0.6
- **@angular/cdk**: 9.1.0
- **@dynatrace/barista-components**: 6.1.1

## Detailed Description

The autocomplete for a tag filter should close when clicking into another tag filter.

## Possible Implementation

Either the [_getFreeTextOutsideClickStream](https://github.com/dynatrace-oss/barista/blob/5a7a5687dd274075552489fb969d4a530f1ab930/libs/barista-components/filter-field/src/filter-field.ts#L1006) method or the [_handleHostClick](https://github.com/dynatrace-oss/barista/blob/master/libs/barista-components/filter-field/src/filter-field.ts#L572) method should probably handle this.

## Attachments

DeepinScreenshot_select-area_20200407101854

tomheller:
To ensure that only one filterfield flap is open at the same time, we need to store
the instance of the open flap in a global / module scope. If another filter field is focussed,
it should close the panels of the currently opened one and then reassign itself into the
open position.

Fixes #840

#### Type of PR
Bugfix

#### Checklist

### DIFF
--- a/libs/barista-components/filter-field/src/filter-field.ts
+++ b/libs/barista-components/filter-field/src/filter-field.ts
@@ -153,6 +153,10 @@ export class DtFilterFieldCurrentFilterChangeEvent<T> {
 
 export const DT_FILTER_FIELD_TYPING_DEBOUNCE = 200;
 
+// We need to save the instance of the filterField that currently
+// has a flap open, to make sure we can close the old one.
+let currentlyOpenFilterField: DtFilterField<any> | null = null;
+
 @Component({
   selector: 'dt-filter-field',
   exportAs: 'dtFilterField',
@@ -254,7 +258,7 @@ export class DtFilterField<T>
       }
 
       if (this._disabled) {
-        this.closeFilterPanels();
+        this._closeFilterPanels();
 
         this.tags.forEach(item => {
           this._previousTagDisabledState.set(item, item.disabled);
@@ -479,7 +483,20 @@ export class DtFilterField<T>
       .monitor(this._elementRef.nativeElement, true)
       .pipe(takeUntil(this._destroy$))
       .subscribe(origin => {
+        // The focusMonitor fires on focus and on blur, that is why we need the check
+        // if the currentlyOpenFilterfield is not the one that is currently being
+        // focused.
+        if (
+          currentlyOpenFilterField !== null &&
+          currentlyOpenFilterField !== this
+        ) {
+          currentlyOpenFilterField._closeFilterPanels();
+        }
         this._isFocused = isDefined(origin);
+        // Assign the currently open filter field when it is focused.
+        if (this._isFocused) {
+          currentlyOpenFilterField = this;
+        }
       });
 
     // tslint:disable-next-line:no-any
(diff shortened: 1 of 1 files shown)


tomheller:
I have added the `_filterLabelsClose()` call also to the ngOnDestroy, to make sure all labels are being closed when the component is being destroyed and the currentlyOpenFilterfield also resets.

libs/examples/src/index.ts
```diff
@@ -203,6 +203,7 @@ import { DtExampleMicroChartInterpolated } from './micro-chart/micro-chart-inter
 import { DtExampleMicroChartStream } from './micro-chart/micro-chart-stream-example/micro-chart-stream-example';
 import { DtExampleOverlayComplexContent } from './overlay/overlay-complex-content-example/overlay-complex-content-example';
 import { DtExampleOverlayDefault } from './overlay/overlay-default-example/overlay-default-example';
+import { DtExampleOverlayFullscreen } from './overlay/overlay-fullscreen-example/overlay-fullscreen-example';
```
ffriedl89, line 206: NIT: I guess this fixes an issue in barista with the newly introduced full screen overlay?

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -479,7 +483,20 @@ export class DtFilterField<T>
...
+        // if the currentlyOpenFilterfield is not the one that is currently being
+        // focused.
+        if (
+          currentlyOpenFilterField !== null &&
+          currentlyOpenFilterField !== this
+        ) {
+          currentlyOpenFilterField._closeFilterPanels();
+        }
         this._isFocused = isDefined(origin);
+        // Assign the currently open filter field when it is focsed.
```
ffriedl89, line 496: NIT
```suggestion
        // Assign the currently open filter field when it is focused.
```

libs/examples/src/index.ts
tomheller: The file is generated every time one runs npm run demos. If it was not run when a new dome was created or not checked in at the time, it will be changed. 

I will remove it from this PR and create a separate one.

Issue #798, opened 2020-03-26
Barista: Universal renderer throws console error with IconColorWheel


lukasholzer:
# Bug Report

Barista Universal rendering is throwing an error when rendering the icon colour wheel (console error):

![image](https://user-images.githubusercontent.com/11156362/77656576-5cadb980-6f74-11ea-9dea-f3ca5e40fcfd.png)

## Steps to Reproduce

1. `npm run examples-tools`
2. `npm run barista-tools`
3. `ng build workspace`
4. `ng run barista-design-system:render:production`

Nevertheless, the build is working it is throwing the console error. Should be investigated!

if a `ng run barista-design-system:serve-ssr` is performed instead of step 4 and a manual navigation to an icon detail page is performed: `resources/icons/xhrprocess` it throws this error as well.

```
ERROR Error: NotYetImplemented
    at Element.XFB9.exports.nyi (/Users/lukas.holzer/Sites/barista/dist/apps/barista-design-system/server/main.js:1:2595009)
    at EmulatedEncapsulationServerRenderer2.DefaultServerRenderer2.setProperty (/Users/lukas.holzer/Sites/barista/dist/apps/barista-design-system/server/main.js:1:2091582)
    at BaseAnimationRenderer.setProperty (/Users/lukas.holzer/Sites/barista/dist/apps/barista-design-system/server/main.js:1:1821057)
    at elementPropertyInternal (/Users/lukas.holzer/Sites/barista/dist/apps/barista-design-system/server/main.js:1:5042408)
    at Object.ɵɵproperty (/Users/lukas.holzer/Sites/barista/dist/apps/barista-design-system/server/main.js:1:5132587)
    at BaIconColorWheel_Template (/Users/lukas.holzer/Sites/barista/dist/apps/barista-design-system/server/main.js:1:909712)
    at executeTemplate (/Users/lukas.holzer/Sites/barista/dist/apps/barista-design-system/server/main.js:1:5036827)
    at refreshView (/Users/lukas.holzer/Sites/barista/dist/apps/barista-design-system/server/main.js:1:5033418)
    at refreshComponent (/Users/lukas.holzer/Sites/barista/dist/apps/barista-design-system/server/main.js:1:5049608)
    at refreshChildComponents (/Users/lukas.holzer/Sites/barista/dist/apps/barista-design-system/server/main.js:1:5035599)
```

rowa-audil:
I cannot reproduce the bug when going over these steps.
Magically fixed??

PR #850, opened 2020-04-08
Barista icon color wheel bug

rowa-audil:
Since the `symbol` Element inside an `SVG` element doesn't support setting innerHTML dynamically, I had to check if the context is the browser and if not then I don't render that Element

Fixes: #798

### DIFF
--- a/apps/barista-design-system/src/components/icon-color-wheel/icon-color-wheel.html
+++ b/apps/barista-design-system/src/components/icon-color-wheel/icon-color-wheel.html
@@ -3,7 +3,7 @@
 </p>
 <dt-switch (change)="_switchChanged($event)">Convert to png</dt-switch>
 
-<svg viewBox="0 0 512 512" style="display: none;">
+<svg *ngIf="_renderSymbol" viewBox="0 0 512 512" style="display: none;">
   <defs>
     <symbol [attr.id]="_iconId" [innerHTML]="_svg"></symbol>
   </defs>
--- a/apps/barista-design-system/src/components/icon-color-wheel/icon-color-wheel.ts
+++ b/apps/barista-design-system/src/components/icon-color-wheel/icon-color-wheel.ts
@@ -19,6 +19,7 @@ import { Component, ElementRef, Input, ViewChild } from '@angular/core';
 import { DtSwitchChange } from '@dynatrace/barista-components/switch';
 import { DtColors } from '@dynatrace/barista-components/theming';
 import { DomSanitizer, SafeHtml } from '@angular/platform-browser';
+import { Platform } from '@angular/cdk/platform';
 
 interface BaColorWheelBlob {
   colorName: string;
@@ -62,13 +63,16 @@ export class BaIconColorWheel {
   _iconFillColor = DtColors.GRAY_700;
   /** @internal all colored circle blobs */
   _coloredBlobs: BaColorWheelBlob[];
+  /** @internal Whether context is browser */
+  _renderSymbol: boolean;
 
   /** whether the icon should be downloaded as png */
   private _convertToPng = false;
 
-  constructor(private _sanitizer: DomSanitizer) {
+  constructor(private _platform: Platform, private _sanitizer: DomSanitizer) {
+    this._renderSymbol = this._platform.isBrowser;
     const groupedBlobs = Object.keys(DtColors)
-      /** breaking-change Can eb removed when flat_white is removed from the color list. 7.0.0 */
+      /** breaking-change Can be removed when flat_white is removed from the color list. 7.0.0 */
       .filter(key => key !== 'FLAT_WHITE') // no flat white for the color wheel
       .map((key): BaColorWheelBlob | undefined => {
         /* tslint:disable no-magic-numbers */


rowa-audil:
Currently, using the renderer2 for setting the innerHTML of the <symbol> Element inside the SVG-tag when the context is not in the browser is the best solution I came up with.
We'll have to update this if we decide to use ivy.

apps/barista-design-system/src/components/icon-color-wheel/icon-color-wheel.ts
```diff
@@ -154,6 +173,16 @@ export class BaIconColorWheel {
     this._iconId = `ba-icon-color-wheel-iconId-${++uniqueColorwheelId}`;
   }
 
+  ngAfterViewInit(): void {
+    if (!this.isBrowser) {
```
tomheller, line 177: We can use the renderer in both cases, can't we? 
I don't see why we need to complicate this

apps/barista-design-system/src/components/icon-color-wheel/icon-color-wheel.ts
```diff
@@ -62,11 +74,18 @@ export class BaIconColorWheel {
...
   /** whether the icon should be downloaded as png */
   private _convertToPng = false;
 
-  constructor(private _sanitizer: DomSanitizer) {
+  constructor(
+    private _sanitizer: DomSanitizer,
+    @Inject(PLATFORM_ID) private platformId: Object,
+    private _renderer2: Renderer2,
+  ) {
+    this.isBrowser = isPlatformBrowser(this.platformId);
```
tomheller, line 88: I guess the platform check is not necessary if you use the renderer to change the view.

apps/barista-design-system/src/components/icon-color-wheel/icon-color-wheel.ts
```diff
@@ -62,11 +74,18 @@ export class BaIconColorWheel {
   _iconFillColor = DtColors.GRAY_700;
   /** @internal all colored circle blobs */
   _coloredBlobs: BaColorWheelBlob[];
+  /** @internal whether context is in browser */
+  isBrowser: boolean;
```
tomheller, line 78:
```suggestion
```

apps/barista-design-system/src/components/icon-color-wheel/icon-color-wheel.ts
```diff
@@ -54,6 +64,8 @@ export class BaIconColorWheel {
   /** @internal */
   @ViewChild('colorwheel', { static: true }) _colorWheel: ElementRef;
 
+  @ViewChild('symbol') _symbol: ElementRef;
```
tomheller, line 67: Can be static if it is no longer `if`ed in the template.

apps/barista-design-system/src/components/icon-color-wheel/icon-color-wheel.ts
```diff
@@ -14,11 +14,21 @@
...
+  Inject,
+  PLATFORM_ID,
+  AfterViewInit,
+  Renderer2,
+} from '@angular/core';
 
 import { DtSwitchChange } from '@dynatrace/barista-components/switch';
 import { DtColors } from '@dynatrace/barista-components/theming';
 import { DomSanitizer, SafeHtml } from '@angular/platform-browser';
+import { isPlatformBrowser } from '@angular/common';
```
tomheller, line 31:
```suggestion
```

apps/barista-design-system/src/components/icon-color-wheel/icon-color-wheel.html
```diff
@@ -5,7 +5,8 @@
 
 <svg viewBox="0 0 512 512" style="display: none;">
   <defs>
-    <symbol [attr.id]="_iconId" [innerHTML]="_svg"></symbol>
+    <symbol *ngIf="!isBrowser" #symbol [attr.id]="_iconId"></symbol>
+    <symbol *ngIf="isBrowser" [attr.id]="_iconId" [innerHTML]="_svg"></symbol>
```
tomheller, line 9:
```suggestion
    <symbol #symbol [attr.id]="_iconId"></symbol>
```

apps/barista-design-system/src/components/icon-color-wheel/icon-color-wheel.ts
```diff
@@ -66,7 +75,7 @@ export class BaIconColorWheel {
   /** whether the icon should be downloaded as png */
   private _convertToPng = false;
 
-  constructor(private _sanitizer: DomSanitizer) {
+  constructor(private _sanitizer: DomSanitizer, private _renderer2: Renderer2) {
     const groupedBlobs = Object.keys(DtColors)
       /** breaking-change Can eb removed when flat_white is removed from the color list. 7.0.0 */
```
tomheller, line 80:
```suggestion
      /** breaking-change Can be removed when flat_white is removed from the color list. 7.0.0 */
```

apps/barista-design-system/src/components/icon-color-wheel/icon-color-wheel.ts
```diff
@@ -54,6 +61,8 @@ export class BaIconColorWheel {
   /** @internal */
   @ViewChild('colorwheel', { static: true }) _colorWheel: ElementRef;
 
+  @ViewChild('symbol', { static: true }) _symbol: ElementRef;
```
tomheller, line 64: Does this work without the #named reference in the template?

rowa-audil: The delete button was too excited ^^

apps/barista-design-system/src/components/icon-color-wheel/icon-color-wheel.ts
```diff
@@ -154,6 +163,14 @@ export class BaIconColorWheel {
     this._iconId = `ba-icon-color-wheel-iconId-${++uniqueColorwheelId}`;
   }
 
+  ngAfterViewInit(): void {
+    this._renderer2.setProperty(
```
lukasholzer, line 167: can you please do it without the renderer since it should be avoided.

```typescript
this._symbol.nativeElement.innerHTML = this._svg
```

tomheller: Will this work in SSR then, when using nativeElement?

rowa-audil:
Found out that setting innerHTML to the symbol element is not possible and throws the bug even when using the renderer. I think the better solution, for now, is to check whether the context is the browser.

apps/barista-design-system/src/components/icon-color-wheel/icon-color-wheel.ts
lukasholzer: Furthermore, I question why you need the view child it looks like that it is not used in your PR if I'm correct so please remove it or use it.

PR #860, opened 2020-04-10
docs(table): Improved observable and responsive examples

rarenal:
Introduces little improvements in table documentation.

### Responsive example:
This example uses an expandable row. Since the expandable-cell was switched to be shown at the left of the table, this example was showing wrongly the columns, and the example code needed to be adapted
image

- Changed Details column alignment to the left (text alignment)
- Changed header columns creation, to put Details column first, in case of expandable rows

### Observable example
This example allows to dynamically add rows from a source. But as the user doesn't know when it was going to finish (in fact in generates a row every second, during 5 seconds), it feels confusing to not be able to stop that subscription.

- Added a button to cancel the subscription, that gets switched when starting a subscription
- Moved buttons to the top, as otherwise the buttons are getting moved every time a row is added (making it more difficult to click them)

#### Type of PR

#### Checklist

### DIFF
--- a/libs/examples/src/table/table-observable-example/table-observable-example.html
+++ b/libs/examples/src/table/table-observable-example/table-observable-example.html
@@ -1,3 +1,19 @@
+<button
+  *ngIf="!isSubscribed; else cancel"
+  class="dt-example-button"
+  dt-button
+  (click)="startSubscription()"
+>
+  Start subscription
+</button>
+<button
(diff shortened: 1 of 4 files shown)


tomheller:
Hi @rarenal ! Thank you for this contribution. It seems that there is a conflict in one of the examples template files. Could you please have a look and resolve it. Otherwise, the changes are good 👍

PR #870, opened 2020-04-14
chore(changelog): Change the order of changelog. Fix stream merging iss…

bardosmisi:
Change the order of changelog. Fix stream merging issue. Fix regex to match new format of changelog.

#### Type of PR


libs/tools/release/src/changelog/changelog.ts
```diff
@@ -32,31 +32,36 @@ export async function prependChangelogFromLatestTag(
   });
 
   const newChangelogStream: any = getNewChangelog(headerPartial, releaseName);
-
   // Stream for reading the existing changelog. This is necessary because we want to
   // actually prepend the new changelog to the existing one.
-  const previousChangelogStream = createReadStream(changelogPath);
+  const previousChangelogStream: any = createReadStream(changelogPath);
```
lukasholzer, line 37: Please avoid the `any` because propper types help us finding bugs when upgrading dependencies and if they break something with their updates.

```suggestion
  const previousChangelogStream: ReadStream = createReadStream(changelogPath);
```

bardosmisi: Hi, yeah, this is something I don't like either. But further down I use ```readableEnded```, and ts thinks ```ReadStream``` does not have this property. This is why I used ```any```.

bardosmisi:
@lukasholzer , Thanks for checking this 👍 , I will fix and presumably extend unit tests.

thomaspink:
Looks good in general, but the commit type should not be `fix` because it is not related to consumer code that is part of the released package and should also not be part in the changelog itself. 

Please change to `chore`, `build`, etc.

libs/tools/release/src/changelog/changelog.ts
lukasholzer: but the `createReadStream` is from fs and the official node types for fs state that it is from the type `ReadStream` – so maybe investigate a little bit further.. have you tried it without nodejs streams?

lukasholzer:
@ffriedl89 and @tomheller can you take a look at this one? would be nice to merge before we release a new version!

PR #888, opened 2020-04-17
ci: Fix broken barista builder due to update of angular ngcc.

lukasholzer:
Due to the update of the angular version ngcc is creating a json file in the node_modules,
that gets loaded when no package.json is available in the repo. Because npm is searching for something
similar to the package.json. But the due to the fact that it is no package json the npm install
of the fonts and icons will fail.

Adding a package.json does the trick.

### DIFF
--- a/.deployment/k8s-barista-builder/Jenkinsfile
+++ b/.deployment/k8s-barista-builder/Jenkinsfile
@@ -9,9 +9,10 @@ pipeline {
 
   options {
     ansiColor("xterm")
-		buildDiscarder(logRotator(numToKeepStr: '10'))
+    buildDiscarder(logRotator(numToKeepStr: '10'))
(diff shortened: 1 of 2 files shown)


.deployment/k8s-barista-builder/Jenkinsfile
```diff
@@ -24,6 +25,38 @@ pipeline {
 
   stages {
 
+    stage('Checkout Files') {
+      steps {
+        checkout([
+          $class: 'GitSCM',
+          branches: [
+            [name: '*/ci/fix-broken-barista-builder']
```
tomheller, line 33:
```suggestion
            [name: '*/master']
```

lukasholzer:
Build is now up and running again:

image

PR #906, opened 2020-04-22
CI add parallelisation of jobs to reduce build time.

lukasholzer:
## Without Parallelization ~10min
image

## Wit the new Parallelization ~5min
image

### DIFF
--- a/libs/barista-components/empty-state/src/empty-state.ts
+++ b/libs/barista-components/empty-state/src/empty-state.ts
@@ -267,9 +267,11 @@ export class DtEmptyState
    * @breaking-change Remove with version 8.0.0
    */
   _updateLayout(): void {
-    const componentWidth = this._elementRef.nativeElement.getBoundingClientRect()
-      .width;
-    this._updateLayoutForSize(componentWidth);
+    if (this._platform.isBrowser) {
+      const componentWidth = this._elementRef.nativeElement.getBoundingClientRect()
+        .width;
+      this._updateLayoutForSize(componentWidth);
+    }
   }
 }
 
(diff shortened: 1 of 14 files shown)


lukasholzer:
Could reduce the **max build time** from **`25m 23s`** to **`13min 10s`** if everything is affected!

nearly half the size!

image

image

libs/barista-components/empty-state/src/empty-state.ts
```diff
@@ -267,9 +267,11 @@ export class DtEmptyState
...
    */
   _updateLayout(): void {
-    const componentWidth = this._elementRef.nativeElement.getBoundingClientRect()
-      .width;
-    this._updateLayoutForSize(componentWidth);
+    if (this._platform.isBrowser) {
+      const componentWidth = this._elementRef.nativeElement.getBoundingClientRect()
+        .width;
+      this._updateLayoutForSize(componentWidth);
+    }
```
ffriedl89, line 274: These changes make PR #972 obsolete. I am fine with merging this and closing #972

ffriedl89: I guess during the review this PR got already merged. Either rebase and use your changeset or remove this changeset from the PR.

lukasholzer: It is not obsolete because the `_updateLayout()` is currently called from the `_updateLayout` function inside the `DtCustomEmptyStateBase`.  
image

lukasholzer: Your PR added only the check that the `_updateLayout` function is not called inside the `ngAfterViewInit`

Issue #739, opened 2020-03-16
Theming concept with custom properties


ffriedl89:
# Feature Request

We are currently shipping a lot of styles for our theming capabilities. Since we are no longer supporting IE11 we can safely move to custom properties for theming purposes. 

Goal: 
Work out a concept for theming that leverages custom properties and therefore decreases our bundle sizes. Especially the button's bundle is huge ~200kb. Make a PoC before we change all components to the new theming

Also consider that we don't want to have breaking changes visually. Since all classes are prefixed with `dt` we can safely change the way we handle them.

Initial idea: The dt-theme directive could set the custom properties correctly for the given theme. The components use the properties inherited from the theme. Therefore we can still have nested themes the same way we do now (context dialog that has a dark theme on a light page) 

- [ ] Switch the dt-button we have to themed version using custom-properties
- [ ] Make a list of all components that need changing
- [ ] Keep the old way of theming intact - so we can switch component by component

--- 

Other components that should be converted: 

* Alert #956 
* Bar indicator #957 
* Breadcrumbs #958 
* Button group
* Card #970 
* Chart heatfield
* Checkbox #971
* Consumption
* Expandable section
* Form field
* Highlight
* Icon
* Info group
* Progress bar
* Progress circle
* Radio button
* Show more
* Switch
* Table
* Tabs
* Tile
* Toggle button group
* Tree table

ffriedl89:
Size of button.js in the esm2015 modules is 242kb - lets compare afterwards

PR #920, opened 2020-04-27
feat(button): Changed button theming to custom properties

NinaKammerlander:

<hr>
Changes the theming of the button component so it works with custom properties.
This PR is part of #739  

#### Type of PR

#### Checklist

### DIFF
--- a/libs/barista-components/button/src/_button-theme.scss
+++ b/libs/barista-components/button/src/_button-theme.scss
@@ -1,227 +1,276 @@
-@import '../../core/src/theming/theming';
-@import '../../core/src/style/interactive-common';
+.dt-button {
+  --dt-button-default-color: var(--dt-theme-main-default-color);
+  --dt-button-hover-color: var(--dt-theme-main-hover-color);
+  --dt-button-active-color: var(--dt-theme-main-active-color);
+}
 
-@mixin dt-theme-button($theme) {
-  $palette-names: 'main', 'warning', 'cta';
-  $default-palette: 'main';
+.dt-button.dt-color-warning {
+  --dt-button-default-color: var(--dt-warning-default-color);
+  --dt-button-hover-color: var(--dt-warning-hover-color);
+  --dt-button-active-color: var(--dt-warning-active-color);
+}
 
-  @each $name in $palette-names {
-    $is-dark: map-get($theme, 'is-dark');
-    $palette: dt-get-theme-palette($theme, $name);
-    $default-color: dt-get-theme-color($palette, 'default');
-    $hover-color: dt-get-theme-color($palette, 'hover');
-    $active-color: dt-get-theme-color($palette, 'active');
+.dt-button.dt-color-cta {
+  --dt-button-default-color: var(--dt-cta-default-color);
+  --dt-button-hover-color: var(--dt-cta-hover-color);
+  --dt-button-active-color: var(--dt-cta-active-color);
+}
 
-    // Temporary fix to not make button color depend on theme.
-    // We only have to override the main color because the warning and
-    // cta colors are the same across all themes for now.
-    @if ($name == 'main') {
-      @if ($is-dark == true) {
-        $default-color: $turquoise-500;
-        $hover-color: $turquoise-400;
-        $active-color: $turquoise-300;
-      } @else {
-        $default-color: $turquoise-600;
-        $hover-color: $turquoise-700;
-        $active-color: $turquoise-800;
-      }
+////////////////////////////
+// Primary Button
+////////////////////////////
+.dt-button-primary {
+  background: var(--dt-button-default-color);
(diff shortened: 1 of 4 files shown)


libs/barista-components/button/src/_button-theme.scss
```diff
@@ -1,227 +1,271 @@
...
-    }
 
-    $namespace: '.dt-button';
-    @if $name != $default-palette {
-      $namespace: '.dt-button.dt-color-#{$name}';
-    }
+.dt-color-warning {
+  --dt-theme-main-default-color: var(--dt-warning-default-color);
+  --dt-theme-main-hover-color: var(--dt-warning-hover-color);
+  --dt-theme-main-active-color: var(--dt-warning-active-color);
```
tomheller, line 5: I'm not entirely certain how good of an idea it is to reset the theme main color in this case. Maybe we should think about scoping the variable names to the button itself, this way we can prevent side effects if other components that allow nesting access i.e. the `--dt-theme-main-default-color`.

```suggestion
  --dt-button-default-color: var(--dt-warning-default-color);
  --dt-button-hover-color: var(--dt-warning-hover-color);
  --dt-button-active-color: var(--dt-warning-active-color);
```

libs/barista-components/button/src/_button-theme.scss
```diff
@@ -1,227 +1,271 @@
...
+}
+
+.dt-color-cta {
+  --dt-theme-main-default-color: var(--dt-cta-default-color);
+  --dt-theme-main-hover-color: var(--dt-cta-hover-color);
+  --dt-theme-main-active-color: var(--dt-cta-active-color);
```
tomheller, line 11: And the same here of course.
```suggestion
  --dt-button-default-color: var(--dt-cta-default-color);
  --dt-button-hover-color: var(--dt-cta-hover-color);
  --dt-button-active-color: var(--dt-cta-active-color);
```

libs/barista-components/button/src/_button-theme.scss
```diff
@@ -1,227 +1,271 @@
...
+}
+
+////////////////////////////
+// Primary Button
+////////////////////////////
+.dt-button-primary {
+  background: var(--dt-theme-main-default-color);
+  border-color: var(--dt-theme-main-default-color);
```
tomheller, line 19: And only use the buttons own variables in assignments: 
```suggestion
  background: var(--dt-button-default-color);
  border-color: var(--dt-button-default-color);
```

libs/barista-components/button/src/_button-theme.scss
```diff
@@ -1,227 +1,271 @@
...
-        $hover-color: $turquoise-700;
-        $active-color: $turquoise-800;
-      }
-    }
 
-    $namespace: '.dt-button';
-    @if $name != $default-palette {
-      $namespace: '.dt-button.dt-color-#{$name}';
-    }
+.dt-color-warning {
```
tomheller, line 2: Maybe we can scope the color definitions to the host and define a default if they are not set.
For custom properties you can set a default, when a value is not defined. I guess this would be a good fallback if the button is not wrapped into a dtTheme. 

```suggestion
:host {
  --dt-button-default-color: var(--dt-theme-main-default-color, $turquoise500);
  --dt-button-hover-color: var(--dt-theme-main-hover-color, $turquoise600);
  --dt-button-active-color: var(--dt-theme-main-active-color, $turquoise700);
}
:host.dt-color-warning {
  --dt-button-default-color: var(--dt-warning-default-color);
  --dt-button-hover-color: var(--dt-warning-hover-color);
  --dt-button-active-color: var(--dt-warning-active-color);
}
```

Please not that this is not tested code and the turquoise color values may not be correct.

ffriedl89: Maybe it makes more sense to set the default/fallback on the main.scss file to basically on the :root element. Because if the consumer does not use the dt-theme it will fallback to the root definitions for the custom properties. 
```
:root {
  // i am turquoise by default
--dt-theme-main-
}
```
Otherwise we would end up with default values spread across the components.
What do you think @tomheller @NinaKammerlander

tomheller: I like the idea of defaulting to the turquoise theme in the `main.scss` file with a `:root` selector. 
We probably will still end up with different defaults on a component level, because I think not every default is set to `main` color palette.

tomheller:
I have rebased the commits. Please give another review @ffriedl89 @lukasholzer. This one looks quite ready.

tomheller:
Fixed formatting issue in PR.

tomheller:
Rebased in CI fix for build to pass again.

tomheller:
Manually cherry picked to 6.x
New head is f698f5885e5aa642d9551314662ab3a68e0c1424

PR #924, opened 2020-04-29
feat(table): Add order functionality to table.

peter-affenzeller:
#### Type of PR

#### Checklist

### DIFF
--- a/apps/components-e2e/src/components/table-order/table.e2e.ts
+++ b/apps/components-e2e/src/components/table-order/table.e2e.ts
@@ -1,0 +1,305 @@
+/**
+ * @license
+ * Copyright 2020 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { resetWindowSizeToDefault, waitForAngular } from '../../utils';
+import {
+  dataCells,
+  dragHandles,
+  disableButton,
+  expandButtons,
+  getDragDistance,
+  orderInputs,
+  changeOrderButton,
+} from './table.po';
+
+fixture('Table Order')
+  .page('http://localhost:4200/table-order')
+  .beforeEach(async () => {
+    await resetWindowSizeToDefault();
+    await waitForAngular();
+  });
+
+test('should only be draggable at the drag handle', async (testController: TestController) => {
+  await testController
+    .expect(dataCells.nth(0).textContent)
+    .eql('I')
+    .expect(dataCells.nth(1).textContent)
+    .eql('II')
+    .drag(dataCells.nth(0), 0, await getDragDistance(0, 1), {
+      speed: 0.2,
+    })
+    .expect(dataCells.nth(0).textContent)
+    .eql('I')
+    .expect(dataCells.nth(1).textContent)
+    .eql('II');
+});
+
+test('should only be draggable at the drag handle', async (testController: TestController) => {
+  await testController
+    .expect(dataCells.nth(0).textContent)
+    .eql('I')
+    .expect(dataCells.nth(1).textContent)
+    .eql('II')
+    .drag(orderInputs.nth(0), 0, await getDragDistance(0, 1), {
+      speed: 0.2,
+    })
+    .expect(dataCells.nth(0).textContent)
+    .eql('I')
+    .expect(dataCells.nth(1).textContent)
+    .eql('II');
+});
+
+test('drag - should reorder the table - drag towards last row', async (testController: TestController) => {
+  await testController
+    .expect(dataCells.nth(0).textContent)
+    .eql('I')
+    .expect(dataCells.nth(1).textContent)
+    .eql('II')
+    .drag(dragHandles.nth(0), 0, await getDragDistance(0, 1), {
+      speed: 0.2,
+    })
+    .expect(dataCells.nth(0).textContent)
+    .eql('II')
+    .expect(dataCells.nth(1).textContent)
+    .eql('I');
+});
+
+test('drag - should reorder the table - drag towards first row', async (testController: TestController) => {
+  await testController
+    .expect(dataCells.nth(0).textContent)
+    .eql('I')
+    .expect(dataCells.nth(1).textContent)
+    .eql('II')
+    .expect(dataCells.nth(2).textContent)
+    .eql('III')
+    .drag(dragHandles.nth(2), 0, await getDragDistance(2, 0), {
+      speed: 0.2,
+    })
+    .expect(dataCells.nth(0).textContent)
+    .eql('III')
+    .expect(dataCells.nth(1).textContent)
+    .eql('I')
+    .expect(dataCells.nth(2).textContent)
+    .eql('II');
+});
+
+test('drag - should not reorder the table if disabled', async (testController: TestController) => {
+  await testController
+    .click(disableButton)
+    .expect(dataCells.nth(0).textContent)
+    .eql('I')
+    .expect(dataCells.nth(1).textContent)
+    .eql('II')
+    .drag(dragHandles.nth(0), 0, await getDragDistance(0, 1), {
+      speed: 0.2,
+    })
+    .expect(dataCells.nth(0).textContent)
+    .eql('I')
+    .expect(dataCells.nth(1).textContent)
+    .eql('II');
+});
+
+test('programmatically - should reorder the table', async (testController: TestController) => {
+  await testController
+    .expect(dataCells.nth(0).textContent)
+    .eql('I')
+    .expect(dataCells.nth(1).textContent)
+    .eql('II')
+    .click(changeOrderButton)
+    .expect(dataCells.nth(0).textContent)
+    .eql('II')
+    .expect(dataCells.nth(1).textContent)
+    .eql('I');
+});
+
+test('programmatically - should not reorder the table if disabled', async (testController: TestController) => {
+  await testController
+    .click(disableButton)
+    .expect(dataCells.nth(0).textContent)
+    .eql('I')
+    .expect(dataCells.nth(1).textContent)
+    .eql('II')
+    .click(changeOrderButton)
+    .expect(dataCells.nth(0).textContent)
+    .eql('I')
+    .expect(dataCells.nth(1).textContent)
+    .eql('II');
+});
+
+test('input - should reorder the table - higher index', async (testController: TestController) => {
+  await testController
+    .expect(dataCells.nth(0).textContent)
+    .eql('I')
+    .expect(dataCells.nth(1).textContent)
+    .eql('II')
+    .typeText(orderInputs.nth(0), '1', { replace: true })
+    .pressKey('enter')
+    .expect(dataCells.nth(0).textContent)
+    .eql('II')
+    .expect(dataCells.nth(1).textContent)
+    .eql('I');
+});
+
+test('input - should reorder the table - lower index', async (testController: TestController) => {
+  await testController
+    .expect(dataCells.nth(0).textContent)
+    .eql('I')
+    .expect(dataCells.nth(1).textContent)
+    .eql('II')
+    .expect(dataCells.nth(2).textContent)
+    .eql('III')
+    .typeText(orderInputs.nth(2), '0', { replace: true })
+    .pressKey('enter')
+    .expect(dataCells.nth(0).textContent)
+    .eql('III')
+    .expect(dataCells.nth(1).textContent)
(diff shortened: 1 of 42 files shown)


libs/barista-components/table/src/order/order.ts
```diff
@@ -0,0 +1,111 @@
...
+  exportAs: 'dtOrder',
+  inputs: ['disabled: dtOrderDisabled'],
+  host: {
+    '[class.dt-order-disabled]': 'disabled',
+  },
+})
+export class DtOrder<T> extends _DtOrderMixinBase
+  implements CanDisable, AfterViewInit, OnChanges, OnDestroy {
+  /** Event that is emitted whenever the input value of the order column changes. */
+  @Output('dtOrderChange') readonly orderChange = new EventEmitter<
```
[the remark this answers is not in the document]
peter-affenzeller: yeah, does fire in both cases, I'll change the comment to make that clearer

libs/barista-components/table/src/table-module.ts
```diff
@@ -91,13 +98,15 @@ const EXPORTED_DECLARATIONS = [
...
     CdkTableModule,
+    DragDropModule,
     DtIconModule,
     DtFormFieldModule,
     DtInputModule,
     DtButtonModule,
     DtIndicatorModule,
     DtFormattersModule,
     DtEmptyStateModule,
+    ReactiveFormsModule,
```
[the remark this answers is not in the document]
peter-affenzeller: I use a form control for validating the input field of the order cell

peter-affenzeller:
@ffriedl89 UX pushed a new drag handle icon, I'll change that as soon as it is available.
I implemented the suggested changes, the discussed indicator animation for moved rows after an input change is also included.

libs/barista-components/table/src/order/order-cell.ts
```diff
@@ -0,0 +1,158 @@
...
+   */
+  _animateOrderChangedIndicator = false;
+
+  /** Previous index of the row to pass to the order function */
+  private _prevIndex: number;
+
+  /** Subject used for unsubscribing */
+  private _destroy$ = new Subject<void>();
+
+  @ViewChild('orderInput', { static: true }) _orderInput: ElementRef<
```
thomaspink, line 78: NIT: If you format it that way you do not need to linebreak the generic
```
@ViewChild('orderInput', { static: true }) 
_orderInput: ElementRef<HTMLInputElement>;
```

libs/barista-components/table/src/order/order-cell.ts
```diff
@@ -0,0 +1,158 @@
...
+  >;
+
+  _orderFormControl = new FormControl('', [
+    Validators.required,
+    Validators.pattern(INPUT_PATTERN),
+  ]);
+
+  constructor(
+    public _columnDef: DtColumnDef,
+    public _changeDetectorRef: ChangeDetectorRef,
```
thomaspink, line 89: Does not need to be public
```suggestion
    columnDef: DtColumnDef,
    changeDetectorRef: ChangeDetectorRef,
```

libs/barista-components/table/src/table-order-data-source.ts
```diff
@@ -0,0 +1,119 @@
...
+  private _data: BehaviorSubject<T[]>;
+
+  /**
+   * Subscription to the changes that should trigger an update to the table's rendered rows, such
+   * as ordering or base data changes.
+   */
+  private _renderChangesSubscription = Subscription.EMPTY;
+
+  /** Array of data that should be rendered by the table, where each object represents one row. */
+  get data(): T[] {
```
thomaspink, line 36: Why do we need to reimplement this if we extend from the datasource which already holds the exact same logic?

libs/barista-components/table/src/table-order-data-source.ts
```diff
@@ -0,0 +1,119 @@
...
+import { BehaviorSubject, Observable, Subscription, of } from 'rxjs';
+import { map, withLatestFrom } from 'rxjs/operators';
+
+import { DtTable } from './table';
+import { DtOrder, DtOrderChangeEvent, DtOrderReorderEvent } from './order';
+import { moveItemInArray } from '@angular/cdk/drag-drop';
+
+export class DtTableOrderDataSource<T> extends DataSource<T> {
+  /** Stream that emits when a new data array is set on the data source. */
+  private _data: BehaviorSubject<T[]>;
```
thomaspink, line 27: Why do we need to reimplement this if we extend from the datasource which already holds the exact same property?

libs/barista-components/table/src/table-order-data-source.ts
```diff
@@ -0,0 +1,119 @@
...
+  get order(): DtOrder<T> | null {
+    return this._order;
+  }
+  set order(order: DtOrder<T> | null) {
+    this._order = order;
+    this._updateChangeSubscription();
+  }
+  private _order: DtOrder<T> | null;
+
+  constructor(initialData: T[] = []) {
```
thomaspink, line 57: Why do we need to reimplement this if we extend from the datasource which already holds the exact same logic?

libs/barista-components/table/src/table-order-data-source.ts
```diff
@@ -0,0 +1,119 @@
...
+    );
+
+    return clonedData;
+  }
+
+  /**
+   * Used by the DtTable. Called when it connects to the data source.
+   */
+  connect(_table: DtTable<T>): Observable<T[]> {
+    return this._data;
```
thomaspink, line 110: Why isn't the ordered table paginated (like in the table data source; see filteredData)?

peter-affenzeller: that was something that also came up in the review and @ursula-wieshofer suggested that pagination in combination with drag&drop usually leads to UX issues and should be avoided.

lukasholzer:
dependencies

libs/barista-components/table/src/order/order-cell.scss
```diff
@@ -0,0 +1,121 @@
...
+    border: 1px solid transparent;
+  }
+}
+
+.dt-simple-order-cell-data {
+  display: inline-flex;
+  align-items: center;
+}
+
+::ng-deep .dt-simple-order-column-icon {
```
tomheller, line 85: can we prefix a host here or do we need to escape the encapsulation here because of the dragging.

libs/barista-components/table/src/order/order-cell.scss
```diff
@@ -0,0 +1,121 @@
...
+  margin-right: 4px;
+  width: auto !important;
+  height: 1.5rem !important;
+}
+
+::ng-deep .dt-table:not(.dt-order-disabled) .dt-simple-order-column-icon {
```
tomheller, line 91: can we prefix a host here or do we need to escape the encapsulation here because of the dragging.

libs/barista-components/table/src/order/order-cell.scss
```diff
@@ -0,0 +1,121 @@
...
+  cursor: pointer;
+}
+
+::ng-deep .dt-simple-order-column-icon svg {
```
tomheller, line 95: can we prefix a host here or do we need to escape the encapsulation here because of the dragging.

libs/barista-components/table/src/order/order-cell.scss
```diff
@@ -0,0 +1,121 @@
...
+
+.dt-simple-order-cell-data {
+  display: inline-flex;
+  align-items: center;
+}
+
+::ng-deep .dt-simple-order-column-icon {
+  margin-right: 4px;
+  width: auto !important;
+  height: 1.5rem !important;
```
tomheller, line 88: Can we do without the important here? maybe double up on the class selector

libs/barista-components/table/src/order/order-cell.scss
```diff
@@ -0,0 +1,121 @@
...
+::ng-deep .dt-table:not(.dt-order-disabled) .dt-simple-order-column-icon {
+  cursor: pointer;
+}
+
+::ng-deep .dt-simple-order-column-icon svg {
+  width: auto;
+  fill: $turquoise-600;
+}
+
+::ng-deep .dt-table.dt-order-disabled .dt-simple-order-column-icon svg {
```
tomheller, line 100: can we prefix a host here or do we need to escape the encapsulation here because of the dragging.

libs/barista-components/table/src/order/order-cell.scss
```diff
@@ -0,0 +1,121 @@
...
+  text-align: left;
+}
+
+.dt-simple-order-column-input::placeholder {
+  opacity: 0.4;
+}
+
+.dt-simple-order-input-invalid {
+  background: $red-200 !important;
+  border: 1px solid $error-color !important;
```
tomheller, line 116: Can we remove the importants?

libs/barista-components/table/src/row.ts
```diff
@@ -29,9 +29,21 @@ import {
...
+import {
+  style,
+  transition,
+  animate,
+  trigger,
+  AnimationEvent,
+  keyframes,
+} from '@angular/animations';
+
+// prettier-ignore
```
tomheller, line 45:
```suggestion
```

libs/barista-components/table/src/row.ts
```diff
@@ -54,17 +66,40 @@ export class DtRowDef<T> extends CdkRowDef<T> {}
...
+    ]),
+  ],
 })
 export class DtRow extends CdkRow implements OnDestroy {
   /**
    * @internal Necessary due to the fact that we cannot get the DtRow via normal DI
    */
   static _mostRecentRow: DtRow | null = null;
 
+  _orderChangedIndicatorAnimationState: 'void' | 'changed' = 'void';
```
tomheller, line 101: missing jsDoc

libs/barista-components/table/src/table-module.ts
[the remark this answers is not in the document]
lukasholzer: the normal FormsModule should be enough for your use case

apps/components-e2e/src/components/table-order/table.e2e.ts
```diff
@@ -0,0 +1,60 @@
...
+    .eql('I')
+    .expect(dataCells.nth(1).textContent)
+    .eql('II')
+    .drag(dragHandle, 0, 120, {
+      speed: 0.2,
+    })
+    .expect(dataCells.nth(0).textContent)
+    .eql('I')
+    .expect(dataCells.nth(1).textContent)
+    .eql('II');
```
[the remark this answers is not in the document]
peter-affenzeller: I'll add a couple more for the input fields and whatever I can think of that makes sense, sure, forgot about that after fighting some other issues...

libs/barista-components/table/src/order/order-cell.scss
```diff
@@ -0,0 +1,121 @@
...
+  * get higher after reordering
+  */
+  .dt-expandable-row-base-collapsed
+    + .dt-expandable-row-collapsible
+    .dt-expandable-row-expandable {
+    padding: 0;
+  }
+}
+
+::ng-deep {
```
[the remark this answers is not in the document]
peter-affenzeller: in this case it is needed to escape encapsulation because of the dragging, yes.

libs/barista-components/table/src/table-module.ts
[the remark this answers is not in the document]
peter-affenzeller: no, I can't bind to formControl if I switch to the FormsModule, or am I missing something?

libs/barista-components/table/src/table-order-data-source.ts
peter-affenzeller: changed the order data source to extend from the table data source

peter-affenzeller:
@tomheller I changed the styling to get rid of the importants and add prefixes if possible. also included the dt-main-font mixin to fix the font-size of the drag preview (changed slightly in the barista examples)
added the examples in the readme and included more e2e tests.

@ffriedl89 since you can enter a very high number in the input to move a row to the last position, I changed the order-directive slightly to clamp the value to the number of rows when emitting the order change event. else it would emit e.g. previousIndex 0, currentIndex 999 even though there are only 10 rows in the table.

the build succeeds now, thx @lukasholzer for helping!

libs/barista-components/table/src/table-module.ts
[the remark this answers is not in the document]
ffriedl89: It's perfectly fine to use the ReactiveFormsModule in my opinion, since we have @angular/forms as a peerDep already - so if you want to use the formControl internally that's ok for me :)

libs/barista-components/table/src/table-order-data-source.ts
peter-affenzeller: @thomaspink I actually changed this back to extend from the cdk data source, else the user would still be able to call things related to sorting, which is what we wanted to avoid.

tomheller: CDK datasource does not implement data or the constructor, so I don't see a problem with this implementation:

CDK table datasource export:
https://github.com/angular/components/blob/14a51ef07e8ec4db4696c2dd4b4fcbb295dfcb24/src/cdk/table/public-api.ts#L18-L19

DataSource in the cdk collections:
https://github.com/angular/components/blob/14a51ef07e8ec4db4696c2dd4b4fcbb295dfcb24/src/cdk/collections/data-source.ts#L12-L31

PR #929, opened 2020-04-29
Removed Table in use headline

rowa-audil:
Removed `Table in use` headline from the table component readme and moved it into database


ffriedl89:
@rowa-audil Does it still show up in the toc?

rowa-audil:
@ffriedl89 
Internally it does.

rowa-audil:
@ffriedl89 Not anymore.

Issue #926, opened 2020-04-29
filter-field: open popup when scrolling down after removing filters


christian-fischer:
# Bug Report

## Expected Behavior

I think your intended behavior is to not be able to scroll, as long as the filter-field popup is open?

## Current Behavior

When scrolling down, the filter-field popup stays visible in place.

## Steps to Reproduce

Used components: filter-field + table.
The table data source changes depending on the filters.

Example: https://stackblitz.com/edit/github-xnqkhf-lasdrw

1. add a filter
2. remove the filter
3. scroll down

## Context (Environment)

My use case: 
load data from the server depending on the filters in the filter-field; this data is then used for the table data source.

### Used Versions:

- **@dynatrace/barista-components**: 6.3.0

## Attachments

![image](https://user-images.githubusercontent.com/5974170/80580090-48455c80-8a0b-11ea-8def-02a2c9360ebe.png)

tomheller:
Hi @christian-fischer!
Thank you for pointing this out. The dropdown should stay open while the filter field has focus and users should be able to scroll as well to interact with the rest of the page, but the positioning is definitely off. 
We will have a look and see what can be done to fix this.

christian-fischer:
Interesting. I noticed that there is some scrollbar disabling going on, when the popup is open. So this is not intentional?
See example https://stackblitz.com/edit/github-ltwlg3 where I increased the height

tomheller:
I notices as well, this might be a sideeffect of the bug and the positioning of the overlay. 
We are using a cdkOverlay to render the dropdown here, and it might get the wrong positioning logic applied. I haven't had time to check this one yet, but I will keep you posted.

christian-fischer:
Awesome! Thank you @tomheller

tomheller:
After some research, i figured out that apparently the dropdown from the filterfield / autocomplete is currently set to be scrollBlocking.

And it seems that the `scrollStrategies.block()` causes this issue. I'm not quire sure why this is only appearing now.

tomheller:
We have figured it out, thanks to the help of @ffriedl89!
For scroll blocking to work, the body element needs to be expandable (it can't be set to 100vh, which it is in the examples and demo app. Changing the body rule to `min-height: 100vh;`, the scroll blocking works again. 

[...]

PR #930, opened 2020-04-29
chore: The body should be expandable for scrollblocking to actually work

tomheller:

Fixes an issue where scroll blocking did not work when the body was not expandable and had a fixed height set. 

Related #926 

#### Type of PR
Chore

#### Checklist


ffriedl89:
@rowa-audil Does it still show up in the toc? I guess the toc is something we do clientside instead of on build right?

PR #943, opened 2020-05-05
Feat/prettier update

lukasholzer:

Due to the fact that we are using typescript `3.8.x` we need to update to prettier 2.0. The support for typescript was added in their [2020 major version](https://prettier.io/blog/2020/03/21/2.0.0.html).

Otherwise, it will break type only imports.
Add specific support for Angular and Web Components parsing on templates.

This PR consists out of 2 commits the second one was auto generated by `yarn format --all`

This PR is needed for #924 to fix the failing formatting for the type only imports.


tomheller:
@lukasholzer you missed to label the PR before merging it, which means the cherry picker did not run. 
All format checks on 6.x were broken. I have cherry picked the 2 commits to 6.x, new head is c3870569316a311062011201033eab4503acc46a

lukasholzer:
oh damn sorry. @tomheller all fine now?

tomheller:
@lukasholzer nothing to be sorry about. yes all fine

Issue #909, opened 2020-04-23
Ng upgrade 7.0.0


ffriedl89:
# Feature Request

Provide an ng-upgrade schematic for version 7.0.0. 
Similar to the upgrade schematic we had for 6.0.0. 
* Rename inputs in templates and ts files (aria-label-close -> ariaLabelClose)
* Install & update peerDependencies (lodash-es, highcharts)

## TODO

- [x] Rename aria labels in html files automatically
    - [x] For example: Consumer is doing [timestamp.html](https://github.com/dynatrace-oss/barista/blob/master/apps/components-e2e/src/components/chart/selection-area/timestamp/timestamp.html#L3) 
- [ ] ~Refactor element from dtBreadcrumbsItem to anchor (**occurrences in cluster**: none)~
- [x] Refactor `dt-cta-*` elements to `dt-empty-state` (occurrences: 1)
- [ ] ~FiltlerField use `currentTags` instead of `tags` (occurrences: none)~
- [ ] ~Refactor dtTime to dtDuration formatter (occurrences: none)~
- [x] Refactor `dt-empty-state-item-img` selector to `dt-empty-state` (occurrences: 116)
- [x] Refactor `dt-table-empty-state-message` selector to `dt-empty-state` (occurrences: 10)
- [x] Refactor `dt-table-empty-state-title` selector to `dt-empty-state` (occurrences: 6)
- [x] Refactor `dt-table-empty-state` selector to `dt-empty-state` (occurrences: 28)
- [x] Refactor `dtTableEmptyState` anchor to `dt-empty-state` (occurrences: 3)

lukasholzer:
We need to update the `ng-add` as well for the latest versions.

PR #944, opened 2020-05-05
Ng upgrade 7 schematic

rowa-audil:
**Update schematics for barista version 7.0.0**
See #909 for a list of migrations.

Fixes: #909


lukasholzer:
This is blocked until this issue #948 after that it will need further refactorings

lukasholzer:
This will be solved by #949  as we have to completely remove the CDK we had to rewrite it.

Issue #922, opened 2020-04-28
Restyle confirmation dialog


SirTroyer:
# Feature Request

Since our confirmation dialog currently is too subtle for our users to see it on larger screens we would like to make it more prominent.

Following changes should be done:
* make it darker
* move buttons to the left (more consistent with our button positioning guidelines + on larger screens the button can be too far away from the content)
* do not dock the confirmation dialog directly to the edge of the screen but give a little margin
* add shadow to make it pop from the content above
* text should be placed above the buttons making the confirmation dialog higher and therefore also improve visibility
* provide option to make the confirmation dialog wiggle (future behaviour for the screens will be the following: when the user tries to navigate away from the screen with unsaved changes we'd like to stay on the page and wait for the user to do either save or discard and therefore draw the users attention to the confirmation dialog with a wiggle animation also indicating that what the user anticipated to do is invalid)

see following mockup:
![confirmation_dialog_mock](https://user-images.githubusercontent.com/61140594/80456472-ee229980-892d-11ea-97d0-4704b6c1ae0d.jpg)

tomheller:
I guess this is related to the #577 
There was a discussion ongoing and based on the comment from @ffriedl89, seems that you are willing to contribute. Is this still accurate? 

Can you outline the API changes that will be required for your improvements?

SirTroyer:
oh right #577 is a duplicate issue
yes I'm still willing to contribute we planed to implement it in this sprint
hmm not sure wat API changes are required but it should not change a lot but just a new capability for the wiggle animation the rest are just styling changes I guess

tomheller:
> but just a new capability for the wiggle animation

This is what I was referring to. Maybe just quickly outline what the default option is, and how you can (opt in / opt out) of the wiggle.

SirTroyer:
hmm I thought actually about providing a stream as a parameter like `wiggle` to trigger the animation because I guess the confirmation dialog component does not know / probably should not know about navigation
or am I wrong?

tomheller:
You are right, the confirmation-dialog should not know about the navigation. 
That's why I was asking about details on this one. 
An input subject that you subscribe on within the confirmation dialog or a public method would be fine.

SirTroyer:
perfect :)

SirTroyer:
[...]

PR #974, opened 2020-05-08
feat(confirmation-dialog): Restyle confirmation dialog + add wiggle a…

SirTroyer:
Implement changes suggested in #922 to make the confirmation dialog more prominent.

<hr>

Following changes should are done:

- [x] make it darker
- [x] move buttons to the left (more consistent with our button positioning guidelines + on larger screens the button can be too far away from the content)
- [x] do not dock the confirmation dialog directly to the edge of the screen but give a little margin
- [x] add shadow to make it pop from the content above
- [x] text should be placed above the buttons making the confirmation dialog higher and therefore also improve visibility
- [x] provide option to make the confirmation dialog wiggle (future behaviour for the screens will be the following: when the user tries to navigate away from the screen with unsaved changes we'd like to stay on the page and wait for the user to do either save or discard and therefore draw the users attention to the confirmation dialog with a wiggle animation also indicating that what the user anticipated to do is invalid)

#### Type of PR


libs/barista-components/confirmation-dialog/src/confirmation-dialog.spec.ts
```diff
@@ -146,10 +141,28 @@ describe('ConfirmationDialogComponent', () => {
...
+      fixture.detectChanges();
+
+      // wiggle again
+      componentInstance.dialog.wiggle();
+      expect(componentInstance.dialog._wiggleState).toEqual(true);
+      fixture.detectChanges();
+      tick();
+      expect(componentInstance.dialog._wiggleState).toEqual(false);
+    }));
   });
```
ffriedl89, line 167: A testcase where we check whether a change in the viewport resizer changes the left position would be nice.

thomaspink:
@SirTroyer Thank you for the PR. Looks great at a first glance. Will start reviewing it soon. :)

ffriedl89:
@SirTroyer  And you need to sign the cla agreement since this is your first contribution. See the comment and help from the bot above.

ffriedl89:
FYI: We are investigating whether we can test the viewporter resizer and position interaction in jest.

libs/barista-components/confirmation-dialog/src/confirmation-dialog.ts
```diff
@@ -177,6 +198,12 @@ export class DtConfirmationDialog
     this.state = null;
     this._overlayRef.dispose();
     this._stateChildrenSubscription.unsubscribe();
+    this._viewportChangesSubscription.unsubscribe();
+  }
+
+  /** Triggers a wiggle animation on the confirmation dialog */
+  wiggle(): void {
+    this._wiggleState = true;
   }
```
ffriedl89, line 207: After rethinking this during finalizing this PR - is a more generic function name like `focusAttention()` better - at some point we might decide to switch to another animation. Then wiggle is tightly coupled to the animation used.

ffriedl89: I've renamed the method to focusAttention and added the entry to the readme

tomheller:
There was still `wiggle` used in the devApp. I have corrected and changed it to `focusAttention`

PR #1015, opened 2020-05-14
chore(schematics): Dt e2e schematic

bardosmisi:

Adding schematics to generate e2e components.

### DIFF
--- a/libs/workspace/src/schematics/dt-component-e2e/index.spec.ts
+++ b/libs/workspace/src/schematics/dt-component-e2e/index.spec.ts
@@ -1,0 +1,155 @@
+/**
+ * @license
+ * Copyright 2020 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Tree } from '@angular-devkit/schematics';
+import {
+  SchematicTestRunner,
+  UnitTestTree,
+} from '@angular-devkit/schematics/testing';
+import {
+  Schema as ApplicationOptions,
+  Style,
+} from '@schematics/angular/application/schema';
+import { Schema as WorkspaceOptions } from '@schematics/angular/workspace/schema';
+import { promises as fs } from 'fs';
+import { join } from 'path';
+
+export async function getFixture(filePath: string): Promise<string> {
+  const fixturesFolder = join(__dirname, './fixtures');
+  return fs.readFile(join(fixturesFolder, filePath), {
+    encoding: 'utf-8',
+  });
+}
+
+export async function addFixtureToTree(
+  tree: Tree,
+  source: string,
+  destination: string,
+): Promise<void> {
+  const content = await getFixture(source);
+  if (tree.exists(destination)) {
+    tree.overwrite(destination, content);
+    return;
+  }
+  tree.create(destination, content);
+}
+
+const getNewFiles = (newFiles: string[], oldFiles: string[]): string[] =>
+  newFiles.filter((file) => !oldFiles.includes(file));
+
+describe('Add E2E Component', () => {
+  const workspaceOptions: WorkspaceOptions = {
+    name: 'workspace',
+    newProjectRoot: 'apps',
+    version: '8.0.0',
+  };
+
+  const appOptions: ApplicationOptions = {
+    name: 'myapp',
+    inlineStyle: false,
+    inlineTemplate: false,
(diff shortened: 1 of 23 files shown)


ffriedl89:
@bardosmisi i think this is ready to be switched to a real PR and get review right?

libs/workspace/src/schematics/dt-component-e2e/index.ts
```diff
@@ -0,0 +1,129 @@
...
+        (module) => module.${options.e2eComponent.module},
+      ),
+  },`;
+    const routesChange = new InsertChange(modulePath, end, toInsert);
+
+    return commitChanges(host, [routesChange], modulePath);
+  };
+}
+
+// TODO: extract reusable naming functions and move to utils
```
lukasholzer, line 109: is this already done?

libs/workspace/src/schematics/dt-component-e2e/index.ts
```diff
@@ -0,0 +1,129 @@
...
+
+export default function (options: DtComponentE2EOptions): Rule {
+  return async (_host: Tree) => {
```
lukasholzer, line 112: as you do not need any await you can remove the async.

```suggestion
  return () => {
```

libs/workspace/src/schematics/dt-component-e2e/index.ts
```diff
@@ -0,0 +1,129 @@
...
+ */
+export function addRoute(options: DtE2EExtendedOptions): Rule {
+  return (host: Tree) => {
+    const modulePath = join(
+      'apps',
+      'components-e2e',
+      'src',
+      'app',
+      'app.routing.module.ts',
+    );
```
lukasholzer, line 75: No need for joining here: If you write the path as a string you gain multiple benefits in this case. 
1. It gets way more readable 
2. If you need to adapt paths and you utilize the search functionality and you search for `apps/components-e2e/` then you will find this occurrence otherwise you may forget about it.

```suggestion
    const modulePath = 'apps/components-e2e/src/app/app.routing.module.ts';
```

libs/workspace/src/schematics/dt-component-e2e/index.ts
```diff
@@ -0,0 +1,129 @@
...
+    ).find(
+      (node: ts.VariableDeclaration) => node.name.getText() === 'routes',
+    ) as ts.VariableDeclaration;
+    const routesElements = (routesDeclaration.initializer as ts.ArrayLiteralExpression)
+      .elements;
+    const lastElement = routesElements[routesElements.length - 1];
+    const end = routesElements.hasTrailingComma
+      ? lastElement.getEnd() + 1
+      : lastElement.getEnd();
+    const indentation = getIndentation(routesElements);
```
lukasholzer, line 93: I would omit the indentation and run a task on as the last step in the schematics that uses prettier to format your changes. a simple `npm run format:write`.

THen you don't have to care about trailing commas indentations and everything.

libs/workspace/src/schematics/dt-component-e2e/index.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+    await addFixtureToTree(
+      initialTree,
+      'app.routing.module.ts.fixture.ts',
+      'apps/components-e2e/src/app/app.routing.module.ts',
+    );
+
+    initialFiles = initialTree.files;
+  });
+
+  it('should run', async () => {
```
lukasholzer, line 107: if you don't expect anything please remove it

libs/workspace/src/schematics/dt-component-e2e/index.ts
ffriedl89: Not joining might result in windows path issues - please keep the join!

libs/workspace/src/schematics/dt-component-e2e/index.ts
ffriedl89: @lukasholzer Please add more information on how to run an npm script as part of the schematic chain :) You have the most knowledge here :)

lukasholzer: You can take a look how nx is doing it [here](https://github.com/nrwl/nx/blob/master/packages/workspace/src/utils/rules/format-files.ts)

They are shipping this rule so just import it and add it in the end of the chain:

```

import { formatFiles } from '@nrwl/workspace';

..
    return chain([
      mergeWith(templateSource),
      addRoute(extendedOptions),
      formatFiles(),
    ]);
```

libs/workspace/src/schematics/dt-component-e2e/index.ts
bardosmisi: I planned to resolve this in the upcoming PR, but my comment was still pending, so it was not visible. Anyway I removed the comment from the code.

libs/workspace/src/schematics/dt-component-e2e/index.ts
bardosmisi: I'll remove this, thanks.

libs/workspace/src/schematics/dt-component-e2e/index.spec.ts
bardosmisi: Will be removed, thanks. My assumption was that if it throws an error during the test it will fail.

Issue #957, opened 2020-05-06
Bar indicator: Change theming to custom properties


tomheller:
Based on what we have learned from re-theming the button in #920, we should switch the theming of the bar indicator component to use custom properties and the new theming.

Bar indicator has the following states:

* main (affected by color-theme)
* recovered
* error

Goal of this issue is to remove our current theming approach of generating css classes for all themes and switch to css custom properties.

PR #1016, opened 2020-05-14
feat(bar-indicator): Changes styling to use custom properties

rowa-audil:
#### <strong>Pull Request</strong>

Changes bar indicator to use custom properties.
Fixes: #957

### DIFF
--- a/libs/barista-components/bar-indicator/src/_bar-indicator-theme.scss
+++ b/libs/barista-components/bar-indicator/src/_bar-indicator-theme.scss
@@ -1,20 +1,15 @@
-@import '../../core/src/theming/theming';
+:host.dt-color-main {
+  --dt-bar-indicator-background-color: var(--dt-theme-main-default-color);
+}
 
-@mixin dt-theme-bar-indicator($theme) {
-  $palette-names: 'main', 'recovered', 'error';
-  $default-palette: 'main';
+:host.dt-color-error {
+  --dt-bar-indicator-background-color: var(--dt-error-default-color);
+}
 
-  @each $name in $palette-names {
-    $palette: dt-get-theme-palette($theme, $name);
-    $default-color: dt-get-theme-color($palette, 'default');
+:host.dt-color-recovered {
+  --dt-bar-indicator-background-color: var(--dt-recovered-default-color);
+}
 
-    $namespace: ':host';
-    @if $name != $default-palette {
-      $namespace: ':host.dt-color-#{$name}';
-    }
-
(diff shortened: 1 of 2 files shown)


rowa-audil:
@NinaKammerlander 
I had to add some colors (e.g error) to our `main.scss` but I'm not sure if that was the right way.

libs/barista-components/style/main.scss
```diff
@@ -30,7 +30,12 @@
...
   --dt-warning-hover-color: #{$yellow-700};
   --dt-warning-active-color: #{$yellow-800};
+  --dt-error-default-color: #{$red-600};
+  --dt-error-hover-color: #{$red-700};
+  --dt-error-active-color: #{$red-800};
   --dt-cta-default-color: #{$green-600};
   --dt-cta-hover-color: #{$green-700};
   --dt-cta-active-color: #{$green-800};
+  --dt-bar-indicator-default-color: #{$royalblue-600};
+  --dt-bar-indicator-recovered-color: #{$green-600};
```
ffriedl89, line 40: Please remove here, since this is for globals only.

libs/barista-components/bar-indicator/src/_bar-indicator-theme.scss
```diff
@@ -1,20 +1,17 @@
...
-  $default-palette: 'main';
-
-  @each $name in $palette-names {
-    $palette: dt-get-theme-palette($theme, $name);
-    $default-color: dt-get-theme-color($palette, 'default');
+:host.dt-color-main {
+  .dt-bar-indicator-bar {
+    background-color: var(--dt-bar-indicator-default-color);
+  }
+}
```
ffriedl89, line 5: I'd suggest to restructure this file a bit. 


1. define a one css custom property for the dt-bar-indicator-background depending on the color set on the host
2. create one css class that sets the value of the background to the css custom property. 
```
.host.dt-color-main {
  --dt-bar-indicator-background-color: var(--dt-theme-main-default-color);
}

... the other colors the same way

:host .dt-bar-indicator-bar {
  background: var(--dt-bar-indicator-background-color);
}
```

Think about custom properties like having a placeholder in the standard selector that gets a value at runtime in the browser inherited from some parent.

rowa-audil: Is there a way to get the royalblue `--dt-theme-main-default-color` property for the main bar indicator, even if I'm not in its theme?
Changing the styles as suggested will change the main color to turquoise.

Issue #958, opened 2020-05-06
Breadcrumbs: Change theming to custom properties


tomheller:
Based on what we have learned from re-theming the button in #920, we should switch the theming of the breadcrumbs component to use custom properties and the new theming.

Breadcrumbs have the following states:

* main (affected by color-theme)
* error
* neutral (always gray-ish, regardless of color theme, scoped only to breadcrumbs)

In the following themes:

* light
* dark

Goal of this issue is to remove our current theming approach of generating css classes for all themes and switch to css custom properties.

PR #1022, opened 2020-05-15
feat(breadcrumbs): Changes styling to use custom properties

rowa-audil:

Fixes: #958


libs/barista-components/style/themes.scss
```diff
@@ -49,6 +61,12 @@
...
   --dt-warning-active-color: #{$yellow-800};
+  --dt-error-default-color: #{$red-600};
+  --dt-error-hover-color: #{$red-700};
+  --dt-error-active-color: #{$red-800};
+  --dt-recovered-default-color: #{$green-600};
+  --dt-recovered-hover-color: #{$green-700};
+  --dt-recovered-active-color: #{$green-800};
   --dt-cta-default-color: #{$green-600};
   --dt-cta-hover-color: #{$green-700};
   --dt-cta-active-color: #{$green-800};
```
tomheller, line 72: We are redefining these in every theme again and again. AS they should stay the same, maybe we want to move them into a mixin?

rowa-audil: Issue: #1056

ffriedl89: Thanks @rowa-audil for creating the issue #1056

PR #1023, opened 2020-05-15
fix(button): Fixes the regression that button colors changed due to t…

ffriedl89:
…he theme.

Fixes a regression introduced that caused purple buttons :D
Since all our buttons need to be turquoise if not set to cta or warning we need to use the scss variable in the dt-button-default-color custom property no matter the theme


tomheller:
@ffriedl89 should we release this fix with 6.5.2 as well?

ffriedl89:
> @ffriedl89 should we release this fix with 6.5.2 as well?

I'd say yes - we should release a patch for that.

Issue #970, opened 2020-05-07
Card: Change theming to custom properties


NinaKammerlander:
Based on what we have learned from re-theming the button in #920, we should switch the theming of the card-component to use custom properties and the new theming.

Card has the following themes:
- light
- dark

Goal of this issue is to remove our current theming approach of generating css classes for all themes and switch to css custom properties.

PR #1026, opened 2020-05-17
feat(card): Changes styling to use custom properties

rowa-audil:

Fixes: #970

### DIFF
--- a/libs/barista-components/card/src/_card-theme.scss
+++ b/libs/barista-components/card/src/_card-theme.scss
@@ -1,26 +1,15 @@
-@import '../../core/src/theming/theming';
+.dt-theme-dark {
+  .dt-card {
+    color: #ffffff;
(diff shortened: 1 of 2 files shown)


libs/barista-components/card/src/_card-theme.scss
```diff
@@ -1,26 +1,15 @@
-@import '../../core/src/theming/theming';
+* > .dt-theme-dark > * {
```
tomheller, line 1: I believe this selector could look a little different. This one seems odd. Can you explain your reasoning behind this?

In the button, this is also sufficient.
https://github.com/dynatrace-oss/barista/blob/4f3f47bc957d328611eb0ca1da95fc38cf526b11/libs/barista-components/button/src/_button-theme.scss#L54-L70

PR #1030, opened 2020-05-18
feat(checkbox): Changes styling to use custom properties

rowa-audil:

Fixes: #971

### DIFF
--- a/libs/barista-components/checkbox/src/_checkbox-theme.scss
+++ b/libs/barista-components/checkbox/src/_checkbox-theme.scss
@@ -1,40 +1,42 @@
-@import '../../core/src/style/variables';
-@import '../../core/src/theming/theming';
 @import '../../core/src/style/interactive-common';
+@import '../../core/src/style/variables';
 
-@mixin dt-theme-checkbox($theme) {
-  $is-dark: map-get($theme, 'is-dark');
+.dt-theme-dark {
+  :host {
+    .dt-checkbox-content {
+      color: #ffffff;
+    }
 
-  @if ($is-dark == true) {
-    :host {
-      .dt-checkbox-content {
-        color: #ffffff;
-      }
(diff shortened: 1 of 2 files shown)


libs/barista-components/checkbox/src/_checkbox-theme.scss
```diff
@@ -1,40 +1,42 @@
...
+      border-color: #ffffff;
+    }
+
+    .dt-checkbox-icon svg {
+      stroke: #ffffff;
+    }
 
+    &:hover {
       .dt-checkbox-container {
         border-color: #ffffff;
```
ffriedl89, line 20: The border-color is already set in L11

libs/barista-components/checkbox/src/_checkbox-theme.scss
```diff
@@ -1,40 +1,42 @@
...
+        background-color: rgba(255, 255, 255, 0.2);
       }
 
       .dt-checkbox-icon svg {
         stroke: #ffffff;
       }
```
ffriedl89, line 26: This is also in there twice now - please remove from hover styles

libs/barista-components/checkbox/src/_checkbox-theme.scss
rowa-audil: Well, the styling isn't correct if I don't set that twice. It will use the usual hover styling as if it isn't in the dark theme.

tomheller: @rowa-audil is correct here, the base hover is overwritten by the checkbox.scss:56 rule. We can either live with the duplication, or group the selectors: 
```scss
:host .dt-checkbox-container,
:host:hover .dt-checkbox-container {
  boder-color: #ffffff;
}
```

PR #1031, opened 2020-05-18
barista(builder): Fixes an issue with the index file generation.

### LINKED ISSUE #905: Barista: Server Side rendering of index page fails.
# Bug Report

Currently, all barista pages are static pages that are rendered on the server. Due to a race condition, the index.html file gets replaced with its original file in the builder. That is the reason why the index page is not server side rendered.

All other pages are fine.

## Possible solution.

Rename the output file to `index.original.html` and launch it for the generation. (This is already done)
* Fix the bug with the renaming.

lukasholzer:
Fixes #905

Fixes a problem where the index file was not generated propperly and therefore the server crashed on the healthcheck.

Added further tests to test the file renaming and rendering.

<hr>

  often helpful).
  guidelines.

#### Checklist

### DIFF
--- a/libs/workspace/src/test-setup.ts
+++ b/libs/workspace/src/test-setup.ts
@@ -13,3 +13,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+jest.mock('fs', () => {
+  const actualFs = jest.requireActual('fs');
+  const unionfs = require('unionfs').default;
+  unionfs.reset = () => {
+    // fss is unionfs' list of overlays
+    unionfs.fss = [actualFs];
+  };
+  return unionfs.use(actualFs);
+});
--- a/package.json
+++ b/package.json
@@ -181,6 +181,7 @@
     "tslint-plugin-prettier": "^2.3.0",
     "typescript": "^3.8.3",
     "typescript-tslint-plugin": "^0.5.0",
+    "unionfs": "^4.4.0",
     "vrsource-tslint-rules": "^6.0.0",
     "xml2js": "^0.4.22",
     "xmlbuilder": "^15.1.1",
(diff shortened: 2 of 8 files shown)


lukasholzer:
@rowa-audil FYI

libs/workspace/src/test-setup.ts
```diff
@@ -13,3 +13,13 @@
...
+
+jest.mock('fs', () => {
+  const actualFs = jest.requireActual('fs');
+  const unionfs = require('unionfs').default;
+  unionfs.reset = () => {
+    // fss is unionfs' list of overlays
+    unionfs.fss = [actualFs];
+  };
+  return unionfs.use(actualFs);
+});
```
ffriedl89, line 25: Can you please explain why we need unionfs here? :)

package.json
```diff
@@ -144,6 +144,7 @@
     "dotenv": "^8.2.0",
     "express": "^4.15.2",
     "extract-loader": "^5.0.1",
+    "fsevents": "^2.1.3",
```
ffriedl89, line 147: why is this added? Is it a peer dep of unionfs?

libs/workspace/src/test-setup.ts
lukasholzer: As stated in the specs - we merge the memes with the actual fs with unionfs so you can still imoprt and read files from your fs but it overlays the memfs. It tries all file operations first on the memfs and if it does not succeed it callbacks to the normal fs is required we cannot use only memfs as the builders need the actual fs to gather files etc.

image


This enables the usage of adding a memfs on top of the normal fs

```typescript
  const vol = Volume.fromJSON(
    {
      'index.original.html': 'original',
    },
    '/test-root',
  );

  // Merge the current fs with the mocked volume
  const fsMock: any = fs;
  fsMock.use(vol);
```

and we can get all the changes that are performed since this call on the fs.

package.json
lukasholzer: nope is a relict of a further try will remove it :)

Had it before unionfs to mock the fs but unionfs is way easier to handle

PR #1032, opened 2020-05-18
fix(core): Fixes issue with the `isNumber` util function

thomaspink:

Fixes issue where the `isNumber` util function also returns true if it is a number-string. Added `isNumberLike` function for these number-string cases.

BREAKING CHANGE: `isNumber` now returns false for number strings like `"12"` as expected. Added `isNumberLike` function for these number-string cases.

#### Type of PR

#### Checklist

### DIFF
--- a/libs/barista-components/core/src/util/type-util.spec.ts
+++ b/libs/barista-components/core/src/util/type-util.spec.ts
@@ -17,7 +17,13 @@
 // tslint:disable no-lifecycle-call no-use-before-declare no-magic-numbers
 // tslint:disable no-any max-file-line-count no-unbound-method use-component-selector
 
-import { isDefined, isEmpty, isNumber, isObject } from './type-util';
+import {
+  isDefined,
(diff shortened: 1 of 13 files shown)


libs/barista-components/core/src/util/type-util.spec.ts
```diff
@@ -127,6 +133,46 @@ describe('TypeUtil', () => {
...
+      expect(isNumberLike(0)).toBeTruthy();
+      expect(isNumberLike(1)).toBeTruthy();
+    });
+    it('should be false if the value is a string containing numbers', () => {
+      expect(isNumberLike('123test')).toBeFalsy();
+    });
+    it('should be false if the value is a string that cannot be converted', () => {
+      expect(isNumberLike('random')).toBeFalsy();
+    });
+    it('should be true if the value is a string can be converted', () => {
```
rowa-audil, line 156: NIT: `... is a string that can be converted`

PR #1049, opened 2020-05-19
chore(chart): Change DtChartSeries types to match supported chart types

ffriedl89:

This helps us have proper typing in other apps. Some other types that are derived from SeriesOptions don't have a data property which makes another check mandatory. Since all our consumers only have, column, bar, pie, area and area range this scopes it well enough


ffriedl89:
This is an additional change that should only be added to the master branch since it changes the type of the breaking change.

PR #1050, opened 2020-05-19
fix(table): Fixes an issue where the table tried to render after it was destroyed

### LINKED ISSUE #1046: [Table]: simple column comparator @Input causes error on navigation
# Bug Report
When setting a custom comparator function to a simple column via the template, and having this column sorting active, on page leave the following exception appears in the console:

`Error: "Cannot move a destroyed View in a ViewContainer!" `

## Expected Behavior

Custom comparator functions, when set in template should behave just like default sorting of columns or setting the comparator function via the dataSource.

## Current Behavior

Custom comparator functions, when set in template, cause an error when leaving the page.
Setting the comparator function via:
`dataSource.addComparatorFunction('colName', comparatorFn)`
is a possible workaround and does not cause the issue mentioned.

## Steps to Reproduce
1. Go to https://stackblitz.com/edit/github-bf3vmb
2. Sort "Memory combined" ASC (important, happens only in ascending in that example)
3. Click "Hide/Show" button

## Versions

**@dynatrace/barista-components**: 6.5.2

tomheller:

After adding the simple columns and adding the ngOnDestroy to the table,
we missed to add the super call the the cdkTable that handled the disconnecting from
the dataSource.
This has now been added again and the datasource is correctly disconnected.

Fixes #1046

#### Type of PR
Bugfix

#### Checklist

### DIFF
--- a/libs/barista-components/table/src/cell.ts
+++ b/libs/barista-components/table/src/cell.ts
@@ -168,7 +168,7 @@ export class DtCell implements AfterContentInit, OnDestroy {
         .subscribe((sort: DtSortEvent) => {
           // If event is void, it is being unregisterd.
           this._isSorted = sort.active === this._columnDef.name;
-          this._changeDetectorRef.detectChanges();
+          this._changeDetectorRef.markForCheck();
         });
     }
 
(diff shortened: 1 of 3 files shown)


ffriedl89:
cherry pick done! 🍒

PR #1069, opened 2020-05-20
Feat/radial chart relative values

### LINKED ISSUE #1053: Radial chart with selection and relative values
# Feature Request

## Summary

Radial chart is missing some features, like displaying percent values and being selectable

## Feature Description

* Selection should be possible. either by input or by click
* Selection is not always needed so it should be disabled by default
* Values in legend should also be toggled from absolute to percent

subarroca:
- Selection should be possible. either by input or by click
- Selection is not always needed so it should be disabled by default
- Values in legend should also be toggled from absolute to percent
- Responsive with min of 400px and max of 800px width

Closes #1053 

#### Checklist

### DIFF
--- a/libs/barista-components/radial-chart/src/radial-chart.scss
+++ b/libs/barista-components/radial-chart/src/radial-chart.scss
@@ -29,6 +29,10 @@ text {
   display: flex;
 }
 
+.dt-radial-chart-series-group {
+  outline: none;
+}
+
 :host.dt-radial-chart-legend-right .dt-radial-chart-wrapper {
   flex-direction: row;
   align-items: center;
--- a/libs/barista-components/radial-chart/src/radial-chart.ts
+++ b/libs/barista-components/radial-chart/src/radial-chart.ts
@@ -31,13 +31,14 @@ import {
   TemplateRef,
   ViewEncapsulation,
 } from '@angular/core';
-import {
-  DT_CHART_COLOR_PALETTE_ORDERED,
-  DtColors,
-} from '@dynatrace/barista-components/theming';
+import { DomSanitizer, SafeStyle } from '@angular/platform-browser';
 import { isNumber } from '@dynatrace/barista-components/core';
+import {
+  DtColors,
+  DT_CHART_COLOR_PALETTE_ORDERED,
+} from '@dynatrace/barista-components/theming';
 import { PieArcDatum } from 'd3-shape';
-import { combineLatest, of, Subject } from 'rxjs';
+import { merge, Subject } from 'rxjs';
 import { switchMap, takeUntil } from 'rxjs/operators';
 import { DtRadialChartOverlayData } from './radial-chart-path';
 import { DtRadialChartSeries } from './radial-chart-series';
@@ -47,10 +48,12 @@ import {
   generatePieArcData,
   getSum,
 } from './utils/radial-chart-utils';
-import { DomSanitizer, SafeStyle } from '@angular/platform-browser';
 
 /** Size of the inner (empty) circle in proportion to the circle's radius. */
 const DONUT_INNER_CIRCLE_FRACTION = 0.8;
+
+/** Minimum width of the chart */
+const MIN_WIDTH = 400;
 
 @Directive({
   selector: '[dtRadialChartOverlay]',
@@ -94,8 +97,11 @@ export class DtRadialChart implements AfterContentInit, OnDestroy {
     return this._maxValue;
   }
   set maxValue(value: number | null) {
-    this._maxValue = coerceNumberProperty(value);
-    this._updateRenderData();
+    const coercedValue = coerceNumberProperty(value);
+    if (coercedValue !== this._maxValue) {
+      this._maxValue = coercedValue;
+      this._updateRenderData();
+    }
   }
   private _maxValue: number | null = null;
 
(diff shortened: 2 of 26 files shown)


libs/barista-components/radial-chart/src/radial-chart.ts
```diff
@@ -109,6 +117,25 @@ export class DtRadialChart implements AfterContentInit, OnDestroy {
   }
   private _legendPosition: 'right' | 'bottom' = 'right';
 
+  /** Sets the display mode for the sunburst-chart values to either 'percent' or 'absolute'.  */
+  @Input()
+  set valueDisplayMode(value: 'absolute' | 'percent') {
+    this._valueAsAbsolute = value !== 'percent';
+  }
+  /** @internal Marks if absolute value should be shown or percent instead */
+  _valueAsAbsolute: boolean = true;
```
tomheller, line 126: do we need this value, it can be inferred by checking the valueDisplayMode against ` === 'absolute'`.

tomheller, line 126:
```suggestion
  @Input()
  valueDisplayMode: 'absolute' | 'percent' = 'absolute';
```

libs/barista-components/radial-chart/src/radial-chart.ts
```diff
@@ -47,11 +51,15 @@ import {
...
 } from './utils/radial-chart-utils';
-import { DomSanitizer, SafeStyle } from '@angular/platform-browser';
 
 /** Size of the inner (empty) circle in proportion to the circle's radius. */
 const DONUT_INNER_CIRCLE_FRACTION = 0.8;
 
+/** Minimum width of the chart */
+const MIN_WIDTH = 400;
+/** Maximum width of the chart */
+const MAX_WIDTH = 800;
```
tomheller, line 61: Why do we now clamp the size of the svg now? This was not part of the issue discription or the API proposal. 
The container should take care of the sizing. Please remove this again.

libs/barista-components/radial-chart/src/radial-chart.ts
```diff
@@ -213,6 +252,20 @@ export class DtRadialChart implements AfterContentInit, OnDestroy {
...
+  /** @internal gets the dimensions of the component and renders */
+  _updateDimensions(): void {
+    if (this._platform.isBrowser) {
+      this._width = Math.min(
+        MAX_WIDTH,
+        Math.max(
+          MIN_WIDTH,
+          this._elementRef.nativeElement.getBoundingClientRect().width,
+        ),
+      );
```
tomheller, line 264: Please do not clamp the width of the svg, the container should take care of the sizing. Please revert to the original behavior.

libs/barista-components/radial-chart/src/radial-chart.ts
```diff
@@ -174,37 +211,39 @@ export class DtRadialChart implements AfterContentInit, OnDestroy {
...
+        ),
+      ),
+    );
+
+    // initial segments
+    const initialSeries$ = merge(
+      ...this._radialChartSeries.map((series) => series._stateChanges$),
+    ).pipe(takeUntil(seriesChanges$));
+
+    merge(this._resizer.change(), initialSeries$, seriesChanges$)
```
tomheller, line 241: Nice cleanup, 👍

libs/barista-components/radial-chart/src/radial-chart.ts
```diff
@@ -174,37 +211,39 @@ export class DtRadialChart implements AfterContentInit, OnDestroy {
     private _platform: Platform,
     // TODO: remove this sanitizer when ivy is no longer opt out
     private _sanitizer: DomSanitizer,
+    private _resizer: DtViewportResizer,
```
tomheller, line 214: Non optional addition in the constructor parameters is considered a breaking change. 
As we do not need to clamp the svg sizes, I guess we can remove this one again and it will not block the progress of this one until barista version 8.

libs/barista-components/radial-chart/src/radial-chart.scss
```diff
@@ -9,10 +9,6 @@ text {
   font-size: 16px;
 }
 
-.dt-radial-chart-svg {
-  width: 100%;
-}
```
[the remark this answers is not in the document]
subarroca: Because now we don't expand it to the full width, it's limited between 400px and 800px, otherwise it's way too big

libs/barista-components/radial-chart/src/radial-chart.ts
```diff
@@ -109,6 +117,25 @@ export class DtRadialChart implements AfterContentInit, OnDestroy {
...
 
+  /** Sets the display mode for the sunburst-chart values to either 'percent' or 'absolute'.  */
+  @Input()
+  set valueDisplayMode(value: 'absolute' | 'percent') {
+    this._valueAsAbsolute = value !== 'percent';
+  }
+  /** @internal Marks if absolute value should be shown or percent instead */
+  _valueAsAbsolute: boolean = true;
+
+  /** Sets the display mode for the sunburst-chart values to either 'percent' or 'absolute'.  */
```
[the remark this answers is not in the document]
subarroca: 😅

libs/barista-components/radial-chart/src/radial-chart.ts
subarroca: As I remember other comments about removing logic from the template I did this (and also the ngIf you commented), but for me it's ok to remove this var

libs/barista-components/radial-chart/src/radial-chart.ts
subarroca: @tomheller I'll remove the clamping butI think I prefer to wait for barista 8 and include resizing of the chart. Before it was stuck with the initial dimensions

tomheller: The svg should have resized with the container. I'll have to check on monday if resizing is broken on the radial chart. Thanks for pointing this out.

tomheller: Resizing of the radial chart seems to work as expected on master. I don't think we need the viewPortResizer. 

May-25-2020 08-52-57

subarroca: Well, yes and no, it scales but it does not keep the width of the donut as it should be. But I'll remove it and review it in the future if need be

tomheller: yes, you are right. it scales the whole svg, which should be fine for now. 
With the viewport resizer, it would be possible to scale the donut-thickness acordingly

subarroca:
I'll make it unselectable from the outside, you're right.

About the hover, I confirmed it is intended as it is :)

ffriedl89:
Cherry pick done!

PR #1071, opened 2020-05-20
feat(checkbox-table): Add checkbox column and selection handling

dt-mafe:

<hr>

Early and rough draft for a checkbox column and respective selection handling in the table data source

#### Type of PR

#### Checklist

### DIFF
--- a/libs/barista-components/table/src/expandable/expandable-row.ts
+++ b/libs/barista-components/table/src/expandable/expandable-row.ts
@@ -170,6 +170,7 @@ export class DtExpandableRow extends DtRow
     elementRef: ElementRef,
   ) {
     super(elementRef);
+    this._table._hasExpandableRows = true;
     this._expansionDispatcher.listen((rowId, tableId) => {
       /**
        * If the table does not allow multiple rows to be expanded at a time,
--- a/libs/barista-components/table/src/selection/selector.spec.ts
+++ b/libs/barista-components/table/src/selection/selector.spec.ts
@@ -1,0 +1,327 @@
+/**
+ * @license
+ * Copyright 2020 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* tslint:disable:no-unused-variable */
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { Component, ViewChild, Predicate } from '@angular/core';
+import { DtTableModule } from '@dynatrace/barista-components/table';
+import { CommonModule } from '@angular/common';
+import { DtIconModule } from '@dynatrace/barista-components/icon';
+import { DtLoadingDistractorModule } from '@dynatrace/barista-components/loading-distractor';
+import { NoopAnimationsModule } from '@angular/platform-browser/animations';
+import { DtFormattersModule } from '@dynatrace/barista-components/formatters';
+import { HttpClientTestingModule } from '@angular/common/http/testing';
+import { By } from '@angular/platform-browser';
+import { DtCheckbox } from '@dynatrace/barista-components/checkbox';
+import { DtTableSelection, DT_TABLE_SELECTION_CONFIG } from './selection';
+import { createComponent } from '@dynatrace/testing/browser';
+
+describe('DtTableSelection', () => {
+  let component: DtTableSelectionComponentForTesting;
+  let fixture: ComponentFixture<DtTableSelectionComponentForTesting>;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      imports: [
+        CommonModule,
+        DtTableModule,
+        DtIconModule.forRoot({ svgIconLocation: `{{name}}.svg` }),
+        DtLoadingDistractorModule,
+        NoopAnimationsModule,
+        DtFormattersModule,
+        HttpClientTestingModule,
+      ],
+      // The providers are necessary here - because otherwise the TestBed does not allow
+      // overriding the providers in a different setup below
+      // The token needs to be already available initially
+      providers: [
+        {
+          provide: DT_TABLE_SELECTION_CONFIG,
+          useValue: {},
+        },
+      ],
+      declarations: [DtTableSelectionComponentForTesting],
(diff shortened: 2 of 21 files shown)


ffriedl89:
Talked to @dt-mafe and this is work in progress right now.

lukasholzer:
> Talked to @dt-mafe and this is work in progress right now.

@dt-mafe if it is work in progress would you be so kind to set the PR in the draft state?

image

dt-mafe:
I am sorry about not setting this PR back into draft state after the initial feedback, but this pull-request should be ready for review now.

libs/barista-components/table/src/expandable/expandable-row.ts
```diff
@@ -170,6 +170,7 @@ export class DtExpandableRow extends DtRow
     elementRef: ElementRef,
   ) {
     super(elementRef);
+    this._table._hasExpandableRows = true;
```
tomheller, line 173: @ffriedl89 How is this going to work combined with the `when` condition on rows? 
If there are expandable rows should initially, and the expression driving the `when` condition changes resulting in no expandable rows shown, how does that one reset on the table?

libs/barista-components/table/src/selection/selector.spec.ts
```diff
@@ -0,0 +1,328 @@
...
+      declarations: [DtTableSelectionComponentForTesting],
+    });
+  });
+
+  describe('without limit', () => {
+    beforeEach(() => {
+      TestBed.compileComponents();
+      fixture = TestBed.createComponent(DtTableSelectionComponentForTesting);
+      component = fixture.componentInstance;
+      fixture.detectChanges();
```
tomheller, line 64: Nitpick:
You could use the provided helper function for this. It creates the component and calls the first detect changes.
Add the import
```
import { createComponent } from '@dynatrace/testing/browser';
```

```suggestion
      fixture = createComponent(DtTableSelectionComponentForTesting);
      component = fixture.componentInstance;
```

libs/barista-components/table/src/selection/selectors.ts
```diff
@@ -0,0 +1,184 @@
...
+  </dt-cell>`,
+  styleUrls: ['./selectors.scss'],
+  encapsulation: ViewEncapsulation.None,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class DtTableRowSelector<T> implements OnDestroy {
+  /** The rows data used for selection */
+  @Input() row: T;
+
+  @Input() ariaLabel: '';
```
tomheller, line 52: We try to stay close to native code as long as possible, please rename the input to `aria-label` and do not assign a value. Angular will not render the attribute if it is undefined, which is more helpful than an empty aria-label string. 
```
  /** Aria label of the row selector checkbox. */
  @Input('aria-label') ariaLabel: string;
```

libs/barista-components/table/src/selection/selectors.ts
```diff
@@ -0,0 +1,184 @@
...
+        [indeterminate]="_isAnySelected"
+        [aria-label]="ariaLabel"
+      ></dt-checkbox>
+    </dt-header-cell>
+  `,
+  styleUrls: ['./selectors.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class DtTableHeaderSelector<T> implements OnDestroy {
+  @Input() ariaLabel: '';
```
tomheller, line 117: Same here, please rename the input to `aria-label` and do not assign a default value empty string.

libs/barista-components/table/src/table-data-source.ts
```diff
@@ -374,6 +373,61 @@ export class DtTableDataSource<T> extends DataSource<T> {
...
+  //       this._selection.selectable(data) &&
+  //       (this._selection.selectionLimit === undefined ||
+  //         this.selectionModel.selected.length < this._selection.selectionLimit))
+  //   );
+  // };
+
+  // private _getSelectableData(data: T[]): T[] {
+  //   return data.filter(this._isSelectable);
+  // }
+
```
tomheller, line 430: Please remove commented code.

libs/barista-components/table/src/table-data-source.ts
```diff
@@ -112,6 +110,7 @@ export class DtTableDataSource<T> extends DataSource<T> {
    */
   private _renderChangesSubscription = Subscription.EMPTY;
   private _searchChangeSubscription = Subscription.EMPTY;
+  private _selectionChangeSubscription = Subscription.EMPTY;
```
tomheller, line 113: I guess this is no longer used, right?
```suggestion
```

libs/barista-components/table/src/table-data-source.ts
```diff
@@ -463,6 +517,7 @@ export class DtTableDataSource<T> extends DataSource<T> {
   disconnect(): void {
     this._renderChangesSubscription.unsubscribe();
     this._searchChangeSubscription.unsubscribe();
+    this._selectionChangeSubscription.unsubscribe();
```
tomheller, line 520:
```suggestion
```

tomheller:
Hey @dt-mafe!
It seems there is something weird going on with your commit history. Before we merge this one, we might need to clean up the history a bit.

dt-mafe:
> Hey @dt-mafe!
> It seems there is something weird going on with your commit history. Before we merge this one, we might need to clean up the history a bit.
> 
> Please reach out if you need any help with the rebase / squashing of the commits.

Not really sure what happened there. Anyway, I cleaned up the git history a bit and rebased master again and it looks fine again.

However, bazel_stylelint build step is now failing and I don't really know why.
Errors are:
`setup_ci_build_cache.sh: 7: [: pull/1071: unexpected operator
Please specify the "BAZEL_REMOTE_CACHE_PASSWORD" environment variable when setting up remote  cache`
and
`ERROR: Could not open auth credentials file './tools/bazel_build_cache/cache-write-key.json': ./tools/bazel_build_cache/cache-write-key.json (No such file or directory)`
which I would not really suspect to be related to my changes.

tomheller:
@dt-mafe Thanks for cleaning up the git history. 

We are aware of the bazel stylelint errors. This is a configuration fault on our end nothing on your end broke this check. 
This is fine and we will review again soon 👍

libs/barista-components/table/src/expandable/expandable-row.ts
dt-mafe: The only reason this property seems to exist at the moment is for styling reasons since expandable rows use a 2px padding vs the 1px padding of the default rows which messes up the alignment of the header-checkbox with those in the table rows.
Note that this problem is not yet fixed and you would have to manually adjust styling when switching between expandable/none expandable rows in that situation to ensure perfect alignment of the header checkbox with the table. 

We could avoid this altogether if left/right paddings were consistent between expandable and none-expandable rows which would be possible since the expandable rows only really need the additional pixel on the top and bottom for the 2px thick border used there in the expanded state. Not sure if that could cause any other styling issues due top/bottom padding no longer being the same as left/right padding.

Issue #796, opened 2020-03-26
Barista: Refactor TOC and scrollspy


lukasholzer:
# Feature Request

Currently, the scroll spy logic and the table of contents generation is copied from angular.io and is not meant to be used with server-side rendering.
Furthermore, the logic is way to complex and overengineered.

We should simplify the logic in there and make it better universal compatible. (No window access allowed - SSR needs to work - ng build barista-design-system needs to pass) 

Maybe the TOC generation can be done in a transformer in the barista page builder and has not to be done on the client. This would save a lot of computing logic and makes barista more lightweight.

Another problem with the generated TOC is that it sometimes highlights the wrong element. The easiest way of reproducing this is to go to https://barista.lab.dynatrace.org/components/alert, click on 'Alerts in use' int the TOC, watch 'Do's and don'ts' gets highlighted.

Also there is a weird behavior of scroll jumps in case of a narrow mode barista. We were able to repro this behavior on the charts barista page in a narrow window mode. Try to go to the pie chart example, hover it a couple of times, and notice unintentional scrolling.

PR #1073, opened 2020-05-22
Refactor/toc and scrollspy

rowa-audil:
PR generates TOC server-side and only renders it in template and refactors scroll spy

Fixes: #796

### DIFF
--- a/apps/barista-design-system/src/pages/single-page/components/toc/toc.html
+++ b/apps/barista-design-system/src/pages/single-page/components/toc/toc.html
@@ -5,7 +5,6 @@
     [ngClass]="{ 'ba-is-active': _expandToc }"
   >
     <p
-      href="#"
       class="ba-toc-header ba-expandable-trigger ba-expandable-trigger-right"
       data-initialized="true"
     >
--- a/apps/barista-design-system/src/pages/single-page/components/toc/toc.html
+++ b/apps/barista-design-system/src/pages/single-page/components/toc/toc.html
@@ -16,9 +16,7 @@
         <li
           *ngFor="let headline of tocitems.headlines"
           class="ba-toc-item"
-          [class.ba-toc-item-active]="
-            _activeItems[0] && _activeItems[0].headlines[0] === headline
-          "
+          [class.ba-toc-item-active]="_activeItems.headlines[0] === headline"
         >
           <a
             #headline
@@ -34,7 +32,7 @@
               *ngFor="let subheadline of headline.children"
               class="ba-toc-item"
               [class.ba-toc-item-active]="
-                _currentActiveHeadline === subheadline
+                _activeItems.headlines[1] === subheadline
               "
             >
               <a
--- a/apps/barista-design-system/src/pages/single-page/components/toc/toc.html
+++ b/apps/barista-design-system/src/pages/single-page/components/toc/toc.html
@@ -13,33 +13,35 @@
     <div class="ba-toc-content ba-expandable-content">
       <ul class="ba-toc-list ba-list">
         <li
-          *ngFor="let heading of _headings$ | async"
+          *ngFor="let headline of tocitems.headlines"
           class="ba-toc-item"
           [class.ba-toc-item-active]="
-            _activeItems && _activeItems[0] === heading
+            _activeItems && _activeItems[0] === headline
           "
         >
           <a
             #headline
-            href="{{ heading.href }}"
+            href="#{{ headline.id }}"
             class="ba-toc-link"
             (click)="_handleTocClick($event)"
-            >{{ heading.title }}</a
           >
-          <ul *ngIf="heading.hasSubheadlines" class="ba-toc-ul">
+            {{ headline.headline }}</a
+          >
+          <ul *ngIf="headline.children" class="ba-toc-ul">
             <li
-              *ngFor="let subheading of heading.subheadlines"
+              *ngFor="let subheadline of headline.children"
               class="ba-toc-item"
               [class.ba-toc-item-active]="
-                _activeItems && _activeItems[1] === subheading
+                _activeItems && _activeItems[1] === subheadline
               "
             >
               <a
                 #headline
-                href="{{ subheading.href }}"
+                href="#{{ subheadline.id }}"
                 class="ba-toc-link"
                 (click)="_handleTocClick($event)"
-                >{{ subheading.title }}</a
+              >
+                {{ subheadline.headline }}</a
               >
             </li>
           </ul>
--- a/apps/barista-design-system/src/pages/single-page/components/toc/toc.html
+++ b/apps/barista-design-system/src/pages/single-page/components/toc/toc.html
@@ -4,23 +4,25 @@
     (click)="_expandTocMenu()"
     [ngClass]="{ 'ba-is-active': _expandToc }"
   >
-    <a
+    <p
       href="#"
       class="ba-toc-header ba-expandable-trigger ba-expandable-trigger-right"
       data-initialized="true"
-      >On this page</a
     >
+      On this page
+    </p>
     <div class="ba-toc-content ba-expandable-content">
       <ul class="ba-toc-list ba-list">
         <li
           *ngFor="let headline of tocitems.headlines"
           class="ba-toc-item"
           [class.ba-toc-item-active]="
-            _activeItems && _activeItems[0] === headline
+            _activeItems[0] && _activeItems[0].headlines[0] === headline
           "
         >
           <a
             #headline
+            id="{{ headline.id }}"
             href="#{{ headline.id }}"
             class="ba-toc-link"
             (click)="_handleTocClick($event)"
@@ -32,11 +34,12 @@
               *ngFor="let subheadline of headline.children"
               class="ba-toc-item"
               [class.ba-toc-item-active]="
-                _activeItems && _activeItems[1] === subheadline
+                _currentActiveHeadline === subheadline
               "
             >
               <a
                 #headline
+                id="{{ subheadline.id }}"
                 href="#{{ subheadline.id }}"
                 class="ba-toc-link"
                 (click)="_handleTocClick($event)"
--- a/apps/barista-design-system/src/pages/single-page/components/toc/toc.ts
+++ b/apps/barista-design-system/src/pages/single-page/components/toc/toc.ts
@@ -22,6 +22,7 @@ import {
   OnDestroy,
   NgZone,
   Inject,
+  Input,
 } from '@angular/core';
 import {
   BaTocService,
@@ -30,6 +31,7 @@ import {
 import { Subscription } from 'rxjs';
 import { Platform } from '@angular/cdk/platform';
 import { DOCUMENT } from '@angular/common';
+import { TOC } from '@dynatrace/shared/barista-definitions';
 
 @Component({
   selector: 'ba-toc',
@@ -40,6 +42,9 @@ import { DOCUMENT } from '@angular/common';
   },
 })
 export class BaToc implements OnInit, AfterViewInit, OnDestroy {
+  @Input()
+  tocitems: TOC;
+
   /** @internal all TOC entries */
   @ViewChildren('headline') _headlines;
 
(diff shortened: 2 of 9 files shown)


apps/barista-design-system/src/pages/single-page/components/toc/toc.html
```diff
@@ -4,42 +4,44 @@
...
-          [class.ba-toc-item-active]="
-            _activeItems && _activeItems[0] === heading
-          "
+          [class.ba-toc-item-active]="_activeItems.headlines[0] === headline"
         >
           <a
             #headline
-            href="{{ heading.href }}"
+            id="{{ headline.id }}"
+            href="#{{ headline.id }}"
```
lukasholzer, line 23: this does not work – you would link to your self and not to the headline in the content if you put there the id.

apps/barista-design-system/src/pages/single-page/components/toc/toc.html
```diff
@@ -4,42 +4,44 @@
...
               class="ba-toc-item"
               [class.ba-toc-item-active]="
-                _activeItems && _activeItems[1] === subheading
+                _activeItems.headlines[1] === subheadline
               "
             >
               <a
                 #headline
-                href="{{ subheading.href }}"
+                id="{{ subheadline.id }}"
```
lukasholzer, line 39: this cannot work you should reference here the id of the headline in the content

apps/barista-design-system/src/pages/single-page/components/toc/toc.ts
```diff
@@ -40,15 +40,16 @@ import { DOCUMENT } from '@angular/common';
   },
 })
 export class BaToc implements OnInit, AfterViewInit, OnDestroy {
+  @Input()
+  tocitems: TOC;
```
lukasholzer, line 44: NIT: `tableOfContentItems` – as stated in our coding standards we don't use abbreviations.

Furthermore please add JsDoc

libs/tools/barista/src/transform.ts
```diff
@@ -236,6 +237,44 @@ export const relativeUrlTransformer: BaPageTransformer = async (source) => {
...
+export const tocGenerator: BaPageTransformer = async (source) => {
+  const transformed = { ...source };
+  let toc: TOC = { headlines: [] };
+  if (source.toc) {
+    // generate TOC and at to source
+    // Find Headlines and corresponding subheadlines (h2 > h3)
+    transformed.content = runWithCheerio(source.content, ($) => {
+      const headlines = $('h2, h3');
+      let currentH2 = -1;
+      headlines.each((_, headline) => {
```
lukasholzer, line 249: NIT:

```suggestion
      headlines.each((, headline) => {
```

apps/barista-design-system/src/pages/single-page/components/toc/toc.html
```diff
@@ -4,42 +4,44 @@
...
           class="ba-toc-item"
-          [class.ba-toc-item-active]="
-            _activeItems && _activeItems[0] === heading
-          "
+          [class.ba-toc-item-active]="_activeItems.headlines[0] === headline"
         >
           <a
             #headline
-            href="{{ heading.href }}"
+            id="{{ headline.id }}"
```
tomheller, line 22:
```suggestion
            [attr.id]="headline.id"
```

libs/shared/barista-definitions/src/lib/barista-definitions.ts
```diff
@@ -155,3 +156,13 @@ export type BaContentTypes =
   | BaIndexPageContent
   | BaIconOverviewPageContent
   | BaCategoryNavigationContent;
+
+export interface TOC {
```
lukasholzer, line 160: In barista we prefix everything with `Ba` please add it in the sake of consistency.
And please stop using abbreviation as we agreed on this topic in our [coding standards](https://github.com/dynatrace-oss/barista/blob/master/CODING_STANDARDS.md#naming)

If this object is an array of Headlines please explain to me why it is needed – then the property can be of type `BaTableOfContentsHeadline[]` instead of `TOC`

```suggestion
export interface BaTableOfContents {
```

libs/shared/barista-definitions/src/lib/barista-definitions.ts
```diff
@@ -155,3 +156,13 @@ export type BaContentTypes =
...
+  headlines: Headline[];
+}
+
+export interface Headline {
```
lukasholzer, line 164:
```suggestion
export interface BaTableOfContentsHeadline {
```

apps/barista-design-system/src/pages/single-page/components/toc/toc.html
tomheller, line 39:
```suggestion
                [attr.id]="subheadline.id"
```

libs/tools/barista/src/transform.spec.ts
```diff
@@ -234,4 +235,27 @@ describe('Barista transformers', () => {
       );
     });
   });
+
+  describe('TOC Generator', () => {
+    it('should generate toc', async () => {
+      const content = `<h2 id=\"hl1\">hl1</h2>\n<h2 id=\"hl2\">hl2</h2><h3 id=\"shl1\">shl1</h3>`;
```
lukasholzer, line 241: Please add test scenario with underscore or dashes inside the id field.

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,81 +14,113 @@
...
-import { auditTime, distinctUntilChanged, takeUntil } from 'rxjs/operators';
+import { Injectable, Inject } from '@angular/core';
+import { Subject, Observable, fromEvent, ReplaySubject } from 'rxjs';
 import { Platform } from '@angular/cdk/platform';
+import { DOCUMENT } from '@angular/common';
+import { auditTime, takeUntil, distinctUntilChanged } from 'rxjs/operators';
 
 /**
  * CODE COPIED FROM: 'https://github.com/angular/angular/blob/master/aio/src/app/shared/scroll-spy.service.ts' and modified
  */
```
lukasholzer, line 25: Please remove this

apps/barista-design-system/src/pages/single-page/components/toc/toc.ts
```diff
@@ -22,14 +22,14 @@ import {
...
 } from '@angular/core';
-import {
-  BaTocService,
-  BaTocItem,
-} from '../../../../shared/services/toc.service';
+import { BaTocService } from '../../../../shared/services/toc.service';
 import { Subscription } from 'rxjs';
 import { Platform } from '@angular/cdk/platform';
 import { DOCUMENT } from '@angular/common';
+import { TOC } from '@dynatrace/shared/barista-definitions';
```
tomheller, line 32: When importing it from shared, please do not use abbreviations. `TableOfContents` is fine

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,81 +14,113 @@
...
+      .subscribe(() => {
+        this._onResize();
+      });
+    fromEvent(window, 'scroll')
+      .pipe(auditTime(10), takeUntil(this._onStopListening))
+      .subscribe(() => {
+        this._onScrollEvent();
+      });
+    this._onResize();
+    this._spiedElements.forEach((_) => {
```
lukasholzer, line 71:
```suggestion
    this._spiedElements.forEach(() => {
```

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,81 +14,113 @@
...
+  /**
+   * Stop tracking this group of elements and emitting active elements. If there is no other group
+   * being spied, stop listening for `resize` or `scroll` events.
+   */
+  unspy(): void {
+    this.activeScrollItem.complete();
+    this._spiedElements = [];
+    if (!this._spiedElements.length) {
+      this._onStopListening.next();
+    }
```
tomheller, line 87: This does not make a lot of sense. You reset the spiedElements to an array with length 0 and in line 86 check if the length is not 0. 
You should always next stop listening, as you do not differentiate if there are elements when you subscribe in lines 62 and 67

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,81 +14,113 @@
...
+
+    this._spiedElements.forEach((_) => {
+      this._onScroll(this._getScrollTop(), this._lastMaxScrollTop);
+    });
   }
 
-  /*
-   * Determine which element is the currently active one, i.e. the lower-most element that is
+  /**
+   * @internal Determine which element is the currently active one, i.e. the lower-most element that is
```
tomheller, line 116: No need for internal if it is private.
```suggestion
   * Determine which element is the currently active one, i.e. the lower-most element that is
```

libs/shared/barista-definitions/src/lib/barista-definitions.ts
```diff
@@ -155,3 +156,13 @@ export type BaContentTypes =
   | BaIndexPageContent
   | BaIconOverviewPageContent
   | BaCategoryNavigationContent;
+
+export interface TOC {
+  headlines: Headline[];
+}
```
tomheller, line 162: Does this need its own interface, it is basically `Headline[]`. I don't think it needs to be nested.

libs/tools/barista/src/transform.ts
```diff
@@ -236,6 +237,44 @@ export const relativeUrlTransformer: BaPageTransformer = async (source) => {
...
+            id: headlineId!,
+            headline: headlineText!,
+          });
+        } else if (headline.tagName === 'h3') {
+          // Add subheadlines array when needed
+          if (!toc.headlines[currentH2].children) {
+            toc.headlines[currentH2] = {
+              ...toc.headlines[currentH2],
+              children: [],
+            };
```
tomheller, line 264: No need for a spread here, this should suffice
```suggestion
            toc.headlines[currentH2].children = [];
```

tomheller:
> I would suggest a completely different API.
> 
> Create a component (component because directives cant be instantiated dynamically) that is a `BaTableOfContentSection` that is used like a directive with an attribute selector on every headline or section.
> 
> We generate our HTML on our own so it would be easy to get this kind of layout into the HTML.
> 
> ```
> <main baTableOfContentsArea>
> 
>   <h2 baTableOfContentSection="my-section-id">...<h2>
>   ...
>   <article  baTableOfContentSection="other-section-id">
> 
> </main>
> ```

I tend to agree with @lukasholzer on this one here, but it might not be as easy as you would imagine to manipulate the html, as this is stiched together from markdown, strapi and other resources. 

Additionally, there will still be some intersectionDetection with the viewport or the scrollheight necessary, as we need to highlight the items on the right hand side.

lukasholzer:
> > I would suggest a completely different API.
> > Create a component (component because directives cant be instantiated dynamically) that is a `BaTableOfContentSection` that is used like a directive with an attribute selector on every headline or section.
> > We generate our HTML on our own so it would be easy to get this kind of layout into the HTML.
> > ```
> > <main baTableOfContentsArea>
> > 
> >   <h2 baTableOfContentSection="my-section-id">...<h2>
> >   ...
> >   <article  baTableOfContentSection="other-section-id">
> > 
> > </main>
> > ```
> 
> I tend to agree with @lukasholzer on this one here, but it might not be as easy as you would imagine to manipulate the html, as this is stiched together from markdown, strapi and other resources.
> 
> Additionally, there will still be some intersectionDetection with the viewport or the scrollheight necessary, as we need to highlight the items on the right hand side.

Maybe we can create a discussion on the issue and a small API proposal regarding the changes.
Where we can collect all the input, concerns and observations.

rowa-audil:
API Proposal:

- Modify headline elements adding the BaTableOfContentSection to each.
- Instantiate those components and calculate bounding top properties for each (used by intersection observer)
- Start intersection observer:
  - Intersects when the headline is passed a point in the viewport
  - sets that item as active
- Highlight that element in the TOC

PR #1077, opened 2020-05-23
feat(alert): Changes styling to use custom properties.

### LINKED ISSUE #956: Alert: Change theming to custom properties
Based on what we have learned from re-theming the button in #920, we should switch the theming of the alert-component to use custom properties and the new theming.

Alert has the following states: 
* default (not affected by color-theme)
* warning
* error

In the following themes: 
* light
* dark

Goal of this issue is to remove our current theming approach of generating css classes for all themes and switch to css custom properties.

NinaKammerlander:

Changed the styling of the alert component to work with custom properties. closes #956.

<hr>

#### Checklist

### DIFF
--- a/libs/barista-components/alert/src/_alert-theme.scss
+++ b/libs/barista-components/alert/src/_alert-theme.scss
@@ -1,36 +1,20 @@
-@import '../../core/src/theming/theming';
+:host {
+  --dt-alert-default-color: var(--dt-error-default-color);
+}
 
-@mixin dt-theme-alert($theme) {
-  $error-palette: dt-get-theme-palette($theme, error);
-  $error-color: dt-get-theme-color($error-palette, default);
-  $warning-palette: dt-get-theme-palette($theme, warning);
-  $warning-color: dt-get-theme-color($warning-palette, default);
-  $is-dark: map-get($theme, 'is-dark');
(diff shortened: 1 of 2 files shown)


libs/barista-components/alert/src/_alert-theme.scss
```diff
@@ -1,36 +1,20 @@
...
+}
 
-  @if ($is-dark == true) {
-    :host.dt-alert-warning,
-    :host {
-      .dt-alert-text-container {
-        border-color: transparent;
-      }
-    }
+.dt-theme-dark {
```
ffriedl89, line 16:
```suggestion
:host-context(.dt-theme-dark) {
```

ffriedl89: We need to add the host context here otherwise the .dt-theme is scoped to this component. And the theming does not work in dark themes

libs/barista-components/alert/src/alert.scss
```diff
@@ -43,8 +43,3 @@
   padding: 10px 16px;
   flex-grow: 1;
 }
-
```
tomheller, line 46: Here as well, can we move the import of the theme file down here, to keep the source order intact.

Issue #1076, opened 2020-05-22
Switch to custom properties for theming


ffriedl89:
# Feature Request

We have a couple of components remaining that use the old theming approach. Based on what we have learned from re-theming the button in #920, we should switch the theming of all the following components. Please check afterwards if the behavior with themes is the same. Switch to dark theme and also use one other theme like purple to verify that the component behaves the same. 

This issue is a bundle of all components that have our theming and have not been migrated to custom properties. 

If you start working on a component please state this here and update the status once you are done. Also link the PRs with this issue. 

- [x] button-group
- [x] chart-heatfield #1357
- [x] consumption #1328 
- [x] copy-to-clipboard #1339
- [x] expandable-section #1710
- [ ] form-field
- [ ] highlight
- [ ] icon
- [x] indicator #1274
- [ ] info-group
- [x] progress-bar #1286 
- [x] progress-circle #1324 
- [x] radio #1367
- [x] show-more #1373
- [ ] switch
- [ ] row
- [ ] expandable-row
- [ ] tab-group
- [x] tile #1272
- [ ] toggle-button-item
- [ ] tree-table-row

ffriedl89:
Moved to internal issue tracking

PR #1084, opened 2020-05-26
feat(button-group): Changed styling to custom properties.

NinaKammerlander:

Changed the styling of the button-group component to work with custom properties. part of #1076
<hr>

#### Checklist

### DIFF
--- a/libs/barista-components/button-group/src/button-group-item.scss
+++ b/libs/barista-components/button-group/src/button-group-item.scss
@@ -2,7 +2,6 @@
 @import '../../core/src/style/variables';
 @import '../../core/src/style/interactive-common';
 @import '../../core/src/theming/theming';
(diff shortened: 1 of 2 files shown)


libs/barista-components/button-group/src/button-group-item.scss
```diff
@@ -46,8 +46,3 @@
     color: $gray-300;
   }
 }
-
-// Apply the button-group-item theme to all theme definitions
-@include dt-apply-theme() {
-  @include dt-theme-button-group-item($dt-current-theme);
-}
```
tomheller, line 53: If you import the theme at the top, it will change the order of the selectors. 
before it was only including a mixin, that was called when the themes were compiled. To keep the source order, you should import the theme at the end of this file.

PR #1093, opened 2020-05-28
chore: Added functionality to define output and formatters for design tokens.

tomheller:

Additionally switched the value definitions in  the themes from the meta key to the value object to
have this consistent and work with the dependency graph.

Adjusted the dependency-tree algorithm to be able to pick up sub-values like the ones in fonts and
themes.

#### Type of PR
Other

#### Checklist

### DIFF
--- a/libs/workspace/src/builders/design-tokens/build/token-converters/util/render-comment.ts
+++ b/libs/workspace/src/builders/design-tokens/build/token-converters/util/render-comment.ts
@@ -14,12 +14,21 @@
  * limitations under the License.
  */
 
-export const generateHeaderNoticeComment = (
-  commentStart = '/*',
-  commentLine = ' *',
-  commentEnd = ' */',
-) => `${commentStart}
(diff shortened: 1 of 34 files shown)


libs/workspace/src/builders/design-tokens/build/token-converters/util/render-comment.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  // If there is a comment defined, generate a multi line
+  // comment output
+  const comment = (prop.get('comment') as string)
+    .split('\n')
+    .map((line) => ` * ${line}`)
+    .join('\n');
+
+  return ['/*', comment, ' */'].join('\n');
```
lukasholzer, line 33: NIT: shouldn't a comment have two stars in the beginning?

https://jsdoc.app/about-getting-started.html

tomheller: Comment !== JSDoc
I guess a default token should be fine.

PR #1110, opened 2020-06-04
feat(scroll-spy): Refactor

rowa-audil:
This PR removes the toc-service and refactors the scroll spy
Additionally, toc items are rendered are generated in the transformer.

Fixes: #796


apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,120 @@
...
+          element.getBoundingClientRect().top -
+          this.OFFSET,
+        element: element,
       };
-    }
-
-    return null;
+      this.headlineElements.push(headlineElement);
+    });
+    this.headlineElements.sort((a, b) => compareValues(a.top, b.top, 'asc'));
```
lukasholzer, line 90:
```typescript
this.headlineElements = elements
  .map((element) => ({
    top: this._getScrollTop() + element.getBoundingClientRect().top - this.OFFSET,
    element,
  }))
  .sort((a,b) =>  compareValues(a.top, b.top, 'asc'));
```

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,120 @@
...
   }
 
-  private _getContentHeight(): number {
-    return this._document.body.scrollHeight || Number.MAX_SAFE_INTEGER;
+  /**
+   * Evaluates by comparing the users scroll position with the element top property and return the id of an element
+   * that should be highlighted
+   */
+  findActiveItem(): string {
```
lukasholzer, line 97:
```suggestion
  findActiveItem(): string | null {
```

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,120 @@
...
-
-    const scrollTop = this._getScrollTop();
-    const maxScrollTop = this._lastMaxScrollTop;
-
-    for (const group of this._spiedElementGroups) {
-      group.onScroll(scrollTop, maxScrollTop);
-    }
+  /** Height of the whole content */
+  private _getContentHeight(): number {
+    return this._document.body.scrollHeight || Number.MAX_SAFE_INTEGER;
```
lukasholzer, line 126: Why do we use the max integer here? I think no document is that long. Please explain the intention here.

libs/shared/barista-definitions/src/lib/barista-definitions.ts
```diff
@@ -182,3 +182,15 @@ export interface BaUxdEdge {
   uxd_node: number; // uxd_node id
   selected?: boolean;
 }
+
+export interface BaUxdEdge {
```
lukasholzer, line 186: How is the BaUxEdge connected with the scrollspy?

libs/tools/barista/src/transform.ts
```diff
@@ -236,6 +237,65 @@ export const relativeUrlTransformer: BaPageTransformer = async (source) => {
...
+export const tableOfContentGenerator: BaPageTransformer = async (source) => {
+  const transformed = { ...source };
+  let toc: TableOfContents[] = [];
+  if (source.toc) {
+    // generate TOC and at to source
+    // Find Headlines and corresponding subheadlines (h2 > h3)
+    transformed.content = runWithCheerio(source.content, ($) => {
+      const headlines = $('h2, h3');
+      let currentH2 = -1;
+      // TODO: Id is lost somewhere? --------------!
```
lukasholzer, line 269: is this todo done?

libs/tools/barista/src/transform.ts
```diff
@@ -236,6 +237,65 @@ export const relativeUrlTransformer: BaPageTransformer = async (source) => {
...
+      headlines.each((_index, headline) => {
+        const headlineId = $(headline).attr('id');
+        const headlineText = $(headline).text();
+        if (headline.tagName === 'h2') {
```
lukasholzer, line 273: Is it always a h2 and a h3?

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,120 @@
...
+    this.calcTopValues(elements);
+    // Resize and Scroll trigger event calculates top values and active item.
+    fromEvent(window, 'resize')
+      .pipe(auditTime(300))
+      .subscribe(() => {
+        this.calculateActiveItem(elements);
+      });
+    fromEvent(window, 'scroll')
+      .pipe(auditTime(10))
+      .subscribe(() => {
```
lukasholzer, line 65: please unsubscribe from this subscription

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
rowa-audil: It's not needed. Leftover from the old version and I overlooked it

libs/shared/barista-definitions/src/lib/barista-definitions.ts
rowa-audil: Good question :D
That's probably a side effect of reading the master into this branch.

libs/tools/barista/src/transform.ts
rowa-audil: Yap every page is structured that way.
Including h4s would explode the toc since h4s are used to section a subsection

apps/barista-design-system/src/pages/single-page/components/toc/toc.ts
```diff
@@ -93,12 +95,15 @@ export class BaToc implements OnInit, AfterViewInit, OnDestroy {
...
+    const targetId = (ev.currentTarget as HTMLAnchorElement).hash;
+    const target = this._document.querySelector(targetId);
 
     if (this._platform.isBrowser && target) {
       // Has to be set manually because of preventDefault() above.
-      window.location.hash = targetId || '';
+      const newLocation =
+        window.location.origin + window.location.pathname + targetId;
+      window.history.replaceState('', '', newLocation);
+
```
ffriedl89, line 106: Please add a comment why we cannot do this with the angular router right now

apps/barista-design-system/src/shared/services/scroll-spy.e2e.ts
```diff
@@ -0,0 +1,27 @@
...
+import { Selector } from 'testcafe';
+
+fixture('Scroll Spy').page('http://localhost:4200/components/chart');
+
+test('should highlight correct headline after scrolling', async (testController: TestController) => {
+  await testController
+    .wait(200)
+    .click('h2[id="imports"')
+    .expect(Selector('h2[id="imports"]').getStyleProperty('background'))
+    .eql('rgb(82, 108, 255)');
```
ffriedl89, line 26: please check whether an active class is present `ba-toc-item-active`

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,123 @@
...
+ * currently active and should be highlighted.
+ * Save that item into a stream and the toc component then subscribes to it.
+ * The TOC component then handles the check which item to highlight
  */
-export class BaScrollSpiedElement implements BaScrollItem {
-  top = 0;
+@Injectable({ providedIn: 'root' })
+export class BaScrollSpyService {
+  /** Top Offset */
+  OFFSET = 66;
```
ffriedl89, line 39: Can be moved to a constant outside of the class.
```
/**
* The offset from the top of the viewport that triggers an toc item to be active
* This enables us to have an item becoming active although its not on the very top of the viewport
* e.g. in case of a sticky navigation 0 offset would not work
*/
const ACTIVE_ITEM_TOP_OFFSET = 66;
```

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,123 @@
...
+      .pipe(auditTime(300))
+      .subscribe(() => {
+        this.calculateActiveItem(elements);
+      });
+    fromEvent(window, 'scroll')
+      .pipe(auditTime(10))
+      .subscribe(() => {
+        this.calculateActiveItem(elements);
+      });
+    return this.activeItemId$.asObservable().pipe(distinctUntilChanged());
```
ffriedl89, line 68: Unused return value

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,123 @@
...
+    // Initially calculate boundclientrect top values for every headline.
+    this.calcTopValues(elements);
+    // Resize and Scroll trigger event calculates top values and active item.
+    fromEvent(window, 'resize')
+      .pipe(auditTime(300))
+      .subscribe(() => {
+        this.calculateActiveItem(elements);
+      });
+    fromEvent(window, 'scroll')
+      .pipe(auditTime(10))
```
ffriedl89, line 64:
```suggestion
    merge(fromEvent(window, 'resize')
      .pipe(auditTime(300)),  fromEvent(window, 'scroll')
      .pipe(auditTime(10)))
      .subscribe(() => {
        this.calculateActiveItem(elements);
      });
   
```

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,123 @@
...
-  /*
-   * Calculate the `top` value of each ScrollSpiedElement of this group (based on te current
-   * `scrollTop` and `topOffset` values), so that the active element can be later determined just by
-   * comparing its `top` property with the then current `scrollTop`.
-   */
-  calibrate(scrollTop: number, topOffset: number): void {
-    for (const spiedElem of this._spiedElements) {
-      spiedElem.calculateTop(scrollTop, topOffset);
+  /** Calculates top values and finds the active item putting it into a stream */
+  calculateActiveItem(elements: Element[]): void {
```
ffriedl89, line 72: Make private

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,123 @@
...
+  ) {}
 
-  constructor(elements: Element[]) {
-    this._spiedElements = elements.map(
-      (elem, i) => new BaScrollSpiedElement(elem, i),
-    );
+  /** Start spying on an element array of headlines returning a stream with the active item */
+  spyOn(elements: Element[]): Observable<string> {
+    // Initially calculate boundclientrect top values for every headline.
+    this.calcTopValues(elements);
```
ffriedl89, line 56: Check whether you still need this

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,123 @@
...
-        if (spiedElem.top <= scrollTop) {
-          return true;
-        }
-        return false;
-      });
-    }
-
-    this.activeScrollItem.next(activeItem || null);
+  /** Calculates top bounding properties for an element array */
+  calcTopValues(elements: Element[]): void {
```
ffriedl89, line 81: Make private

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,123 @@
...
+    this.headlineElements = [];
```
ffriedl89, line 82:
```suggestion
```
Can be omitted since map always gives you a new array anyway

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,123 @@
...
-          this._onResize();
-        });
-      fromEvent(window, 'scroll')
-        .pipe(auditTime(10), takeUntil(this._onStopListening))
-        .subscribe(() => {
-          this._onScroll();
-        });
-      this._onResize();
-
+  findActiveItem(): string | null {
```
ffriedl89, line 98:
```suggestion
  private _findActiveItemId(): string | null {
```

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,123 @@
...
-          .pipe(distinctUntilChanged()),
-        unspy: () => {
-          this._unspy(spiedGroup);
-        },
-      };
-    }
+      const contentHeight = this._getContentHeight();
+      const viewportHeight = this._getViewportHeight();
+      for (let i = 0; i < this.headlineElements.length; i++) {
+        console.log(scrollTop + viewportHeight >= contentHeight);
```
ffriedl89, line 106:
```suggestion
```

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,123 @@
...
+        // Check whether an elements top value is smaller then the users scroll top position and offset
+        // resulting in setting that value as `active`
+        if (this.headlineElements[i].top <= scrollTop + this.OFFSET) {
+          activeItem = this.headlineElements[i].element.id;
+        } else if (scrollTop + viewportHeight >= contentHeight) {
+          // Special case when user is at the bottom of the page and there's no way tthe last item will be active
+          activeItem = this.headlineElements[this.headlineElements.length - 1]
+            .element.id;
+          console.log(activeItem, this.headlineElements);
```
ffriedl89, line 115:
```suggestion

```

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,123 @@
...
-      fromEvent(window, 'scroll')
-        .pipe(auditTime(10), takeUntil(this._onStopListening))
-        .subscribe(() => {
-          this._onScroll();
-        });
-      this._onResize();
-
+  findActiveItem(): string | null {
+    // The element id of the item to be highlighted
+    let activeItem: string | null = null;
```
ffriedl89, line 100:
```suggestion
    let activeItemId: string | null = null;
```

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,123 @@
...
-        },
-      };
-    }
+      const contentHeight = this._getContentHeight();
+      const viewportHeight = this._getViewportHeight();
+      for (let i = 0; i < this.headlineElements.length; i++) {
+        console.log(scrollTop + viewportHeight >= contentHeight);
+        // Check whether an elements top value is smaller then the users scroll top position and offset
+        // resulting in setting that value as `active`
+        if (this.headlineElements[i].top <= scrollTop + this.OFFSET) {
```
ffriedl89, line 109: I don't think you need to remove the offset up top and add it here. I guess removing it when assigning the headlineElements is enough or do it here by adding the offset here

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,123 @@
...
-          .asObservable()
-          .pipe(distinctUntilChanged()),
-        unspy: () => {
-          this._unspy(spiedGroup);
-        },
-      };
-    }
+      const contentHeight = this._getContentHeight();
+      const viewportHeight = this._getViewportHeight();
+      for (let i = 0; i < this.headlineElements.length; i++) {
```
ffriedl89, line 105: Maybe a while loop is better suited here since you don't have to continue the loop once you have your active item.

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,123 @@
...
-        .pipe(auditTime(10), takeUntil(this._onStopListening))
-        .subscribe(() => {
-          this._onScroll();
-        });
-      this._onResize();
-
+  findActiveItem(): string | null {
+    // The element id of the item to be highlighted
+    let activeItem: string | null = null;
+    this._zone.runOutsideAngular(() => {
```
ffriedl89, line 101: Can be removed - no need since you don't register event listeners here

apps/barista-design-system/src/pages/single-page/single-page.ts
```diff
@@ -67,11 +69,16 @@ export class BaSinglePage implements OnInit, AfterViewInit {
   ngAfterViewInit(): void {
     if (this._platform.isBrowser) {
       const allTables = [].slice.call(this._document.querySelectorAll('table'));
-
+      const headlines = [].slice.call(
+        this._document.querySelectorAll(
+          'h2[contentsection], h3[contentsection]',
+        ),
+      );
```
ffriedl89, line 76:
```suggestion
        this._document.querySelectorAll(
          '.ba-page h2, .ba-page h3',
        ),
      );
```

libs/tools/barista/src/transform.ts
```diff
@@ -236,6 +237,66 @@ export const relativeUrlTransformer: BaPageTransformer = async (source) => {
...
+      // console.log(sectionId);
+      $(element.attribs).append(
+        $(element).attr('contentSection', ''),
+        $(element).attr('id', sectionId),
+      );
+    });
+  });
+
+  return transformed;
+};
```
ffriedl89, line 259: check whether id's on headlines are already covered

libs/tools/barista/src/transform.ts
```diff
@@ -236,6 +237,66 @@ export const relativeUrlTransformer: BaPageTransformer = async (source) => {
...
+
+/** generates a toc json containing headlines with children */
+export const tableOfContentGenerator: BaPageTransformer = async (source) => {
+  const transformed = { ...source };
+  let toc: TableOfContents[] = [];
+  if (source.toc) {
+    // generate TOC and at to source
+    // Find Headlines and corresponding subheadlines (h2 > h3)
+    transformed.content = runWithCheerio(source.content, ($) => {
+      const headlines = $('h2, h3');
```
ffriedl89, line 269:
```suggestion
      const headlines = $('.ba-page h2, .ba-page h3');
```
If we scope this to the body of the page - we avoid side-effects of h2 and h3 appearing in the toc that are maybe in the footer or header section outside the ba-page. And we can remove the section transformer.

apps/barista-design-system/src/pages/single-page/components/toc/toc.html
```diff
@@ -4,42 +4,50 @@
...
       <ul class="ba-toc-list ba-list">
         <li
-          *ngFor="let heading of _headings$ | async"
+          *ngFor="let headline of tocItems"
           class="ba-toc-item"
           [class.ba-toc-item-active]="
-            _activeItems && _activeItems[0] === heading
+            _activeItems &&
+            _activeItems[0] &&
+            _activeItems[0].id === headline.id
```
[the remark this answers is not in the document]
rowa-audil: **Angular throws this:** 
Parser Error: Conditional expression _activeItems?[0]?.id === headline.id requires all 3 expressions at the end of the expression [_activeItems?[0]?.id === headline.id] in /Users/rowa.audil/Documents/Projects/GitHub/Barista-Repo/barista/apps/barista-design-system/src/pages/single-page/components/toc/toc.html@17:38

libs/tools/barista/src/transform.ts
rowa-audil: Yes, they are.

lukasholzer:
build of barista is still failing so I guess it is still WIP

libs/tools/barista/src/transform.ts
rowa-audil: There's no need for that. The `source.content` coming in only includes the headlines from the corresponding components markdown.
Sorry, I should have clarified that.

apps/barista-design-system/src/pages/single-page/components/toc/toc.html
```diff
@@ -4,42 +4,50 @@
...
                 #headline
-                href="{{ subheading.href }}"
+                id="{{ subheadline.id }}"
+                href="#{{ subheadline.id }}"
                 class="ba-toc-link"
                 (click)="_handleTocClick($event)"
-                >{{ subheading.title }}</a
+              >
+                {{ subheadline.headline }}</a
               >
```
lukasholzer, line 51: First of all we are using the router and we should leverage the routing instead of using hrefs that trigger full page reloads. https://angular.io/api/router/RouterLink#description

```html
<a
   [routerlink]="['/']"
   [fragment]="subheadline.id"
   [attr.id]="subheadline.id"
>{{ subheadline.headline }}</a>
```

But as far as I can remember we have our own router link component that is working with serverside rendering.

We don't want to trigger a page reload when clicking on this.

> Small Tipp: never do this in Angular `id="{{ subheadline.id }}"` always use bindings!

https://angular.io/guide/template-syntax#attribute-binding

apps/barista-design-system/src/pages/single-page/components/toc/toc.ts
```diff
@@ -39,46 +35,44 @@ import { DOCUMENT } from '@angular/common';
...
-      const docElement = this._document.getElementById('main') || undefined;
-      this._tocService.genToc(docElement);
-    });
-  }
-
-  ngOnDestroy(): void {
-    this._activeItemsSubscription.unsubscribe();
+    this._scrollSpyService.activeItemId$
+      .asObservable()
+      .subscribe((activeItemId) => {
```
lukasholzer, line 61: please unsubscribe from this subscription!

apps/barista-design-system/src/pages/single-page/components/toc/toc.ts
```diff
@@ -93,12 +87,17 @@ export class BaToc implements OnInit, AfterViewInit, OnDestroy {
...
+    const targetId = (ev.currentTarget as HTMLAnchorElement).hash;
+    const target = this._document.querySelector(targetId);
 
     if (this._platform.isBrowser && target) {
       // Has to be set manually because of preventDefault() above.
-      window.location.hash = targetId || '';
+      // We shouldn't use angular router since that would build the page again by setting its metadata again.
+      // By manipulating the url manually like below we avoid unnecassery page reloads.
+      const newLocation =
+        window.location.origin + window.location.pathname + targetId;
```
lukasholzer, line 98: Please inject the location and not access it via the window. Then the platform check should not be needed.

https://stackblitz.com/edit/angular-scroll-spy-routing?file=app%2Fhappy-scrolling.component.ts

apps/barista-design-system/src/pages/single-page/components/toc/toc.html
```diff
@@ -4,42 +4,50 @@
...
-            href="{{ heading.href }}"
+            id="{{ headline.id }}"
+            href="#{{ headline.id }}"
             class="ba-toc-link"
             (click)="_handleTocClick($event)"
-            >{{ heading.title }}</a
           >
-          <ul *ngIf="heading.hasSubheadlines" class="ba-toc-ul">
+            {{ headline.headline }}
+          </a>
```
lukasholzer, line 32: here the same please use our own routerlink

lukasholzer:
Maybe take a look at this stackblitz it can be so dead simple https://stackblitz.com/edit/angular-scroll-spy-routing?file=app%2Fhappy-scrolling.component.ts

apps/barista-design-system/src/pages/single-page/components/toc/toc.html
```diff
@@ -4,42 +4,50 @@
...
-                _activeItems && _activeItems[1] === subheading
+                _activeItems &&
+                _activeItems[1] &&
+                _activeItems[1].id === headline.id
               "
             >
               <a
                 #headline
-                href="{{ subheading.href }}"
+                id="{{ subheadline.id }}"
```
tomheller, line 45:
```suggestion
                [id]="subheadline.id"
```

apps/barista-design-system/src/pages/single-page/components/toc/toc.html
tomheller: We have run into a lot of problems when using the routerLink (either our own or the one from angular) here, as it will always trigger the canActivate in the guard. The guard will reset the input in the page-service, which will force angular to rerender the whole page. I'd rather not use the routerLink here.

apps/barista-design-system/src/pages/single-page/components/toc/toc.ts
tomheller: Instead of push in the new location and manually scrolling there, we can use the injected location and leverage `location.go`, that should work around the router problems we had and we can drop the platform checks.

Suggestion, while switching the element in the template away from a an `<a href="target"...` to something like: `span (click)="_handleToClick(target)"`

```ts
/** @internal handle the click on a TOC item */
  _handleTocClick(target: string): void {
    this._location.go(target);
  }

```
See: https://angular.io/api/common/Location#go

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,127 @@
...
+    merge(
+      fromEvent(window, 'resize').pipe(
+        auditTime(300),
+        takeUntil(this._onStopListening),
+      ),
+      fromEvent(window, 'scroll').pipe(
+        auditTime(10),
+        takeUntil(this._onStopListening),
+      ),
+    ).subscribe(() => {
```
tomheller, line 63:
```suggestion
    merge(
      fromEvent(window, 'resize').pipe(
        auditTime(300),
      ),
      fromEvent(window, 'scroll').pipe(
        auditTime(10),
      ),
    )
    .pipe(takeUntil(this._onStopListening))
    .subscribe(() => {
```

apps/barista-design-system/src/shared/services/scroll-spy.service.ts
```diff
@@ -14,212 +14,127 @@
...
+    const activeItemId = this.findActiveItemId();
+    if (activeItemId) {
+      this.activeItemId$.next(activeItemId);
+    }
   }
 
-  private _getViewportHeight(): number {
-    return (this._platform.isBrowser && window.innerHeight) || 0;
+  /** Calculates top bounding properties for an element array */
+  private _calcTopValues(elements: Element[]): void {
```
tomheller, line 78: NIT: Coding guidelines state to avoid abbreviations.
```suggestion
  private _calculateTopValues(elements: Element[]): void {
```

apps/barista-design-system/src/pages/single-page/components/toc/toc.ts
rowa-audil: Added unsubscribe by calling `complete()` on the subject in the service when `unspy()` is called. Which is called when destroying the single-page component.

apps/barista-design-system/src/pages/single-page/components/toc/toc.html
```diff
@@ -4,42 +4,48 @@
...
+            _activeItems &&
+            _activeItems[0] &&
+            _activeItems[0].id === headline.id
           "
         >
-          <a
+          <span
             #headline
-            href="{{ heading.href }}"
+            [id]="headline.id"
```
tomheller, line 26: you are duplicating the id's on the page here. If there is a headline with this id, I don't think you will need to bind the id on this one here.
Sorry, this was my bad on the last review.

apps/barista-design-system/src/pages/single-page/components/toc/toc.html
```diff
@@ -4,42 +4,48 @@
...
+                _activeItems &&
+                _activeItems[1] &&
+                _activeItems[1].id === headline.id
               "
             >
-              <a
+              <span
                 #headline
-                href="{{ subheading.href }}"
+                [id]="subheadline.id"
```
tomheller, line 44: Same here, there probably is already an element with that id, if you bind the id again you duplicate the id on the page.
I don't think we need the ID binding here.

apps/barista-design-system/src/pages/single-page/components/toc/toc.ts
```diff
@@ -87,23 +87,15 @@ export class BaToc implements OnInit, AfterViewInit, OnDestroy {
...
-    ev.preventDefault();
-    ev.stopImmediatePropagation();
-    const targetId = (ev.currentTarget as HTMLElement).getAttribute('href');
-    const target = this._document.querySelector(targetId || '');
-
-    if (this._platform.isBrowser && target) {
-      // Has to be set manually because of preventDefault() above.
-      window.location.hash = targetId || '';
+  _handleTocClick(target: string): void {
+    this._location.onUrlChange(() => {
```
tomheller, line 91: You are adding this listener on every click. I don't think we need that one at all. Does location.go not scroll to the correct anchored position already?
It might have been thrown off by the duplicate ids on the page, see comments above.

rowa-audil: I removed the listener and the scroll animation. `location.go()` does not scroll. Also not stated like that in the documentation.
FYI: Changes the browser's URL to a normalized version of a given URL, and pushes a new item onto the platform's history.

apps/barista-design-system/src/pages/single-page/components/toc/toc.ts
```diff
@@ -87,23 +86,13 @@ export class BaToc implements OnInit, AfterViewInit, OnDestroy {
...
-    if (this._platform.isBrowser && target) {
-      // Has to be set manually because of preventDefault() above.
-      window.location.hash = targetId || '';
-      requestAnimationFrame(() => {
-        target.scrollIntoView({
-          behavior: 'smooth',
-        });
+  _handleTocClick(target: string): void {
+    this._location.go(`${this._location.path()}#${target}`);
+    const targetHeadline = this._document.querySelector(`#${target}`);
```
tomheller, line 91: NIT: I know this should not happen, but with this selector we could potentially limit id clashes that may occur within the document. 
```suggestion
    const targetHeadline = this._document.querySelector(`h2[id="${target}"], h3[id="${target}"]`);
```

apps/barista-design-system/src/pages/single-page/single-page.ts
```diff
@@ -80,23 +80,15 @@ export class BaSinglePage implements OnInit, AfterViewInit, OnDestroy {
...
         applyTableDefinitionHeadingAttr(table);
       }
-      this._scrollSpyService.spyOn(headlines);
     }
   }
 
-  ngOnDestroy(): void {
-    this._scrollSpyService.unSpy();
-  }
+  ngOnDestroy(): void {}
```
ffriedl89, line 91: NIT: Can be removed

Issue #1057, opened 2020-05-20
Design Tokens - Requirements gathering


tomheller:
We are planning to include Design Tokens into our library for everyone to use. If you are not familiar with design tokens, you can use these resources to familiarize yourself with the subject, but the gist of it is to have a central definition for every spacing / color / border / radius / opacity / font / ... used within the design language to make the design more consistent.

* https://spectrum.adobe.com/page/design-tokens/
* https://css-tricks.com/what-are-design-tokens/
* https://www.lightningdesignsystem.com/design-tokens/

We already have started toying with design tokens and you can find our [concept specification within the repo](https://github.com/dynatrace-oss/barista/blob/master/concepts/design-tokens/README.md).

So far, we are have a couple of design tokens defined with dummy values like spacing, colors and some for the web-component based button. These design tokens currently generate the following output variable(formats):

**sass variables**
Variable definitions that can be used within your sass/scss environment, for example: `$palette-color-key-100`. 

**css custom properties**
Variable definitions that can be used within your css environment, for example: `var(--palette-color-key-100)`. 

**Typescript constants**
Currently for completions sake. Sometimes one does need a specific color or spacing definition in your scripts and it is good to know that we can keep them in sync via the design tokens definition, for example: `export const PALETTE_COLOR_KEY_100='#ff00ff';`

## Help wanted
As we are ramping up speed with the design tokens, we now want your input on what you would expect from these. What outputs would help you to use the tokens within your (angular) application / prototyping setup / ect...

Please let us know and we will definitely consider your needs.

ffriedl89:
We were debating whether utility classes ala tailwind make sense for us. 
For spacings this would mean that we would need to generate a lot of css classes. Without proper tooling (purgeCSS for unused classes) the amount of css code shipped can get quite out of hand. 
e.g. for spacing-xl we would need to create 
```
// padding
.dt-p-xl
// padding top
.dt-p-t-xl
.dt-p-r-xl
.dt-p-b-xl
.dt-p-l-xl

// margin
.dt-m-xl
.dt-m-t-xl
.dt-m-r-xl
.dt-m-b-xl
.dt-m-l-xl
```

So roughly 10 css classes for one spacing token. If unused and not removed from the bundle this gets really big. 
Another approach would be to generate mixins that can be used by the consumer. These mixins don't generate code unless used which would remove the bloating problem.

ffriedl89:
[...]

PR #1111, opened 2020-06-04
chore: Added spacing css outlet for design tokens

tomheller:
Added an css outlet for the spacing tokens, as described in #1057 

#### Type of PR

Other

#### Checklist

### DIFF
--- a/libs/shared/design-tokens/src/lib/global/typography.yml
+++ b/libs/shared/design-tokens/src/lib/global/typography.yml
@@ -14,8 +14,8 @@ props:
     meta:
       friendlyName: Headline 1
       pangram: 'The quick brown fox jumps over a lazy dog'
-      selector: .heading-1
-      mixinName: heading-1
+      selector: .dt-h1
+      mixinName: dt-h1
   - name: heading-2
     value:
       fontSize: '{!typography--font-size-3xl}'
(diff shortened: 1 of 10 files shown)


libs/shared/design-tokens/src/lib/global/typography.yml
```diff
@@ -14,8 +14,8 @@ props:
     meta:
       friendlyName: Headline 1
       pangram: 'The quick brown fox jumps over a lazy dog'
-      selector: .heading-1
-      mixinName: heading-1
+      selector: .dt-heading-1
```
ffriedl89, line 17:
```suggestion
      selector: .dt-h1
```
What do you think about shortening the selectors a bit - here dt-h1 would work I guess. 
What about `dt-body-up-sem` for `dt-body-uppercased-super-emphasized`?

lukasholzer: We stated in our Coding guidelines that we don't want to follow abbreviations. So I would avoid it here as well.

ffriedl89: To be honest in code i totally agree, but those classes need to be learned by our consumers and written by hand repeatedly. I think convenience wins out over not having abbreviations here.

Issue #1043, opened 2020-05-19
CI: Add a test matrix to run jobs on different OS


lukasholzer:
# Feature Request

As we run in several platform issues with windows we should run tests on the ci (at least on the master branch) in different platforms like windows and ubuntu.

**This should not be a test for an output bundle!** 
It should be a test that everything is working as expected for our contributors. Testing a contributors workflow should be key.

## Feature Description

In the circleci config add another workflow that is running test, build, and lint on windows.

This can be achieved using the cirecleci orb for windows: https://circleci.com/docs/2.0/configuration-reference/#example-usage-5

PR #1118, opened 2020-06-05
build: Add windows install, build and test checks

tomheller:

As our contributions grow, we should at least make sure that our repo is installable, buildable and runnable on windows machines. 

Relates to #1043 

#### Type of PR

Other: build 

#### Checklist

### DIFF
--- a/libs/workspace/src/builders/barista-build/utils/start-server.spec.ts
+++ b/libs/workspace/src/builders/barista-build/utils/start-server.spec.ts
@@ -68,25 +68,30 @@ it('should capture an internal server error on route failure', async () => {
   expect.assertions(3);
 });
 
-it('should capture a child process crash', async () => {
-  try {
-    forkPid = (
-      await startServer(fakeServerPath, 3333, {
-        FAKE_ERROR: 'true',
-      })
-    ).pid;
-  } catch (error) {
-    forkPid = error.pid;
-    expect(error.error).toMatch(/Error: FAKED ERROR/);
-  } finally {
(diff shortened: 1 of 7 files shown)


libs/workspace/src/builders/barista-build/utils/start-server.spec.ts
```diff
@@ -68,25 +68,30 @@ it('should capture an internal server error on route failure', async () => {
...
-  } catch (error) {
-    forkPid = error.pid;
-    expect(error.error).toMatch(/Error: FAKED ERROR/);
-  } finally {
-    expect(typeof forkPid).toBe('number');
-    process.kill(forkPid);
-  }
+// Skip this test on windows, as it is flaky
+// TODO: Figure out on how to test this properly in the future.
+(process.platform === 'win32' ? it.skip : it)(
```
tomheller, line 73: It's a little sad that I needed to do this, but I was not able to figure out why this test was flaky on the CI. 
The error captured from the process was always a little different, but contained a path rather than the expected error string. 
The `finally` crashed afterwards, because the process was no longer available to be killed...

lukasholzer: It is ok – as long it is running on any os. The builder will always run on Linux for creating barista.
The windows checks are only for our contributors so that if they try to build or run things don't have broken environments

tomheller:
@lukasholzer thanks you. It took me enough tries though, with the one or other complete breakage of the pipeline :wink:

PR #1129, opened 2020-06-08
chore: Extend accepted barista-icons range up to version 7

tomheller:

Extend the accepted range to version 7 as it is compatible with the icons usage within the library


ffriedl89:
This is a duplicate of #1124 - I think we can close either one of them, but i think this should be marked as feat in the commit

PR #1133, opened 2020-06-08
Added support for checkbox/switch as a form-field-control.

### LINKED ISSUE #1120: Checkbox & Switch should also work with a form-field
# Feature Request

## Summary

Currently the checkbox & switch components do not work with the form-field even as they are input components that represent a single value.

## Feature Description

Checkbox & Switch should implement the `DtFormFieldControl` to ensure they work with a `DtFormField`. 

The value of the checkbox/switch in the formfield should bei the passed in Generic value type if the it is checked or undefined if not.

thomaspink:
Added support for checkbox/switch as a form-field-control.

Fixes #1120

#### Type of PR

### DIFF
--- a/libs/barista-components/form-field/src/form-field-control.ts
+++ b/libs/barista-components/form-field/src/form-field-control.ts
@@ -57,6 +57,9 @@ export abstract class DtFormFieldControl<T> {
   /** Whether the control is in an error state. */
   readonly errorState: boolean;
 
+  /** @internal Whether the control is a box control and does use the border from the form-field. */
+  readonly _boxControl?: boolean;
+
   /** Sets the list of element IDs that currently describe this control. */
   abstract setDescribedByIds(ids: string[]): void;
 
--- a/libs/barista-components/form-field/src/form-field.scss
+++ b/libs/barista-components/form-field/src/form-field.scss
@@ -12,10 +12,16 @@
 
 .dt-form-field-body {
   @include dt-form-control-body();
+}
 
-  .dt-focused & {
-    @include dt-focus-style();
-  }
+.dt-form-field-non-box .dt-form-field-body {
+  border: none;
+  border-radius: 0;
+  min-height: 0;
+}
+
+.dt-focused:not(.dt-form-field-non-box) .dt-form-field-body {
+  @include dt-focus-style();
 }
 
 .dt-form-field-prefix,
--- a/apps/dev/src/checkbox/checkbox-demo.component.html
+++ b/apps/dev/src/checkbox/checkbox-demo.component.html
@@ -1,3 +1,8 @@
 <dt-checkbox checked>Check me</dt-checkbox>
 <dt-checkbox [indeterminate]="true">Indeterminate</dt-checkbox>
 <dt-checkbox disabled checked>Disabled</dt-checkbox>
+
+<dt-form-field>
+  <dt-checkbox checked>Check me</dt-checkbox>
+  <dt-hint>A cool hint.</dt-hint>
+</dt-form-field>
--- a/apps/dev/src/switch/switch-demo.component.html
+++ b/apps/dev/src/switch/switch-demo.component.html
@@ -9,3 +9,8 @@
   <dt-switch disabled>Disabled</dt-switch>
   <dt-switch checked disabled>Checked Disabled</dt-switch>
 </div>
+
+<dt-form-field>
+  <dt-switch checked>Checked</dt-switch>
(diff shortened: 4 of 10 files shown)


libs/barista-components/form-field/src/form-field.scss
```diff
@@ -13,9 +13,15 @@
...
-  .dt-focused & {
+  .dt-focused:not(.dt-form-field-non-box) & {
     @include dt-focus-style();
   }
+
+  .dt-form-field-non-box & {
+    border: none;
+    border-radius: 0;
+    min-height: 0;
+  }
```
ffriedl89, line 24: can we move this outside the nesting and write the combined selector here? 
I nearly missed the & and the end - its just hard to read that way

libs/barista-components/form-field/src/form-field-control.ts
```diff
@@ -57,6 +57,9 @@ export abstract class DtFormFieldControl<T> {
   /** Whether the control is in an error state. */
   readonly errorState: boolean;
 
+  /** Whether the control is not a box control and does not use the border from the form-field. */
+  readonly nonBoxControl?: boolean;
```
ffriedl89, line 61: can we make this internal? 
I'd also rename this a negated boolean is always a bit problematic in terms of readability. 
I'd rather have _isBoxControl and the opposite default

thomaspink: The idea with the negated name was that the value `false` or `undefined` (it's optional, to achieve backwards compatibility) lead to the same result

libs/barista-components/form-field/src/form-field.scss
thomaspink: 👍 changed

libs/barista-components/form-field/src/form-field-control.ts
ffriedl89: I'd prefer the non negated alternative ( see another double negative :D Its already stressing me out ). 
What about making it internal?

libs/barista-components/form-field/src/form-field.scss
```diff
@@ -13,11 +13,17 @@
 .dt-form-field-body {
   @include dt-form-control-body();
 
-  .dt-focused & {
+  .dt-focused:not(.dt-form-field-non-box) & {
```
ffriedl89, line 16: I just spotted another & at the end here ... sorry for not pointing that out in the initial review

libs/barista-components/form-field/src/form-field-control.ts
thomaspink: Cool for me :)

Whats the reasoning to make it internal?
The other props are also public. This field might also be interesting for consumer custom controls that want to support the filter field.

libs/barista-components/form-field/src/form-field.scss
thomaspink: This nesting was not introduced by me, i just added the `:not(.dt-form-field-non-box)`. 
But i can move it top level

libs/barista-components/form-field/src/form-field-control.ts
```diff
@@ -40,7 +40,7 @@ export abstract class DtFormFieldControl<T> {
   readonly id: string;
 
   /** Gets the NgControl for this control. */
-  readonly ngControl: NgControl | null;
+  readonly ngControl?: NgControl | null;
```
lukasholzer, line 43: This would be a breaking change or? If somebody is relying on this in his script he would get currently always a value now he can get undefined as well...

thomaspink: this is the least breaking i could make it. otherwise i would break all new form-field compatible components like switch/checkbox/...

thomaspink:
Can not think of useful tests for the `ngControl`. That's why we never had any. But if you have an idea, let me know

libs/barista-components/form-field/src/form-field-control.ts
ffriedl89: I highly doubt that this is important for consumers and having it internally keeps the api surface smaller, i'd suggest make it internal now and once we get requests for accessing this feature, open it up.

thomaspink:
@tomheller Great idea. As i also have to make the radio compatible i would suggest to do this in a separate PR once the radio PR also has been merged

tomheller:
> i would suggest to do this in a separate PR once the radio PR also has been merged

as long as you don't forget to do it 😉

PR #1143, opened 2020-06-09
Add Jest Unit tests with bazel

furtnerl:
Added Unit tests with bazel for the following components:

- core
- icon
- button

So when adding a unit test with bazel this is a drop-in replacement for the angular CLI

<hr/>

The change of the imports for `colors` is not related to bazel we got a transitive dependency called `colors` and scss now cannot resolve the import path from the `inlcudePaths`. To solve this issue provide a relative path.

image

To be honest I think the style `includePaths` are not the best choice anymore as it gets confusing from where the file is imported as there are many possible locations. So changing to a relative path makes it more verbose and readable IMO.


lukasholzer:
The change of the imports for `colors` is not related to bazel we got a transitive dependency called `colors` and scss now cannot resolve the import path from the `inlcudePaths`. To solve this issue provide a relative path.

image

To be honest I think the style `includePaths` are not the best choice anymore as it gets confusing from where the file is imported as there are many possible locations. So changing to a relative path makes it more verbose and readable IMO.

postinstall.js
```diff
@@ -1,15 +1,97 @@
...
+ * the processed Ivy bundles.
+ */
+function updateNgccMainFields() {
+  sync('node_modules/@angular/**/package.json').forEach((filePath) => {
+    // Do not update `package.json` files for deeply nested node modules (e.g. dependencies of
+    // the `@angular/compiler-cli` package).
+    if (filePath.lastIndexOf('node_modules/') !== 0) {
+      return;
+    }
+    const parsedJson = JSON.parse(readFileSync(filePath, 'utf8'));
```
ffriedl89, line 78: A lot could go wrong here, invalid json the fields not being there etc. Does it make sense wrap this in a try catch and handle a bit more gracefully?

tools/bazel_rules/jest/jest-resolver.js
```diff
@@ -0,0 +1,108 @@
...
+ * @param {string} options.basedir
+ * @param {defaultResolver} options.defaultResolver
+ * @param {string} options.extensions
+ * @param {string} options.moduleDirectory
+ * @param {string} options.paths
+ * @param {string} options.rootDir
+ */
+function moduleResolver(moduleId, options) {
+  // resolve workspace imports with the bazel module mappings
+  if (moduleId.includes('@dynatrace')) {
```
ffriedl89, line 84: NIT: Does a startsWith make more sense here?

tools/bazel_rules/jest/jest-resolver.js
```diff
@@ -0,0 +1,108 @@
...
+  } catch {
+    const resolved = sync(moduleId, {
+      basedir: options.basedir,
+      preserveSymlinks: false,
+    });
+    return resolved;
+    return sync(moduleId, {
+      basedir: options.basedir,
+      preserveSymlinks: false,
+    });
```
ffriedl89, line 104:
```suggestion
    return resolved;
```

ffriedl89, line 104:
```suggestion
    return sync(moduleId, {
      basedir: options.basedir,
      preserveSymlinks: false,
    });

```

tools/bazel_rules/jest/jest-runner.js
```diff
@@ -0,0 +1,85 @@
...
+        outputName: './test.xml',
+        classNameTemplate: '{classname}-{title}',
+        titleTemplate: '{classname}-{title}',
+        ancestorSeparator: ' › ',
+        usePathForSuiteName: 'true',
+      },
+    ],
+  ];
+
+  const updatedFileName = `${dir}/${name}-updated${ext}`;
```
ffriedl89, line 32: This might be an issue on windows - maybe use join instead

tools/bazel_rules/jest/jest-runner.js
```diff
@@ -0,0 +1,85 @@
...
+  const jestConfigPath = updateJestConfig(
+    resolve(jestConfig),
+    suite,
+    dirname(jestConfig),
+  );
+  const base = dirname(jestConfigPath);
+
+  process.env.BAZEL_TEST_MODULE_MAPPING = `${dirname(
+    jestConfig,
+  )}/_${suite}.module_mappings.json`;
```
ffriedl89, line 49: same here - maybe a path issue on windows

tools/bazel_rules/jest/jest-resolver.js
lukasholzer: will try.. had the problem that the moduleId can be a resolved path like `/bazel-out/bin/@dynatrace/barista-componets/...` but maybe this was solved through a different thing.

.circleci/config.yml
```diff
@@ -211,13 +235,13 @@ jobs:
           # Don't forget to adapt push command and bazel executor if you want to bump the Bazel version
           command: |
             docker build -t $WORKSPACE_IMAGE:latest ./ -f ./.deployment/workspace-base.Dockerfile
-            docker build -t $BAZEL_IMAGE:bazel-3.2.0 ./ -f ./.deployment/workspace-bazel.Dockerfile --build-arg bazel_version=3.2.0
+            docker build -t $BAZEL_IMAGE:bazel-3.3.1 ./ -f ./.deployment/workspace-bazel.Dockerfile --build-arg bazel_version=3.3.1
```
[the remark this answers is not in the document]
lukasholzer: we will enhance this on a later point.

WORKSPACE
```diff
@@ -1,3 +1,13 @@
+# The nodejs rules
+RULES_NODEJS_VERSION = "2.0.0-rc.1"
+RULES_NODEJS_SHA256 = "aa0b0a71afd4e1f203b7092c3284a6606a5bfac77e0bd31f071b37bcac0f7cf3"
+# Rules for compiling sass
+RULES_SASS_VERSION = "1.26.3"
+RULES_SASS_SHA256 = "9dcfba04e4af896626f4760d866f895ea4291bc30bf7287887cefcf4707b6a62"
+# Bazel toolchain needed for remote execution
+BAZEL_TOOLCHAIN_VERSION = "3.2.0"
```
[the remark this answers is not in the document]
lukasholzer: the toolchain is a different repository

postinstall.js
lukasholzer: This is only a parse for the angulars package.json if this one is broken all angular apps would be broken so I cannot see any benefit of adding a try catch here.

lukasholzer: For reference where I copied the ngcc madness that have to be done: https://github.com/angular/components/blob/master/tools/postinstall/update-ngcc-main-fields.js#L31

PR #1147, opened 2020-06-09
fix(chart): Fixes showing inactive series opaque.

### LINKED ISSUE #1146: Chart: Hovering over datapoint where series is undefined, adds an inactive state to the series for the whole chart
# Bug Report

Highcharts 7 introduced a new default to the inactive state of a series, which you can see when you hover over a timestamp where there is no value for a series. What happens is, that the series gets an opacity of 0.2 for the rest of the chart, which is very confusion especially for stacked area charts.

## Possible Solution

As @ffriedl89 suggested opacity in the inactive value of the series plotoptions should be set to 1.

I will create a pr shortly!

bug

jsjsat:
Closes #1146

<hr>

  often helpful).
  guidelines.

#### Checklist


libs/barista-components/chart/src/chart-options.ts
```diff
@@ -56,6 +56,11 @@ export const DT_CHART_DEFAULT_OPTIONS: DtChartOptions = {
   },
   plotOptions: {
     series: {
+      states: {
+        inactive: {
+          opacity: 1,
```
ffriedl89, line 61: NIT: It would be really awesome to add a comment here what this option does, otherwise I'll probably forget until early next week. 
```suggestion
          // This handles enabling / disabling the inactive state
          // Inactive state happens for example when a series has data missing (null as value for datapoints)
          // by default the series in area charts gets hidden up from v7. Setting this to 1 disables this behavior 
          opacity: 1,
```

ffriedl89:
Thx @jsjsat for contributing this. If you could add the comment that would be amazing! Besides that all 🚀

libs/barista-components/chart/src/chart-options.ts
jsjsat: Done!

PR #1151, opened 2020-06-12
chore(design-tokens): Build additional metadata for consumption by the Design Tokens UI.

SkaillZ:

Changes the design tokens builder to build JSON files for alias files instead of hardcoding `palette-source.alias.yml`

#### Checklist

### DIFF
--- a/libs/workspace/src/builders/design-tokens/build/schema.json
+++ b/libs/workspace/src/builders/design-tokens/build/schema.json
@@ -15,12 +15,25 @@
         "type": "string"
       }
     },
+    "aliasesEntrypoints": {
+      "type": "array",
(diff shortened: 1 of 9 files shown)


libs/workspace/src/builders/design-tokens/build/schema.json
```diff
@@ -15,12 +15,25 @@
         "type": "string"
       }
     },
+    "aliasesEntrypoints": {
+      "type": "array",
+      "description": "List of alias files to build JSON metadata for as required by some applications.",
```
lukasholzer, line 20: NIT: I would add that it is a list of glob patterns

PR #1153, opened 2020-06-15
radio: Added support for form-field

### LINKED ISSUE #1121: Radio should work with a form-field
# Feature Request

## Summary

Currently the radio components do not work with the form-field even as they are input components that represent a single value.

## Feature Description

As multiple radios or a single radio-group wrapping multiple radios represent one value, we limit the support for the form-field to the radio-group.

The radio-group therefore should implement the `DtFormFieldControl` interface.

The value of the radio-group in the formfield should bei either the passed in Generic value type of the radios if one is selected or undefined if no radio is selected at the beginning.

thomaspink:
Blocked until #1133 is merged

Fixes #1121

### DIFF
--- a/libs/barista-components/radio/src/radio-group.ts
+++ b/libs/barista-components/radio/src/radio-group.ts
@@ -25,35 +25,39 @@ import {
   Output,
   QueryList,
   forwardRef,
+  Inject,
+  Optional,
+  Self,
 } from '@angular/core';
-import { NG_VALUE_ACCESSOR } from '@angular/forms';
+import { ControlValueAccessor, NgControl } from '@angular/forms';
 
 import { CanDisable } from '@dynatrace/barista-components/core';
+import { DtFormFieldControl } from '@dynatrace/barista-components/form-field';
 
 import { DtRadioButton, DtRadioChange } from './radio';
+import { Subject } from 'rxjs';
 
 let nextUniqueId = 0;
 
 @Directive({
   selector: 'dt-radio-group',
   exportAs: 'dtRadioGroup',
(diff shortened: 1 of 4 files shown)


libs/barista-components/radio/src/radio-group.ts
```diff
@@ -113,11 +119,40 @@ export class DtRadioGroup<T> implements AfterContentInit, CanDisable {
...
+  /** Unique id of the element. */
+  @Input()
+  get id(): string {
+    return this._id;
+  }
+  set id(value: string) {
+    this._id = value || this._uid;
+  }
+  private _uid = `dt-checkbox-${nextUniqueId++}`;
+  private _id: string;
```
tomheller, line 131: If the setter is never called, the id will never be set. We need to add a default here.

tomheller:
Additionally please follow this one up with the extension of the form-field readme and some examples.

libs/barista-components/radio/src/radio-group.ts
lukasholzer: shouldn`t be the default value for the _id? Because if you never call the setter the id will be undefined.

```diff
 @Input()
  get id(): string {
    return this._id;
  }
  set id(value: string) {
-    this._id = value || this._uid;
+    this._id = value;
  }
  private _uid = `dt-checkbox-${nextUniqueId++}`;
- private _id: string;
+ private _id: string = this._uid;
```

thomaspink: Setter is called in the constructor. It now works the same as in the checkbox.

PR #1176, opened 2020-06-22
Chore/fluid checkbox

tomheller:
Web component representation of a checkbox example

#### Type of PR
Other

#### Checklist

### DIFF
--- a/libs/fluid-elements/checkbox/src/lib/checkbox.scss
+++ b/libs/fluid-elements/checkbox/src/lib/checkbox.scss
@@ -1,0 +1,108 @@
+@import '~@dynatrace/fluid-design-tokens/global/spacing';
+@import '~@dynatrace/fluid-design-tokens/global/typography';
+
+:host {
+  /**
+   * Legibility definitions should probably be
+   * shipped or imported from a core
+  */
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+  text-rendering: optimizeLegibility;
+}
+
+/**
+* Disabled state
+*/
+:host([disabled]) {
+  pointer-events: none;
+
+  .fluid-label {
+    color: var(--color-neutral-150);
+    opacity: 0.5;
+  }
+
+  .fluid-svg-checkbox {
+    opacity: 0.5;
+  }
+}
+
+.fluid-checkbox {
+  margin-left: -$fluid-spacing--2x-small;
+  display: flex;
+  position: relative;
+}
+
+.fluid-label {
+  @include fluid-dt-text();
+  color: var(--color-neutral-150);
+  margin-left: $fluid-spacing--x-small;
+}
+
+label {
+  cursor: pointer;
+  display: inline-flex;
+  align-items: center;
+
+  &:hover .fluid-svg-checkbox-background {
+    opacity: 0.1;
+    transform: scale(1);
+  }
+}
+
+.fluid-checkbox-input {
+  position: absolute;
+  width: 24px;
+  height: 24px;
+  top: -$fluid-spacing--2x-small;
+  left: -$fluid-spacing--2x-small;
+  opacity: 0;
+}
+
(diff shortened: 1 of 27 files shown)


libs/fluid-elements/checkbox/src/lib/checkbox.scss
```diff
@@ -0,0 +1,102 @@
+@import '~@dynatrace/fluid-design-tokens/global/spacing';
+@import '~@dynatrace/fluid-design-tokens/global/typography';
+
+:host {
```
ffriedl89, line 4: I think we should do something similar to the button here and declare intermediate css custom properties. 
https://github.com/dynatrace-oss/barista/blob/master/libs/fluid-elements/button/src/lib/button.scss

ffriedl89: But i am fine with doing that in a followup issue if you like. Please create an issue in that case :)

libs/fluid-elements/checkbox/src/lib/checkbox.scss
```diff
@@ -0,0 +1,102 @@
...
+  align-items: center;
+
+  &:hover .fluid-svg-checkbox-background {
+    opacity: 0.1;
+    transform: scale(1);
+  }
+}
+
+.fluid-checkbox-input {
+  display: none;
```
ffriedl89, line 53: This might not be optimal for screen readers and a11y - This article is quite an interesting read. https://www.sarasoueidan.com/blog/inclusively-hiding-and-styling-checkboxes-and-radio-buttons/

libs/fluid-elements/checkbox/src/lib/checkbox.ts
```diff
@@ -0,0 +1,293 @@
...
+        </label>
+      </div>
+    `;
+  }
+
+  /**
+   * Dispatches a change event for the checkbox
+   */
+  private _dispatchChangeEvent(): void {
+    requestAnimationFrame(() => {
```
ffriedl89, line 244: I am interested why you request an animationFrame here? :)

libs/fluid-elements/checkbox/src/lib/checkbox.ts
```diff
@@ -0,0 +1,293 @@
...
+    this.checked = (event.target as HTMLInputElement).checked;
+    if (this.indeterminate) {
+      this.indeterminate = false;
+      this._dispatchIndeterminateChangeEvent();
+    }
+    this._dispatchChangeEvent();
+  }
+
+  /**
+   * Handles the kwyup event on the checkbox to proxy it to the nativ input
```
ffriedl89, line 274:
```suggestion
   * Handles the keyup event on the checkbox to proxy it to the native input
```

tomheller:
@ffriedl89 pushed again to fix the now renamed mixin from the design tokens, changed from `fluid-body` to `fluid-dt-text`

libs/fluid-elements/checkbox/src/lib/checkbox.scss
tomheller: Issue created at #1191

libs/fluid-elements/checkbox/src/lib/checkbox.scss
tomheller: Very good point, resolved in the next commit.

libs/fluid-elements/checkbox/src/lib/checkbox.ts
tomheller: Eventually, when grouped elements need to react to a downstream change it seemed like a good idea to defer this to the next cycle.

Weightless does this, but i checked with lion and they are not. 
https://github.com/andreasbm/weightless/blob/6c7965981752e85206d9da459110b2dddeecf9ec/src/lib/behavior/switch/switch-behavior.ts#L119-L124

I'm not sure now and we can remove it if you want.

Issue #1173, opened 2020-06-18
Sunburst-chart: various improvements


subarroca:
# Feature Request

## Summary

Currently we have v1 of sunburst but after taking it to a real project there are some improvements to be done

## Feature Description

- Responsiveness. Currently the text scales and it's inconsistent with the rest of charts
- Better managing of overlay. Currently we have flickers in some cases
- Central label. Currently it can fall behind the slices and become unreadable

No API changes should take place

tomheller:
Hey @subarroca! Thank you for sticking with the component and keep improving it. Very much appreciated 🙌 🥳

PR #1189, opened 2020-06-22
feat(sunburst-chart): Add ariaLabel and responsiveness

subarroca:
Feature

- Added aria labels
- Changed overlay to avoid flickering
- Add responsiveness to get more consistent label sizes
- Move central label to front and enable multiline

#### Checklist


subarroca:
closes #1173

apps/components-e2e/src/components/sunburst-chart/sunburst-chart.po.ts
```diff
@@ -18,18 +18,18 @@ import { Selector } from 'testcafe';
...
 export const centralValue = Selector('.dt-sunburst-chart-selected-value');
 
 export const absoluteBtn = Selector('#chart-value-mode-absolute');
 export const percentBtn = Selector('#chart-value-mode-percent');
 
-export const segments = Selector('.dt-sunburst-chart-group');
+export const segments = Selector('[dt-sunburst-chart-segment]');
 export const sliceShephard = Selector(
-  '.dt-sunburst-chart-group:nth-of-type(5) .dt-sunburst-chart-slice',
+  '[dt-sunburst-chart-segment]:nth-of-type(5) .dt-sunburst-chart-slice',
```
tomheller, line 31: Could you please change the `nth-of-type` selector to the `nth` helper function, like we did for the stacked-series-chart.

libs/barista-components/sunburst-chart/src/sunburst-chart.ts
```diff
@@ -135,27 +139,64 @@ export class DtSunburstChart {
...
 
+  /** AfterContentInit hook */
+  ngAfterContentInit(): void {
+    this._updateDimensions();
+    this._resizer
+      .change()
+      .pipe(
+        // delay to postpone the reflow to the next change detection cycle
+        takeUntil(this._destroy$),
+        delay(0),
```
tomheller, line 170: Nitpick: We could also think about using an AnimationFrameScheduler here to not have to hack the `delay(0)`.

libs/barista-components/sunburst-chart/src/sunburst-chart.ts
```diff
@@ -135,27 +139,64 @@ export class DtSunburstChart {
...
+      )
+      .subscribe(() => {
+        this._updateDimensions();
+        this._cdr.markForCheck();
```
tomheller, line 174: Our coding guidelines state to not abbreviate variable names if at all possible.
```suggestion
        this._changeDetectorRef.markForCheck();
```

libs/barista-components/sunburst-chart/src/sunburst-chart.ts
```diff
@@ -194,110 +235,47 @@ export class DtSunburstChart {
...
+   * Updates the position of the overlay to create a mouseFollow position
+   */
+  _handleOnMouseMove(event: MouseEvent): void {
     if (this._overlayRef) {
-      this._overlayRef.detach();
-      this._portal = null;
+      this._overlayService._positionStrategy.setOrigin({
+        x: event.clientX,
+        y: event.clientY,
+      });
```
tomheller, line 266: Please check this in firefox if you haven't already. We had the problems in the past with clientX or event offsets in general not being the same on svg elements in firefox as in blink-based browsers.

libs/barista-components/sunburst-chart/src/sunburst-chart.ts
```diff
@@ -194,110 +235,47 @@ export class DtSunburstChart {
...
-
-    this._dismissOverlay();
-    if (origin) {
-      //  TODO: lukas.holzer
-      // Bazel cannot resolve the extends of the interface properly and
-      // throws error: Property 'data' does not exist on type 'DtSunburstChartSlice'.
-      // @ts-ignore
-      this._createOverlay(origin, node.data);
-    }
-  }
```
tomheller, line 214: If you deprecate the code, you cannot actually delete it. It still needs to work. 
Why are you removing this functionality exactly? Nothing was stated in the issue that you are planning to remove this.

libs/barista-components/sunburst-chart/src/sunburst-chart.ts
```diff
@@ -194,110 +235,47 @@ export class DtSunburstChart {
       this._selectedRelativeValue = 1;
     }
   }
-  /** Programmatically close the overlay. */
-  closeOverlay(): void {
-    this._dismissOverlay();
-  }
```
tomheller, line 200: If you deprecate the code, you cannot actually delete it. It still needs to work.
Why are you removing this functionality exactly? Nothing was stated in the issue that you are planning to remove this.

libs/barista-components/sunburst-chart/src/sunburst-chart.ts
```diff
@@ -135,27 +139,64 @@ export class DtSunburstChart {
...
     private _platform: Platform,
     // TODO: remove this sanitizer when ivy is no longer opt out
     private _sanitizer: DomSanitizer,
-    private _elementRef?: ElementRef<HTMLElement>,
-    @Optional()
-    @Inject(DT_UI_TEST_CONFIG)
-    private _config?: DtUiTestConfiguration,
+    private _elementRef: ElementRef<HTMLElement>,
+    private _resizer: DtViewportResizer,
+    private _cdr: ChangeDetectorRef,
```
tomheller, line 159: Could we do this without a breaking change as well? When making the new parameters optional and have a fallback to the way that it is currently working, we should be able to do this without a breaking change. 
As I assume that after we released version 7 a couple of weeks ago, version 8 will take some time to be released.

libs/barista-components/sunburst-chart/src/sunburst-chart.ts
```diff
@@ -49,13 +49,10 @@ import {
...
-import { Platform } from '@angular/cdk/platform';
-import { DomSanitizer, SafeStyle } from '@angular/platform-browser';
-import { DtColors } from '@dynatrace/barista-components/theming';
 
-const OVERLAY_PANEL_CLASS = 'dt-sunburst-chart-overlay-panel';
-const VIEWBOX_WIDTH = 480;
-const VIEWBOX_HEIGHT = 400;
+/** Minimum width of the chart */
+const MIN_WIDTH = 480;
+const MAX_WIDTH = 800;
```
ffriedl89, line 55: Why do we need a max width here? The surrounding layout should limit the chart from growing.

libs/barista-components/sunburst-chart/src/sunburst-chart.ts
ffriedl89: Yeah please leave even unused parameters in there to avoid breaking changes. Version 8 is far out. No plans or timeline yet.

libs/barista-components/sunburst-chart/README.md
```diff
@@ -47,10 +48,10 @@ attribute is required to create a valid output.
 
 #### Methods
 
-| Name           | Params                              | Description                         |
-| -------------- | ----------------------------------- | ----------------------------------- |
-| `openOverlay`  | `node: DtSunburstChartTooltipNode>` | Open an overlay for the given node. |
-| `closeOverlay` |                                     | Closes the overlay if open.         |
+| Name           | Params                              | Description                                     |
+| -------------- | ----------------------------------- | ----------------------------------------------- |
+| `openOverlay`  | `node: DtSunburstChartTooltipNode>` | Deprecated. Open an overlay for the given node. |
```
ffriedl89, line 53: Why deprecate this method? It seems rather useful to have a possibility to open the overlay programmatically.

libs/barista-components/sunburst-chart/src/sunburst-chart.ts
subarroca: works as a charm :D

libs/barista-components/sunburst-chart/src/sunburst-chart.ts
subarroca: @tomheller @ffriedl89 I can keep old params but for new params we're gonna loose the ability of scaling the chart on resize. I think it would be good to open a new PR for this so I'll leave the code commented

libs/barista-components/sunburst-chart/src/sunburst-chart-segment.ts
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Component, Input } from '@angular/core';
+import { DtSunburstChartNodeSlice } from './sunburst-chart.util';
+
+@Component({
+  selector: 'svg:g[dt-sunburst-chart-segment]',
+  templateUrl: 'sunburst-chart-segment.html',
```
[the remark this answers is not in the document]
subarroca: Tests fail when using svg file instead of html

tomheller:
@subarroca 
Seems that there is still a build error, one of the functions called from the template has a parameter mismatch

```
------------------------------------------------------------------------------
 
Building entry point '@dynatrace/barista-components/sunburst-chart'
 
------------------------------------------------------------------------------
 
Compiling TypeScript sources through ngc
 
ERROR: libs/barista-components/sunburst-chart/src/sunburst-chart.html(14,5): Expected 0 arguments, but got 1.
 ```

File: libs/barista-components/sunburst-chart/src/sunburst-chart.ts [hunk 11]
```diff
@@ -135,27 +112,133 @@ export class DtSunburstChart {
...
-  private _overlayRef: OverlayRef;
+  private _overlayRef: DtOverlayRef<DtSunburstChartTooltipData> | null;
+
+  /** @internal Destroy subject to clear subscriptions on component destroy. */
+  private readonly _destroy$ = new Subject<void>();
 
   constructor(
-    private _overlayService: Overlay,
-    private _viewContainerRef: ViewContainerRef,
+    private _overlayService: DtOverlay,
```
tomheller, line 168: I guess this would still be a breaking change, when changing the type from `Overlay` to `DtOverlay`.

libs/barista-components/sunburst-chart/src/sunburst-chart.ts
```diff
@@ -135,27 +112,133 @@ export class DtSunburstChart {
...
+    _viewContainerRef: ViewContainerRef,
     private _platform: Platform,
     // TODO: remove this sanitizer when ivy is no longer opt out
     private _sanitizer: DomSanitizer,
-    private _elementRef?: ElementRef<HTMLElement>,
+    private _elementRef: ElementRef<HTMLElement>,
+    // private _resizer: DtViewportResizer,
+    // private _changeDetectorRef: ChangeDetectorRef,
```
tomheller, line 175: We can make the parameters provided by the DI optional. This way, if the component is used in the template, everything will work as expected. The breaking change in the constructor usually only affects consumers, that would create components programatically using a componentFactory. 

```suggestion
    /** @breaking-change Make the DtViewportResizer mandatory in version 8.0.0 */
    @Optional() private _resizer?: DtViewportResizer,
    /** @breaking-change Make the ChangeDetectorRef mandatory in version 8.0.0 */
    @Optional() private _changeDetectorRef?: ChangeDetectorRef,
```

libs/barista-components/sunburst-chart/src/sunburst-chart.ts
```diff
@@ -135,27 +112,133 @@ export class DtSunburstChart {
...
     @Optional()
     @Inject(DT_UI_TEST_CONFIG)
-    private _config?: DtUiTestConfiguration,
+    _config?: DtUiTestConfiguration,
   ) {}
 
+  /** AfterContentInit hook */
+  // TODO: Uncomment this code to allow resizing.
+  // It's commented now to avoid a breaking change
```
tomheller, line 183:
```suggestion
```

libs/barista-components/sunburst-chart/src/sunburst-chart.ts
```diff
@@ -135,27 +112,133 @@ export class DtSunburstChart {
...
+    // this._resizer
+    //   .change()
+    //   .pipe(
+    //     takeUntil(this._destroy$),
+    //   )
+    //   .subscribe(() =>
+    //     animationFrameScheduler.schedule(() => {
+    //       this._updateDimensions();
+    //       this._changeDetectorRef.markForCheck();
+    //     }, 0, 0));
```
tomheller, line 196:
```suggestion
// @breaking-change Remove the if check when the resizer is no longer optional
if(this._resizer) {
     this._resizer
       .change()
       .pipe(
         takeUntil(this._destroy$),
       )
       .subscribe(() =>
         animationFrameScheduler.schedule(() => {
           this._updateDimensions();
           this._changeDetectorRef.markForCheck();
         }, 0, 0));
}
```

tomheller:
@subarroca We don't need to uncomment the breaking change, but we can make the changes in the constructor optional, and within the component deal with the fact that values provided by the DI might not be there (i.e. the component should not break if `this._resizer` is undefined). 

This way we can roll out the functionality to most consumers without breaking the components api and progressively enhance the features.

ffriedl89:
cherry pick done

PR #1193, opened 2020-06-23
Fix changedetection issue simple order column

### LINKED ISSUE #1166: Table: Expression changed after checked error when reordering rows.
# Bug Report

## Expected Behavior

Reordering should not throw an error.

## Current Behavior

When reordering rows, an ExpressionChangedAfterItHasBeenCheckedError is thrown, which has not been reported before. So far, I've not been able to find the issue, already talked to @ffriedl89, who will help me figure this one out.

## Possible Solution

Must have something to do with where/when the new index of the rows is being set.

## Steps to Reproduce

Reorder table rows in dev-app/demos with angular v9.1.9 (error is not reported with v9.1.6)

1. 

### Used Versions:

- **node**: 
- **npm**: 
- **typescript**: ^3.8.3
- **angular**: ^9.1.9
- **@angular/cdk**: ^9.2.4
- **@dynatrace/barista-components**:

ffriedl89:

Fixes #1166 

We went away from simple columns since they cause more problems than they solve. We even had a CD issue after upgrading a patch Angular version. In this case after upgrading from 9.1.6 to 9.1.9. This works the same way and there are no real downsides to this changed approach.

### DIFF
--- a/apps/dev/src/table-order/table-order-demo.component.html
+++ b/apps/dev/src/table-order/table-order-demo.component.html
@@ -8,11 +8,11 @@
   cdkDropList
   [cdkDropListData]="dataSource"
 >
-  <dt-simple-order-column
-    name="order"
-    label="Order"
-    dtColumnProportion="0.2"
-  ></dt-simple-order-column>
+  <ng-container [dtColumnDef]="'order'" dtColumnAlign="text">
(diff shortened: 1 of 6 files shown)


ffriedl89:
Rebase done :)

PR #1199, opened 2020-06-24
chore: Added tailwind output to design tokens generation.

### LINKED ISSUE #1182: Generate tailwind styles based on our design tokens
# Feature Request

Generate tailwind stylesheets based on our design tokens. 
Tailwind lets you customize the values used for generating the classes. We can already leverage our design tokens for colors, spacings etc. 

PoC done and looking good. We should now integrate this into the repository and create a script to generate the styles.

tomheller:

Added an additional output to the design-tokens output. It sources what we currently have in the
alias definitions and patches it into a tailwindcss configuration, which is then passed to the
tailwind generator.

It ouputs a css file that will be shipped with the design-tokens package and can be used standalone
without tailwindcss or postcss.

Fixes #1182

#### Type of PR

Other

#### Checklist

### DIFF
--- a/libs/workspace/src/builders/design-tokens/tailwindcss/utils/generate-color-group-config.ts
+++ b/libs/workspace/src/builders/design-tokens/tailwindcss/utils/generate-color-group-config.ts
@@ -1,0 +1,80 @@
+/**
+ * @license
+ * Copyright 2020 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { DesignTokenSource } from '../../interfaces/design-token-source';
+
+/** Removes the color and theme prefix from the color name. */
+function replaceColorPrefixes(colorname: string): string {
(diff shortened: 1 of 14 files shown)


libs/workspace/src/builders/design-tokens/tailwindcss/utils/generate-color-group-config.ts
```diff
@@ -0,0 +1,70 @@
...
+/** Generate the base colors assignments for `base` and `text` colors */
+export function generateBaseColors(
+  basecolorAliasSource: DesignTokenSource,
+): {
+  [colorShade: string]: string;
+} {
+  // Get all colors and replace their themenames
+  const colors = Object.keys(basecolorAliasSource.aliases!).map((colorname) =>
+    colorname.replace('color-abyss', '').replace('color-surface', ''),
+  );
```
ffriedl89, line 62: NIT: This could be moved to a function so you DRY :D

libs/workspace/src/builders/design-tokens/tailwindcss/utils/generate-font-config.ts
```diff
@@ -0,0 +1,94 @@
...
+/** Generates a config object for tailwinds font weight configuration. */
+export function generateFontWeight(
+  typographyAliasSource: DesignTokenSource,
+): { [key: string]: string } {
+  const fontWeightConfig = {};
+  for (const [typoAliasName, typoAliasValue] of Object.entries(
+    typographyAliasSource.aliases!,
+  )) {
+    if (typoAliasName.includes('emphasis')) {
+      // Get the correct fontweight naming from the alias name
```
ffriedl89, line 65: NIT: It would be cool to have an example in here whats coming in and what you expect as a result after transformation. 
E.g. `typography--emphasis-regular` from the aliases would become `regular`

ffriedl89: Otherwise i'd have to know while reading what the structure of those aliases is

PR #1201, opened 2020-06-25
Bugfix/event chart decimals

subarroca:
- Added a maxDecimals param in the DurationFormatter
- Use this param in event chart

#### Checklist

### DIFF
--- a/libs/barista-components/formatters/src/duration/duration-formatter-utils/transform-result-precise.ts
+++ b/libs/barista-components/formatters/src/duration/duration-formatter-utils/transform-result-precise.ts
@@ -28,6 +28,7 @@ import { dtConvertToMilliseconds } from './convert-to-milliseconds';
  * @param inputUnit dtTimeUnit value describing which unit the duration is in
  * @param outputUnit dtTimeUnit | undefined value describing the unit to which it should format
  * @param formatMethod the formatting method
+ * @param maxDecimals max amount of decimals
  */
 
 export function dtTransformResultPrecise(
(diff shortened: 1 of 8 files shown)


libs/barista-components/formatters/src/duration/duration-formatter-utils/transform-result-precise.ts
```diff
@@ -57,6 +60,9 @@ function calcResult(
     if (unit === DtTimeUnit.MICROSECOND || unit === DtTimeUnit.NANOSECOND) {
       amount *= MOVE_COMMA;
     }
+    if (maxDecimals) {
+      amount = Math.round(amount * 10 ** maxDecimals) / 10 ** maxDecimals;
+    }
```
tomheller, line 65: Would not run if `maxDecimals` is 0. please check for that as well.

libs/barista-components/formatters/src/duration/duration-formatter.ts
```diff
@@ -31,12 +31,14 @@ import {
...
  * @param outputUnit dtTimeUnit | undefined value describing the unit to which it should format e.g to seconds
  * @param inputUnit dtTimeUnit value describing which unit the duration is in (default: milliseconds)
+ * @param maxDecimals max amount of decimals
  */
 export function formatDuration(
   duration: number,
   formatMethod: DurationMode = 'DEFAULT',
   outputUnit?: DtTimeUnit,
   inputUnit: DtTimeUnit = DtTimeUnit.MILLISECOND,
+  maxDecimals?: number,
```
ffriedl89, line 41: I think we should add this parameter also to the pipe DtDuration. Since this function should be identical to the pipe. We need to add a default to the pipe though

ffriedl89: Could you also add this to the documentation README then.

subarroca: right now we don't have a default, that's why I left it unset, so we would not add any change to the current users. But if you want I can set it to 3 decimals, for instance.

ffriedl89: I totally agree that it should not change anything, so maybe having it as undefined is fine as a default and just not do anything in the case it is not provided. My point was mainly about having the option on the pipe to define it as well. I am sorry i was not clear about that :)

tomheller:
This looks good to me, great work @subarroca!

Dear cherry picker, please mind that this pr consists of 2 commits (1 _fix_ and 1 _feat_).
Please cherry pick the _fix_ into the minor and patch branches, and the _feat_ only into the minor branch.

tomheller:
Manually cherry picked.

Issue #601, opened 2020-02-17
Run a11y tests only for affected components


lukasholzer:
# Feature Request

## Summary

Currently, if anything affects something in the components we are running all a11y tests. This consumes a lot of ci resources because we have around 300 examples that we are testing.

To shorten our CI load we should only run affected examples.

## Feature Description

To accomplish this we should tag our a11y tests with the component name to run them by specific metatdata.

```
fixture(component.name)
  .meta('component', component.name);
```

Then we can run the a11y tests by the affected component name: The list of all affected libs should be retrieved by `nx affected:libs`.

tomheller:
@lukasholzer I see a little bit of an issue with that. 
Right now, we just iterate over all of the example component folders, but these do not necessarily correlate the the nx dependency graph. I.e. a change in the button should also run the table tests, as it is included there and should also serve it's a11y responsibilities there. 

Would that mean that we would have to write the fixtures for every component ourselves, like:  
```ts
fixture('table')
  .meta('component', 'table')
  .meta('component', 'button')
  .meta('component', 'empty-state');
```

lukasholzer:
To clarify it a little bit. When I change something in the empty state the command `nx affected:libs` will output something like `empty-state, table, tree-table, components`. The picture of the dependency graph will hopefully explain that to you.

![image](https://user-images.githubusercontent.com/11156362/74922613-c5cb6b80-53cf-11ea-94c9-84c1643a87a7.png)


This will mean when we run testcafe with `--fixture-component=empty-state=mobile,component=table,component=tree-table`.  The a11y tests of all affected components will run. But we don"t have to run the radio or checkbox a11y tests.

tomheller:
After an initial trial with defining this, we hit a little bit of a bump here. We were able to extract the affected components from a PR or changeset and pass them via a custom builder to the polite testcafe builder ( which runs a11y and e2e tests). 
Unfortunately the `fixtureMeta` filter outlined above is a match all metadata filter and not a a match any. 
So if the filter passed to the builder has multiple components affected, the fixture would only run if it defines more than the affected components. 
Examples: 
* Button component has a metadata of `button` and `icon` and the affected component would be `button`, the tests would run.
* Button component has a metadata of `button` and `icon` and the affected component would be `button`, `expandable` and `table`, the tests for the button would not run, because it does not contain `expandable` or `table`. 

At this point I have no idea how to mitigate this problem.

lukasholzer:
[...]

PR #1203, opened 2020-06-25
build(e2e): Create a custom builder that only passes affected sources into the e2e test environment.

tomheller:

Using the nx affected projects methods to figure out, which componenten modules are affected
and looking for these component modules within the imports of the e2e-testapp-modules
to determine which e2e tests need to be run.

Relates to #601 

#### Type of PR

Other

#### Checklist

### DIFF
--- a/libs/workspace/src/builders/affected-e2e/build.ts
+++ b/libs/workspace/src/builders/affected-e2e/build.ts
@@ -1,0 +1,91 @@
+/**
+ * @license
+ * Copyright 2020 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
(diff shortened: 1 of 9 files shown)


libs/workspace/src/builders/affected-e2e/build.ts
```diff
@@ -0,0 +1,86 @@
...
+import { findE2eModulesUsingAffectedComponents } from './utils/find-modules-using-affected-components';
+import { join } from 'path';
+import { transformModulePathToGlob } from './utils/transform-module-path-to-glob';
+
+/**
+ * Custom builder for the web-components package builder.
+ * This builder will schedule and run all projects that are tagged
+ * with a given tag in the nx.json. It is primarily used to run and package
+ * the web-components part of the library.
+ */
```
ffriedl89, line 38:
```suggestion
/**
 * Custom builder for run only affected e2e tests
 * This builder will run all e2e tests that use components that are 
 * affected by the changeset 
 */
```

libs/workspace/src/builders/affected-e2e/build.ts
```diff
@@ -0,0 +1,86 @@
...
+Running e2e tests matching the following globs
+${globSource.map((comp) => `- ${comp}`).join('\n')}
+`);
+      const e2eTarget = targetFromTargetString(options.e2eTarget);
+      return context.scheduleTarget(e2eTarget, { src: globSource });
+    }),
+    // Wait for the output and result Observable to resolve
+    switchMap((build) => forkJoin(build.output, build.result)),
+    // Switch over to the result of the e2e target
+    map((results) => results[1]),
```
lukasholzer, line 84: please add a catch statement where you map to `{success: false}` if something is failing on the affected computation. Currently, you only pass the result of the e2e test run.

PR #1205, opened 2020-06-25
Feat/next renaming shared

rowa-audil:
### Renaming to more general class names

This PR changes the class names of the page-guard and page-service and also changes the imports to these two and the universal.interseptor to `@dynatrace` imports.


nx.json
```diff
@@ -358,7 +358,7 @@
       "implicitDependencies": ["fluid-button"]
     },
     "shared-data-access-strapi": {
-      "tags": ["scope:cms", "type:shared"]
+      "tags": ["scope:cms", "type:shared", "type:library"]
```
lukasholzer, line 361: is the scope cms needed for a lint rule? cannot see the benefit of it currently.

lukasholzer:
I think those changes are implemented in the add next app PR.

ffriedl89:
@rowa-audil I think we can close this PR. Lets go over these cleanup renames together next week :)

PR #1207, opened 2020-06-26
chore: Added api guardian to build process

tomheller:

A verification step that checks if the api surface of the components has changed. A couple of things that I would appreciate @lukasholzer input on:

* How can we speed up the API check. right now attaching the workspace after the build takes about 5 minutes (granted that is for a full build, because I have changed the package.json in this one), but I guess there could be some improvements made.
* We should think about a way on how to interact with pr / forked prs the same way (commenting / labeling / ...). Due to the security contexts, it looks like a bot would a good way to go... but let's discuss.
* The output of the verify task would be a valid diff.patch file, maybe we can leverage that to accept / reject the api changes easily.

#### Type of PR

Other

#### Checklist

### DIFF
--- a/libs/barista-components/chart/src/selection-area/selection-area.ts
+++ b/libs/barista-components/chart/src/selection-area/selection-area.ts
@@ -128,7 +128,8 @@ import {
     '[attr.tabindex]': '0',
   },
 })
-export class DtChartSelectionArea implements AfterContentInit, OnDestroy {
+// tslint:disable-next-line: class-name
+export class _DtChartSelectionArea implements AfterContentInit, OnDestroy {
   /** @internal The timestamp that follows the mouse */
   @ViewChild('hairline', { static: true })
   _hairline: ElementRef<HTMLDivElement>;
--- a/libs/workspace/src/scripts/api-guard-verify.ts
+++ b/libs/workspace/src/scripts/api-guard-verify.ts
@@ -1,0 +1,87 @@
+/**
+ * @license
+ * Copyright 2020 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { options } from 'yargs';
+import { affectedArgs } from './affected-args';
+import { getAffectedProjects } from './util';
+import { executeCommand } from '@dynatrace/shared/node';
+import { writeFileSync } from 'fs';
+
+export async function runApiGuardVerify(): Promise<string | null> {
+  const { verifyDir, rootDir, allowModuleIdentifiers, _: inputs } = options({
+    verifyDir: {
+      type: 'string',
+      demandOption: true,
+      description: 'Read golden file structure from directory',
+    },
+    rootDir: {
+      type: 'string',
+      description: 'Specify the root directory of input files',
+    },
+    allowModuleIdentifiers: {
+      type: 'string',
+      description: 'Comma separated list of projects that should be excluded',
+    },
+    _: {
+      type: 'array',
+      description: '<entrypoint .d.ts files>',
+    },
+  }).argv;
+  const baseSha = await affectedArgs();
+  const projects = getAffectedProjects(baseSha, 'build');
+
+  // If the components are not affected, verification does not need to run.
+  if (!projects.includes('components')) {
+    console.log('No components were affected, api surface has not changed');
+    return null;
+  }
+
+  // Initialize the basic command that needs to be run.
+  const command = ['npx ts-api-guardian', `--verifyDir ${verifyDir}`];
+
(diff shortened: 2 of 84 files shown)


lukasholzer:
@tomheller I like the idea of the applying the `.patch` file this could be automatically done by a bot 🤖  by something like commenting `/barista-bot apply-patch`. 

I guess a bot would solve the second question as well as it should have the same permissions as we as users.

Regarding the speed it is a difficult topic as we have to download 1.4GB of persisted workspace:
```
workflows/workspaces/ff777176-b9a6-40a0-9f30-a4f5cf73c8e7/0/9feb5cf6-78b3-43f4-9bca-a8a5d62e91cd/0/106.tar.gz - 254 MB
  workflows/workspaces/ff777176-b9a6-40a0-9f30-a4f5cf73c8e7/0/a47f40e5-a9c8-4e81-8a69-795e11b17c8e/0/104.tar.gz - 312 MB
  workflows/workspaces/ff777176-b9a6-40a0-9f30-a4f5cf73c8e7/0/a47f40e5-a9c8-4e81-8a69-795e11b17c8e/1/104.tar.gz - 297 MB
  workflows/workspaces/ff777176-b9a6-40a0-9f30-a4f5cf73c8e7/0/a47f40e5-a9c8-4e81-8a69-795e11b17c8e/2/104.tar.gz - 254 MB
  workflows/workspaces/ff777176-b9a6-40a0-9f30-a4f5cf73c8e7/0/a47f40e5-a9c8-4e81-8a69-795e11b17c8e/3/104.tar.gz - 268 MB
```

This will take that long. The only big improvement would be the shared build cache with Bazel where we only have to download the barista components dist.

libs/barista-components/chart/src/selection-area/selection-area.ts
```diff
@@ -128,7 +128,8 @@ import {
     '[attr.tabindex]': '0',
   },
 })
-export class DtChartSelectionArea implements AfterContentInit, OnDestroy {
+// tslint:disable-next-line: class-name
+export class _DtChartSelectionArea implements AfterContentInit, OnDestroy {
```
lukasholzer, line 132: Why do we need to rename the className as it is exported with this name?

tomheller:
@lukasholzer I'm already looking into how the interaction with the bot can work. 
I will need to make some adjustments to this one, as we need to store the .patch as a build artifact to make it available for the bot. 

Regarding the build speed, I figured as much. It should be better if not everything is affected. Or we maybe able to build individual components (I think that was also on the table once, right?).

package.json
```diff
@@ -42,7 +42,9 @@
     "publish-release-next": "ng run tools-release:build-publish-release-next && node dist/libs/tools/publish-release-next/main.js",
     "barista:prerender": "ng run barista-design-system:prerender --routesFile dist/barista-data/routes.txt",
     "bazel:format": "bazel build //:buildifier",
-    "bazel:build": "bazel build //libs/workspace/src/builders/barista-build --remote_upload_local_results=true --google_credentials=tools/bazel_build_cache/cache-write-key.json"
+    "bazel:build": "bazel build //libs/workspace/src/builders/barista-build --remote_upload_local_results=true --google_credentials=tools/bazel_build_cache/cache-write-key.json",
+    "api-guard:generate": "ts-api-guardian --outDir api_guard --rootDir dist/libs/components/ --allowModuleIdentifiers Highcharts dist/libs/components/*/index.d.ts",
+    "api-guard:verify": "node dist/libs/workspace/src/scripts/api-guard-verify.js --verifyDir api_guard --rootDir dist/libs/components/ --allowModuleIdentifiers Highcharts dist/libs/components/*/index.d.ts"
```
lukasholzer, line 47: why do we have to provide Highchcarts as moduleIdentifier and only Highcharts?

lukasholzer:
> @lukasholzer I'm already looking into how the interaction with the bot can work.
> I will need to make some adjustments to this one, as we need to store the .patch as a build artifact to make it available for the bot.
> 
> Regarding the build speed, I figured as much. It should be better if not everything is affected. Or we maybe able to build individual components (I think that was also on the table once, right?).

Yes you are right with the individual component builds but for the library as long we don't have bazel we need to build all at least on the specific branches.

libs/barista-components/chart/src/selection-area/selection-area.ts
tomheller: The `ts-api-guard` script cannot deal with renames in exports and fails. That's why I renamed the class to prevent the renaming of the export.

package.json
tomheller: Highcharts is the only module within the components, that we need to import as `import * as Highcharts fomr 'highcharts';` These imports need to be in the allowList for the API guardian, i guess.

libs/workspace/src/scripts/api-guard-verify.ts
```diff
@@ -0,0 +1,89 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { execSync } from 'child_process';
+import { options } from 'yargs';
+import { affectedArgs } from './affected-args';
+import { getAffectedProjects } from './util';
+import { executeCommand } from '../../../shared/node/src/utils/execute-command';
```
lukasholzer, line 21: please use the path alias to import `@dynatrace/shared/node` it should be.

tomheller:
Pr has been blocked until we hear back from github about the bot integration especially regarding forks.

heartdisease (approved):

rowa-audil (approved):

PR #1218, opened 2020-07-02
chore(snyk): Remove outdated snyk ignores and extend lodash ignore

ffriedl89:

Since the lodash dep is inside a babel package we cannot really do something here - extending so we can recheck in September.

Moreover this does not affect our bundle at all.


lukasholzer:
@ffriedl89  we should start patching the vulnerabilities instead of ignoring with the `https://www.npmjs.com/package/patch-package`

PR #1225, opened 2020-07-03
chore: Disable windows tests.

tomheller:

Disabling windows test for now, as they use up a lot of credits.

#### Type of PR
Other

#### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 orbs:
   sonarcloud: sonarsource/sonarcloud@1.0.1
   snyk: snyk/snyk@0.0.8
-  win: circleci/windows@2.2.0
+  # win: circleci/windows@2.2.0
 
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 #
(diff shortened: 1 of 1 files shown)


lukasholzer:
So sad I really liked them 😢  – Does it make sense to run the windows workflow only on master or release branch at least? 
so putting to the install in the win-workflow a branch filter regex 

```yml

  # Workflow for running build and test on windows
  win-check:
    jobs:
      - install-win:
           filters:
             branches:
               only: /^([0-9]{1,}\.x|[0-9]{1,}\.[0-9]{1,}\.x)$/
          context: barista
      - build-win:
          context: barista
          requires:
            - install-win
      - unit-test-win:
          context: barista
          requires:
            - install-win

```

Then it would run only on the release branches at least we can detect it there. Or kind a smoke stage that runs every day at midnight?

P.s. It would be enough to uncomment only the workflow.

ffriedl89:
> So sad I really liked them 😢 – Does it make sense to run the windows workflow only on master or release branch at least?
> so putting to the install in the win-workflow a branch filter regex
> 
> ```yaml
>   # Workflow for running build and test on windows
>   win-check:
>     jobs:
>       - install-win:
>            filters:
>              branches:
>                only: /^([0-9]{1,}\.x|[0-9]{1,}\.[0-9]{1,}\.x)$/
>           context: barista
>       - build-win:
>           context: barista
>           requires:
>             - install-win
>       - unit-test-win:
>           context: barista
>           requires:
>             - install-win
> ```
> 
> Then it would run only on the release branches at least we can detect it there. Or kind a smoke stage that runs every day at midnight?
> 
> P.s. It would be enough to uncomment only the workflow.

I am for a smoke stage setup that runs once a day. Running on release branches does not work since this would run after each merge as well.

tomheller:
> I am for a smoke stage setup that runs once a day. Running on release branches does not work since this would run after each merge as well.

Can we merge this to get our credit usage down and create a follow up for the smoke setup?

PR #1233, opened 2020-07-06
chore: Fixes nginx routing for the design-system app.

tomheller:
#### Type of PR
Other

#### Checklist


tomheller:
> you could change the baseHref directly in the index.html as well.

Yes, but this would be also affecting local build and serve targets.

PR #1249, opened 2020-07-08
chore: Fix package to lock file inconsistencies.

tomheller:
package-lock ran out of sync, this should fix it.


tomheller:
> And what is resolved "" with an empty string?

I don't know 🤷

PR #1275, opened 2020-07-10
[table] limit dtIndicator styles inheritance

bartoszbobin:
#### Type of PR
Fixes issue where active indicator on expandable-row is inherited by nested table(s)  expandable-row(s) with indicator

## Situation when nested tables inherited dtIndicator styles from main expandable-row 
rows where indicator is visible but text have standard color shouldn't be marked with indicator
image

## After change
image

#### Checklist

- [-] I have added tests that prove my fix is effective or that my feature works
- [-] I have added necessary documentation (if appropriate)

### DIFF
--- a/libs/barista-components/table/src/expandable/_expandable-row-theme.scss
+++ b/libs/barista-components/table/src/expandable/_expandable-row-theme.scss
@@ -11,7 +11,7 @@
   }
 
   :host {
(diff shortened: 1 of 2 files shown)


tomheller:
We should check #1273 after this one is merged. This could potentially interfere with the custom properties solution.

Issue #1113, opened 2020-06-04
[Filter field]: Chaining of selected nodes to construct custom key:value pairs


airime:
# Feature Request

Allow dropdown selection of various filter options in succession, in order to compose a user-defined key:value pair.

## Summary

The `DtFilterField` component is currently implemented in a way that the first node you selected in the dropdown is used as the key, and after completing the filter the last node you selected is chosen as the value. This request is for adding a configuration that allows composing all the selected nodes (except for the last selected one) together in the key, and the last selected one as the value.

## Feature Description

The use case to cover is to go through a chain of selections and end up with a composed key:value pair of all the choices made by the user, where only the last choice the user made is the value and the rest are concatenated in the key. It is important that the user choices are made in different steps, and this feature is needed for a coming release.

Here are some screenshots of how it could look:

![Screenshot 2020-06-04 at 18 35 49](https://user-images.githubusercontent.com/6683824/83786021-6d745d00-a692-11ea-8449-753ff70e304f.png)
![Screenshot 2020-06-04 at 18 36 16](https://user-images.githubusercontent.com/6683824/83786043-78c78880-a692-11ea-891b-4ff04f97816e.png)
![Screenshot 2020-06-04 at 18 36 28](https://user-images.githubusercontent.com/6683824/83786064-7f560000-a692-11ea-9944-255fb1a7bcd7.png)
![Screenshot 2020-06-04 at 18 36 37](https://user-images.githubusercontent.com/6683824/83786072-8250f080-a692-11ea-8cdb-9ab5acccb625.png)

In order to achieve it, one way could be to add a new `concatenate` keyword that allows defining a separating character between nodes, so that, when several nested autocomplete options are configured, the component knows that they should be added together and the user knows all the path of options that was chosen with one look.

tomheller:
Hi @airime ! Thank you for opening up this feature request. 

I guess if we are talking about customizing the `key` displayed in the filter-field tag we should discuss a more generic approach that lets the consumer customize this based on a custom function that is provided. 

__Scope__ 
Do you know if these customizations need to happen on a filter-field or filter basis?
Is there anything besides the `key` that should be customizable for this use case (like the separator character between the key and the value)? 

__Possible solution__
[...]

PR #1276, opened 2020-07-10
feat(filter-field): Added config for user defined tag parser function

airime:

This PR addresses issue https://github.com/dynatrace-oss/barista/issues/1113.

#### Type of PR
 Feature

#### Checklist

### DIFF
--- a/libs/barista-components/filter-field/index.ts
+++ b/libs/barista-components/filter-field/index.ts
@@ -22,7 +22,11 @@ export * from './src/filter-field-range/filter-field-range-trigger';
 export * from './src/filter-field-data-source';
 export * from './src/filter-field-default-data-source';
 export * from './src/filter-field-errors';
-export { applyDtOptionIds, DELIMITER } from './src/filter-field-util';
+export {
+  applyDtOptionIds,
+  DELIMITER,
+  defaultTagDataForFilterValuesParser,
+} from './src/filter-field-util';
 export {
   DtNodeFlags,
   DtNodeDef,
@@ -44,4 +48,10 @@ export {
   dtGroupDef,
   isDtGroupDef,
   isDtRenderType,
+  DtFilterValue,
+  DtFilterFieldTagData,
+  isDtAutocompleteValue,
+  isDtFreeTextValue,
+  isDtRangeValue,
 } from './src/types';
+export { DT_FILTER_VALUES_PARSER_CONFIG } from './src/filter-field-config';
--- a/libs/barista-components/filter-field/src/filter-field-config.ts
+++ b/libs/barista-components/filter-field/src/filter-field-config.ts
@@ -1,0 +1,33 @@
+/**
+ * @license
+ * Copyright 2020 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { InjectionToken } from '@angular/core';
+import { defaultTagDataForFilterValuesParser } from './filter-field-util';
+import { DtFilterValue, DtFilterFieldTagData } from './types';
+
+/** User defined parser function for tag key:values, overrides default parser function */
+export type TagParserFunction = (
+  filterValues: DtFilterValue[],
+  editable?: boolean,
+  deletable?: boolean,
+) => DtFilterFieldTagData | null;
+
+/** Injection token for the external configuration of the filter-field component */
+export const DT_FILTER_VALUES_PARSER_CONFIG = new InjectionToken<
+  TagParserFunction
+>('dt-filter-value-config');
+
+export const DT_FILTER_VALUES_DEFAULT_PARSER_CONFIG: TagParserFunction = defaultTagDataForFilterValuesParser;
--- a/libs/barista-components/filter-field/src/filter-field-util.ts
+++ b/libs/barista-components/filter-field/src/filter-field-util.ts
@@ -18,24 +18,24 @@ import { isDefined } from '@dynatrace/barista-components/core';
 
 import { DtFilterFieldDataSource } from './filter-field-data-source';
 import {
-  _DtAutocompleteValue,
-  _DtFilterFieldTagData,
-  _DtFilterValue,
-  _DtFreeTextValue,
+  DtAutocompleteValue,
+  DtFilterFieldTagData,
+  DtFilterValue,
+  DtFreeTextValue,
   DtNodeDef,
-  _DtRangeValue,
+  DtRangeValue,
   dtAutocompleteDef,
   dtFreeTextDef,
   dtGroupDef,
   isAsyncDtAutocompleteDef,
   isDtAutocompleteDef,
-  _isDtAutocompleteValue,
+  isDtAutocompleteValue,
   isDtFreeTextDef,
   isDtFreeTextValue,
   isDtGroupDef,
   isDtOptionDef,
   isDtRangeDef,
-  _isDtRangeValue,
+  isDtRangeValue,
   isDtRenderType,
   isPartialDtAutocompleteDef,
   DtOptionDef,
@@ -261,7 +261,7 @@ function isOptionSelected(def: DtNodeDef, selectedIds: Set<string>): boolean {
 
 /** Transforms a RangeSource to the Tag data values. */
 function transformRangeSourceToTagData(
-  result: _DtRangeValue,
+  result: DtRangeValue,
 ): { separator: string; value: string } {
   if (result.operator === 'range') {
     return {
@@ -288,11 +288,11 @@ function transformRangeSourceToTagData(
   return { separator, value };
 }
 
-export function createTagDataForFilterValues(
-  filterValues: _DtFilterValue[],
+export function defaultTagDataForFilterValuesParser(
+  filterValues: DtFilterValue[],
   editable?: boolean,
   deletable?: boolean,
-): _DtFilterFieldTagData | null {
+): DtFilterFieldTagData | null {
   let key: string | null = null;
   let value: string | null = null;
   let separator: string | null = null;
@@ -300,7 +300,7 @@ export function createTagDataForFilterValues(
   let isFirstValue = true;
 
   for (const filterValue of filterValues) {
-    if (_isDtAutocompleteValue(filterValue)) {
+    if (isDtAutocompleteValue(filterValue)) {
       if (isFirstValue && filterValues.length > 1) {
         key = filterValue.option.viewValue;
       }
@@ -312,7 +312,7 @@ export function createTagDataForFilterValues(
         separator = '~';
       }
       break;
-    } else if (_isDtRangeValue(filterValue)) {
+    } else if (isDtRangeValue(filterValue)) {
       // Assigning variables destructed variables to already defined ones needs to be within braces.
       ({ value, separator } = transformRangeSourceToTagData(filterValue));
       break;
@@ -321,7 +321,7 @@ export function createTagDataForFilterValues(
   }
 
   return filterValues.length && value !== null
-    ? new _DtFilterFieldTagData(
+    ? new DtFilterFieldTagData(
         key,
         value,
         separator,
@@ -338,8 +338,8 @@ export function findFilterValuesForSources<T>(
   rootDef: DtNodeDef<T>,
   asyncDefs: Map<DtNodeDef<T>, DtNodeDef<T>>,
   dataSource: DtFilterFieldDataSource<T>,
-): _DtFilterValue[] | null {
-  const foundValues: _DtFilterValue[] = [];
+): DtFilterValue[] | null {
+  const foundValues: DtFilterValue[] = [];
   let parentDef = rootDef;
 
   for (let i = 0; i < sources.length; i++) {
@@ -352,7 +352,7 @@ export function findFilterValuesForSources<T>(
     if (
       isLastSource &&
       ((isDtFreeTextValue(source) && isDtFreeTextDef(parentDef)) ||
-        (_isDtRangeValue(source) && isDtRangeDef(parentDef)))
+        (isDtRangeValue(source) && isDtRangeDef(parentDef)))
     ) {
       foundValues.push(source);
       return foundValues;
@@ -365,7 +365,7 @@ export function findFilterValuesForSources<T>(
           const asyncDef = asyncDefs.get(def);
           if (asyncDef) {
             parentDef = asyncDef;
-            foundValues.push(def, asyncDef as _DtAutocompleteValue<T>);
+            foundValues.push(def, asyncDef as DtAutocompleteValue<T>);
           } else {
             parentDef = def;
             foundValues.push(def);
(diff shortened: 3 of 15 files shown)


libs/barista-components/filter-field/src/filter-field-util.ts
```diff
@@ -333,6 +333,53 @@ export function createTagDataForFilterValues(
     : null;
 }
 
+export function createComposedTagDataForFilterValues(
```
tomheller, line 336: Exported functions need a JsDoc.

airime:
Thank you for your feedback, @tomheller. Some thoughts regarding the two points you asked more feedback about (the size of the filter-field test file and the difficulty of navigating that context, and the exposure of several internal utility functions):

If we want to limit ourselves strictly to solving issue https://github.com/dynatrace-oss/barista/issues/1113, the real change that adds the described functionality can be found on lines 351-353 of the filter-field-util.ts file: https://github.com/dynatrace-oss/barista/pull/1276/commits/1cb7d80366122e532a4433b75db671a51e8e830d#diff-18f0994266e5cce70a76b1b00c168394R351

As you can see, aside from the newly added else if clause, the rest of the new custom parser `createComposedTagDataForFilterValues` function is exactly the same as the old `createTagDataForFilterValues` function. It looks to me that a simpler implementation could be to add a boolean @Input property to enable concatenation of keys (default false) so that the user can choose what strategy to use, without the need of an injection token or extensive modification of the test contexts. However, this proposal seems to be in contradiction of the [API design coding standards](https://github.com/dynatrace-oss/barista/blob/master/CODING_STANDARDS.md#api-design), so I'd like to know what do you think about it before changing anything.

tomheller:
Hi @airime!

A couple of things:
1. The size of the filter field spec file was merely a note to fabian and thomas pink, that we eventually should think about splitting up the test file, as it is already too large. That is nothing that should concern you. 
2. Please remove your modification for the concatenation of the keys from the filter-field utils file. This function serves as the default behaviour and should stay the same as it is now. 
3. Yes, I agree that it would be simpler to have a switch for the concatenation. But this will not scale as there will be other teams that have other use cases and ideas on how they would want to display their tag data, which would lead to the component team needing to maintain all of these and also a lot of input switches that control these. Therefore it is much more feasible to expose the function and have the consumer provide a function that can customize the key however they want.

A consumers custom function could eventually look like this:
```ts
import { createComposedTagDataForFilterValues } from '@dynatrace/barista-components/filter-field';

function myCustomParser(filterValues: _DtFilterValue[], editable?: boolean, deletable?: boolean) {
  const tagData = createComposedTagDataForFilterValues(filterValues, editable, deletable);
   // There is probably a lot more logic involved, but for the sake of the example, I'm going to glance over this.
  tagData.key = filterValues.map(f => f.option.viewValue).join('.');
  return tagData;
}
```

This way the consumer can still leverage all of the logic and default behaviour provided by the barista-components, but has the ability to change certain aspects afterwards.

airime:
Hello all, it seems that my GitHub account was not set up properly on my first commit and now I have a commit author that cannot be linked to any email. This seems to be causing trouble for this PR to be merge-ready, as it is blocking the signing of the CLA, and in order to correct the previous commit I would need to rewrite history, if I understand [this guide](https://docs.github.com/en/github/committing-changes-to-your-project/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user) correctly. Do I need to ask permission for this?

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -429,7 +433,11 @@ export class DtFilterField<T = any>
     // tslint:disable-next-line:no-any
     @Optional() @Inject(DOCUMENT) private _document: any,
     defaultErrorStateMatcher: ErrorStateMatcher,
+    @Optional()
+    @Inject(DT_FILTER_VALUES_CONFIG)
+    private _config: DtFilterValuesConfig /** @internal used for the selection area to calculate the bounding client rect */,
```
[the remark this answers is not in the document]
airime: Indeed it was. Good catch!

airime:
Hi @tomheller! Thank you for your explanations, I understand better the coding standards now.

Regarding the custom `createComposedTagDataForFilterValues` function that is present in the utils file, I was not aware on the usage of this file, thank you for your clarification. It is currently only being used in the spec file, so, if you agree, I'll move it there once there is more feedback on how to expose the utility functions to the public (so that the newly moved function can use them already in the proper way).

tomheller:
>
I would need to rewrite history, if I understand this guide correctly. Do I need to ask permission for this?

On your own branch, before merging, you con rewrite history as often as you like :)

> Regarding the custom createComposedTagDataForFilterValues function that is present in the utils file, I was not aware on the usage of this file, thank you for your clarification.

Yes, you can move it into the spec file. You could also try the way I have suggested in the spec file, as this is I guess the way that most will use it. It's seems like the least intrusive to the internal workings of the filter field.

libs/barista-components/filter-field/src/filter-field-config.ts
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { InjectionToken } from '@angular/core';
+import { createTagDataForFilterValues } from './filter-field-util';
+
+/** User defined configurations that can override some settings of the filter-field component */
+export interface DtFilterValuesConfig {
+  tagValuesParser: Function;
```
thomaspink, line 22: Instead of `Function` let's use a better type here. Something like:
```suggestion
  tagValuesParser: (filterValues: DtFilterValue[], editable?: boolean, deletable?: boolean) => DtFilterFieldTagData | null;
```

Note:
We also need to change the types `_DtFilterValue` and `_DtFilterFieldTagData` to be public so the consumer can use it.

`_DtFilterValue` -> `DtFilterValue`
`_DtFilterFieldTagData` -> `DtFilterFieldTagData`

thomaspink:
> A consumers custom function could eventually look like this:
>
> import { createComposedTagDataForFilterValues } from '@dynatrace/barista-components/filter-field';
> ```ts
> function myCustomParser(filterValues: _DtFilterValue[], editable?: boolean, deletable?: boolean) {
>   const tagData = createComposedTagDataForFilterValues(filterValues, editable, deletable);
>   // There is probably a lot more logic involved, but for the sake of the example, I'm going to glance over this.
>   tagData.key = filterValues.map(f => f.option.viewValue).join('.');
>   return tagData;
> }
> ```
> This way the consumer can still leverage all of the logic and default behaviour provided by the barista-components, but > has the ability to change certain aspects afterwards.

@tomheller I agree that we should expose our default parser function to the consumer so that it can be used if needed (as you have shown in the example) but the default case i think should rather be that the consumer just writes it's own parser function.

Also i am not so sure about the naming of these function. Maybe calling it just a "default" function for the parser makes more sense to indicate to the consumer that this is just the default way of generating tag data, but it is easy to override it.

libs/barista-components/filter-field/src/filter-field.spec.ts
```diff
@@ -232,14 +211,62 @@ describe('DtFilterField', () => {
...
   ));
 
-  it('should focus the input field when focusing the host', () => {
-    const input = fixture.debugElement.query(By.css('.dt-filter-field-input'))
-      .nativeElement;
-    filterField.focus();
-    expect(document.activeElement).toBe(input);
+  describe('focus on input', () => {
+    beforeEach(fakeAsync(() => {
+      TestBed.configureTestingModule({
```
ffriedl89, line 216: you don't have to copy the same beforeEach into every describe. I guess you did that because you added another describe at the end that needs a different setup. 
You can move the last describe onto the root level - e.g. DtFilterField with FilterValueConfig on the root. 
And leave the old setup in the before each where it was.

libs/barista-components/filter-field/src/filter-field-config.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+/** Injection token for the external configuration of the filter-field component */
+export const DT_FILTER_VALUES_CONFIG = new InjectionToken<DtFilterValuesConfig>(
+  'dt-filter-field-config',
```
ffriedl89, line 27: NIT: 
```suggestion
  'dt-filter-value-config',
```

ffriedl89:
Please also see the CLA notice above - you need to sign the Contributors License Agreement and therefore have a github account associated with email address used for the commit.

libs/barista-components/filter-field/src/filter-field-config.ts
thomaspink: Update: I do not see the need for the params `editable` & `deletable`. Because either the consumer sets these properties for all generated tag the same, so this can be set statically in the custom parser function or the values of these properties depend on the filter values itself which are already provided as a param.

thomaspink: So i would suggest the type: `(filterValues: DtFilterValue[]) => DtFilterFieldTagData | null`

libs/barista-components/filter-field/src/filter-field-config.ts
```diff
@@ -0,0 +1,32 @@
...
+import { InjectionToken } from '@angular/core';
+import { createTagDataForFilterValues } from './filter-field-util';
+
+/** User defined configurations that can override some settings of the filter-field component */
+export interface DtFilterValuesConfig {
+  tagValuesParser: Function;
+}
+
+/** Injection token for the external configuration of the filter-field component */
+export const DT_FILTER_VALUES_CONFIG = new InjectionToken<DtFilterValuesConfig>(
```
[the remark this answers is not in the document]
tomheller: > We should also consider that the function may be passed into the filter field via an input.

This has been discussed previously and I would rather not add that one via an input. 
It does not need to change at runtime and could potentially be really hard to handle...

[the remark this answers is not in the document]
thomaspink: > This has been discussed previously and I would rather not add that one via an input.
> It does not need to change at runtime and could potentially be really hard to handle...

The issue for me is, that without an additional input it is way harder to provide a parser function for just one filter-field without creating side effects for other filter-fields in child components.

e.g.

Root (defines parser function for all filter fields via DI
  - component (has filter field, uses root parser function because of DI)
    - child component (has filter field, needs different parser function which is set as a provider on this level)
      - child child component (has filter field, needs root parser function but inherits the function of its parent component even)

With an additional input the child component can set the special parser function via input and all other components (even child child component) inherit the correct root parser function.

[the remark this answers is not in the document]
airime: From what I see in the Barista documentation, there is a BaLiveExample component that hides by default the component decorator configuration, which is the place where the injection token would be provided. I don't see a way to configure the BaLiveExample component so that this configuration can be shown, so I guess that in order for the usage to be shown in the documentation, the component would have to be modified. If we go with the input approach however, this modification would not be needed.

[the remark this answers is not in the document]
tomheller: I'm still not happy about the input approach. This setting to me screams `configuration`, which I don't think should be changed at runtime. I understand the scoping problem of the DI, that you have outlined @thomaspink.

But if we have an input, and the input function changes, we would need to run though all active tag elements again and rerender them. This could maybe lead to some sort of flashing while the input is being evaluated. 

I'd still opt for the InjectionProvider... but if you think the input approach is feasible as well, I'm fine with that.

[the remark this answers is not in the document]
thomaspink: Setting the function via DI should definitely be implemented. Not sure if i made it clear enough, the input should only be an extra option for these specific cases. So it is working the same way, for example, as the select uses the `ErrorStateMatcher` which can either be provided via DI or for special cases an Input.

I do not see the issue @tomheller has mentioned that the input function changes at runtime which could lead to rerendering the tags: 1. There is probably no use case for changing it at runtime. The function will be set on init in i guess all of the cases. Maybe you can help me with a use case where the runtime change is needed. 2. Even if the function changes at runtime, i do not see a problem. Input changes that causes a rerender in the case for 90% of the components. Also because the evaluation is done before rendering, i do not see why it would cause flashing. Maybe you can help me out with your thoughts there.

Regarding the effort, it should be kinda easy to add the input and override the value provided by the DI if an input is set. But I can agree to a compromise, as it handles only a specific use case, that we stick to the DI-only approach right now and add the input later if needed. :)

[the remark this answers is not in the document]
tomheller: > Not sure if i made it clear enough

Sorry, I guess I misunderstood and thought the input would be considered a replacement for the Di provided funciton. 

> 1. There is probably no use case for changing it at runtime.

There probably isn't a use case for it, but if it is there it should also work as expected, i.e. filters are set programmatically right on init and may be set before the function is set. But you are right, if the input changes is will cause re-rendering. I just wanted to make sure that it is not only re-renderd, but also the already committed tags run through the new function again. But as you said, this is probably not a very common use case, so I'm fine with the additional input to match the pattern of the `ErrorStateMatcher` 
👍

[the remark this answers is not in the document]
airime: Good points, both of you. I added an input option that takes priority over the injected configuration if it exists, let me know if this is what you had in mind.

libs/barista-components/filter-field/index.ts
```diff
@@ -22,7 +22,11 @@ export * from './src/filter-field-range/filter-field-range-trigger';
 export * from './src/filter-field-data-source';
 export * from './src/filter-field-default-data-source';
 export * from './src/filter-field-errors';
-export { applyDtOptionIds, DELIMITER } from './src/filter-field-util';
+export {
+  applyDtOptionIds,
+  DELIMITER,
+  defaultTagDataForFilterValuesParser,
```
tomheller, line 28: @thomaspink If we export them now, do we need to prefix or at least annotate them to be from the library?

```suggestion
  applyDtOptionIds,
  DT_FILTER_FIELD_DELIMITER,
  dtDefaultTagDataForFilterValuesParser,
```

libs/barista-components/filter-field/src/filter-field-config.ts
```diff
@@ -0,0 +1,37 @@
...
+  'dt-filter-value-config',
+);
+
+export const DT_FILTER_VALUES_DEFAULT_CONFIG: DtFilterValuesConfig = {
+  tagValuesParser: defaultTagDataForFilterValuesParser,
+};
```
tomheller, line 37: Maybe we can create an injection token for the tagValuesParser function itself. We try to stay away from what can potentially be a larger configuration object. If we add an injection token for each of the configs, we can control this more granular.

libs/barista-components/filter-field/src/filter-field-util.ts
```diff
@@ -288,11 +288,57 @@ function transformRangeSourceToTagData(
...
+        separator,
+        isFreeText,
+        filterValues,
+        editable,
+        deletable,
+      )
+    : null;
+}
+
+/*
```
tomheller, line 336: Please remove commented code, if it is no longer needed.

libs/barista-components/filter-field/src/filter-field.spec.ts
```diff
@@ -38,6 +38,7 @@ import {
   dtRangeDef,
```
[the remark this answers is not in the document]
tomheller: Will be handled in #1296

airime:
@tomheller Thanks for the feedback! I implemented more of your suggestions now.

@thomaspink @ffriedl89 I refined this PR a bit more, whenever yo can could you take a look at it again? Thanks!

libs/barista-components/filter-field/index.ts
thomaspink: would suggest so but not that important for me

thomaspink:
LGTM. Thank you

tomheller:
@airime Everything looks good now, I think we can merge this one. 
After I have merged it I will cherry pick the merged commit into the respective release branch. 

Thank you again for this contribution, great work 💯  🎉

tomheller:
Cherry picked into 7.6.x

Issue #1250, opened 2020-07-08
Next top nav: fetching next categories from strapi and displaying them


rowa-audil:
# Feature Request

## Summary

Currently we have our top bar navigation items hard-coded. Implementing a service into the page-service that returns an array of routes to each category would solve this.

## Feature Description

A Service that looks into what the next categories are that then are displayed in the top bar navigation.
Find a way to differentiate between current and next categories. A possible solution would be to expand the current next builder to also include categories. In there is the evaluation logic determining what next categories are and saves it in a json file into the dist folder (e.g dist/next-data/categories.json)

The determination could look into the relation / foreign-key which are pages and nextpages and see whether there are any nextpages.

PR #1278, opened 2020-07-13
Next(top-nav): Section generator

rowa-audil:
This PR removes the hardcoded top nav sections and adds a generator to the barista-tools, which creates a file containing the corresponding categories and saves them as a string array.
The top nav then displays that array.

Fixes: #1250

### DIFF
--- a/apps/next/src/app/components/nav/nav.html
+++ b/apps/next/src/app/components/nav/nav.html
@@ -5,12 +5,12 @@
     </a>
     <nav class="next-nav-bar">
       <ul class="next-nav-list">
-        <li class="next-nav-item" *ngFor="let item of (_navData$)?.navItems">
+        <li class="next-nav-item" *ngFor="let item of (_navData$ | async)">
           <a
             class="next-nav-link"
-            [routerLink]="item.url"
+            [routerLink]="item.link"
             routerLinkActive="next-nav-link-active"
-            >{{ item.label }}</a
+            >{{ item.section }}</a
           >
         </li>
       </ul>
--- a/apps/next/src/app/components/nav/nav.ts
+++ b/apps/next/src/app/components/nav/nav.ts
@@ -16,6 +16,8 @@
 
 import { Component, ChangeDetectionStrategy } from '@angular/core';
 import { ThemeService } from '../../theme.service';
+import { BaPageService } from '@dynatrace/shared/data-access-strapi';
+import { map } from 'rxjs/operators';
 
 @Component({
   selector: 'next-nav',
@@ -28,30 +30,16 @@ import { ThemeService } from '../../theme.service';
 })
 export class Nav {
   /** @internal Data needed to render the navigation. */
-  _navData$ = {
-    navItems: [
-      {
-        url: 'overview',
-        label: 'Overview',
-      },
-      {
-        url: 'design',
-        label: 'Design',
-      },
-      {
-        url: 'develop',
-        label: 'Develop',
-      },
-      {
-        url: 'components',
-        label: 'Components',
-      },
-      {
-        url: 'design-tokens',
-        label: 'Design Tokens',
-      },
-    ],
-  };
-
-  constructor(public _themeService: ThemeService) {}
+  _navData$ = this._pageService.getCategories().pipe(
+    map((data) =>
+      data.map((category) => ({
+        section: category,
+        link: category.toLowerCase(),
+      })),
+    ),
+  );
+  constructor(
+    public _themeService: ThemeService,
+    private _pageService: BaPageService,
+  ) {}
 }
--- a/libs/shared/data-access-strapi/src/lib/page.service.ts
+++ b/libs/shared/data-access-strapi/src/lib/page.service.ts
@@ -84,6 +84,17 @@ export class BaPageService<T = any> {
   }
 
   /**
+   * Provides an array of categories for the Design System as an observable.
+   */
+  getCategories(): Observable<string[]> {
+    const requestPath = `${this._baseHref}/data/categories.json`;
+
+    return this._http.get<string[]>(requestPath, {
+      responseType: 'json',
+    });
+  }
+
+  /**
    * Fetches page from data source.
    * @param id - page id (path).
    */
--- a/libs/tools/barista/src/builder/strapi.ts
+++ b/libs/tools/barista/src/builder/strapi.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { join } from 'path';
+import { join, dirname } from 'path';
 
 import {
   BaSinglePageMeta,
@@ -31,6 +31,7 @@ import {
   BaStrapiContentType,
   NextStrapiPage,
   NextContentType,
+  BaStrapiCategory,
 } from '../types';
 
 import { fetchContentList } from '@dynatrace/shared/data-access-strapi';
@@ -44,6 +45,7 @@ import {
   relativeUrlTransformer,
   tableOfContentGenerator,
 } from '../transform';
+import { mkdirSync, writeFileSync } from 'fs';
 
 const TRANSFORMERS: BaPageTransformer[] = [
   markdownToHtmlTransformer,
@@ -57,10 +59,10 @@ const TRANSFORMERS: BaPageTransformer[] = [
 export const strapiBuilder: BaPageBuilder = async (
   globalTransformers: BaPageTransformer[],
   next: boolean = false,
+  dsEnvironment: { distDir: string },
 ) => {
   // Return here if no endpoint is given.
   if (!environment.strapiEndpoint) {
-    console.log('No Strapi endpoint given.');
     return [];
   }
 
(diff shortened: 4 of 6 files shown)


apps/next/src/app/components/nav/nav.html
```diff
@@ -5,12 +5,12 @@
     </a>
     <nav class="next-nav-bar">
       <ul class="next-nav-list">
-        <li class="next-nav-item" *ngFor="let item of (_navData$)?.navItems">
+        <li class="next-nav-item" *ngFor="let item of _navItems">
           <a
             class="next-nav-link"
-            [routerLink]="item.url"
+            [routerLink]="item.toLowerCase()"
```
tomheller, line 11: Please prepare the item in the typescript file, logic like toLowerCase should not be handled in the template.
```suggestion
            [routerLink]="item.toLowerCase()"
```

apps/next/src/app/components/nav/nav.ts
```diff
@@ -27,31 +32,17 @@ import { ThemeService } from '../../theme.service';
   },
 })
 export class Nav {
+  _navData$ = this._pageService._getCategories().subscribe((data) => {
+    this._navItems.push(...data);
+    this._changeDetectorRef.markForCheck(); // Angular doesn't understand that the array values changed.
+  });
```
tomheller, line 38: Please either subscribe in the constructor or in ngOnInit.
Also don't forget to unsubscribe.

apps/next/src/app/components/nav/nav.ts
```diff
@@ -27,31 +32,17 @@ import { ThemeService } from '../../theme.service';
   },
 })
 export class Nav {
+  _navData$ = this._pageService._getCategories().subscribe((data) => {
+    this._navItems.push(...data);
```
tomheller, line 36: Why do you need to push these, are these not all categories that are coming in? 
Maybe we do not need to subscribe in the component at all?

rowa-audil: Regarding the subscription, I've tried to get the categories with a http get request and to return them as an array to no avail. Anything other the `{ responseType: 'json' }` returns an ArrayBuffer and I would rather not use that in our service.

tomheller: The `responseType: 'json'` is fine, but you could this here:
```suggestion
    _navData$ = this._pageService._getCategories();
```

and use this in the template like this:
```html
<li class="next-nav-item" *ngFor="let item of (_navData | async)">
```

apps/next/src/app/components/nav/nav.ts
```diff
@@ -26,32 +34,29 @@ import { ThemeService } from '../../theme.service';
     class: 'next-nav',
   },
 })
-export class Nav {
+export class Nav implements OnInit, OnDestroy {
+  categorySubscription: Subscription;
```
tomheller, line 38: It's unused, please remove it.
```suggestion
```

apps/next/src/app/components/nav/nav.ts
```diff
@@ -26,32 +34,29 @@ import { ThemeService } from '../../theme.service';
...
+          section: category,
+          link: category.toLowerCase(),
+        })),
+      ),
+    );
+  }
+
+  ngOnDestroy(): void {
+    this.categorySubscription.unsubscribe();
+  }
```
tomheller, line 61: It's unused, please remove it.
```suggestion
```

apps/next/src/app/components/nav/nav.ts
```diff
@@ -26,32 +34,29 @@ import { ThemeService } from '../../theme.service';
...
-      },
-      {
-        url: 'design-tokens',
-        label: 'Design Tokens',
-      },
-    ],
-  };
 
-  constructor(public _themeService: ThemeService) {}
+  _navData$: Observable<any>;
```
tomheller, line 41:
```suggestion
  _navData$: Observable<{ section: string; link: string; }>;
```

tomheller:
Looks good to me, only needs a rebase now 👍

libs/tools/barista/src/builder/strapi.ts
```diff
@@ -96,9 +117,28 @@ export const strapiBuilder: BaPageBuilder = async (
...
+function writeCategoriesJson(distDir: string, categories: string[]): void {
+  const outFile = join(distDir, 'categories.json');
+
+  // Creating folder path if it does not exist
+  mkdirSync(dirname(outFile), { recursive: true });
+
+  // Write file with page content to disc.
+  // tslint:disable-next-line: no-magic-numbers
+  fs.writeFile(outFile, JSON.stringify(categories), {
+    flag: 'w', // "w" -> Create file if it does not exist
```
lukasholzer, line 137: This should be default

libs/tools/barista/src/builder/strapi.ts
```diff
@@ -96,9 +117,28 @@ export const strapiBuilder: BaPageBuilder = async (
     transformed.push({ pageContent, relativeOutFile });
   }
 
+  categories.push(...categoriesData.map((category) => category.title));
+
+  const distDir = next ? 'dist/next-data' : 'dist/barista-data';
```
lukasholzer, line 122: Can you make this configurable via the options that are passed to the barista builder – there should be an option for the `outputPath` that points to the next data instead of the barista data for the other build no hardcoded paths inside the code makes refactoring hard.

apps/next/src/app/components/nav/nav.html
```diff
@@ -5,12 +5,12 @@
     </a>
     <nav class="next-nav-bar">
       <ul class="next-nav-list">
-        <li class="next-nav-item" *ngFor="let item of (_navData$)?.navItems">
+        <li class="next-nav-item" *ngFor="let item of (_navData$ | async)">
```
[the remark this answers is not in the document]
tomheller: But it is nicer too read when in a loop. I would consider this a nitpick

libs/shared/data-access-strapi/src/lib/page.service.ts
```diff
@@ -83,6 +83,14 @@ export class BaPageService<T = any> {
     return of(this._cache.get(key)!);
   }
 
+  _getCategories(): Observable<string[]> {
```
ffriedl89, line 86:
```suggestion
  getCategories(): Observable<string[]> {
```
This is the public API of the service, we should not need a prefix here.

libs/tools/barista/src/builder/strapi.ts
```diff
@@ -76,12 +78,31 @@ export const strapiBuilder: BaPageBuilder = async (
...
+  // Results from transforming the pagecontent
   const transformed: BaPageBuildResult[] = [];
 
+  // Array of categories for corresponding design systems
+  const categories: string[] = [];
+
   for (const page of pagesData) {
+    if (next && page.category === null) {
+      categories.push(page.title);
+    }
```
ffriedl89, line 105: could you add a comment here please - why if there is no category we use the title instead. Maybe use an example to outline it more clearly.

apps/next/src/app/components/nav/nav.html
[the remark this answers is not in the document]
lukasholzer: it is a nit you are right :D

apps/next/src/app/components/nav/nav.ts
```diff
@@ -26,32 +28,20 @@ import { ThemeService } from '../../theme.service';
...
+      })),
+    ),
+  );
+  constructor(
+    public _themeService: ThemeService,
+    private _pageService: BaPageService,
+  ) {}
 
-  constructor(public _themeService: ThemeService) {}
+  ngOnInit(): void {}
```
lukasholzer, line 46: if not needed please remove

libs/shared/data-access-strapi/src/lib/page.service.ts
```diff
@@ -83,6 +83,14 @@ export class BaPageService<T = any> {
     return of(this._cache.get(key)!);
   }
 
+  getCategories(): Observable<string[]> {
```
lukasholzer, line 86: please add JsDoc to method

libs/tools/barista/src/main.ts
```diff
@@ -101,6 +102,7 @@ async function buildPages(): Promise<void[]> {
     linkReplacement: { type: 'string', alias: 'lr' },
   }).argv;
   const environment = next ? nextEnvironment : baristaEnvironment;
+  console.log(next, environment);
```
lukasholzer, line 105: please remove console.log

libs/tools/barista/src/builder/strapi.ts
```diff
@@ -96,9 +123,32 @@ export const strapiBuilder: BaPageBuilder = async (
...
 };
 
+/**
+ * Creates a file containing an array of the categories from strapi
+ */
+function writeCategoriesJson(
+  dsEnvironment: { distDir: string },
+  categories: string[],
+): void {
+  console.log(dsEnvironment.distDir);
```
lukasholzer, line 140: please remove logging

libs/shared/data-access-strapi/src/lib/page.service.ts
```diff
@@ -83,6 +83,17 @@ export class BaPageService<T = any> {
     return of(this._cache.get(key)!);
   }
 
+  /**
+   * Get requests the categories for the Design System.
```
lukasholzer, line 87: NIT:
```suggestion
   * Provides an array of all the categories that are needed in the design system as an observable
```

libs/tools/barista/src/builder/strapi.ts
lukasholzer: @rowa-audil Please use next time for comments inside the code the `//` annotation and not the jsDoc. JsDoc is used to annotate a method, property, etc but for code comments always use `//` as there is a difference on cleaning up the code.

libs/tools/barista/src/builder/strapi.ts
```diff
@@ -76,12 +78,36 @@ export const strapiBuilder: BaPageBuilder = async (
...
 
+  // Array of categories for corresponding design systems
+  const categories: string[] = [];
+
   for (const page of pagesData) {
+    /**
+     * Check if any pages have no a category assigned. These pages would be considered main pages.
+     * E.g. A page like the overview page describing the Design System itself does not relate to any category (angular-components or design-tokens for example).
+     * It is considered a main (or landing page) that the user should be able to navigate to.
+     */
```
lukasholzer, line 107: @rowa-audil Please use next time for comments inside the code the `//` annotation and not the jsDoc. JsDoc is used to annotate a method, property, etc but for code comments always use `//` as there is a difference on cleaning up the code.

```suggestion
    
     // Check if any pages have no a category assigned. These pages would be considered main pages.
     // E.g. A page like the overview page describing the Design System itself does not relate to any category (angular-components or design-tokens for example).
     // It is considered a main (or landing page) that the user should be able to navigate to.
```

rowa-audil: Ah right, thank you!

lukasholzer:
needs rebase now :)

Issue #1215, opened 2020-07-02
Next: Cleanup Barista references & styles


ffriedl89:
# Quality improvement

## Shared interfaces, functions and services

We copied or reused interfaces and functions from Barista. We should cleanup those things. Everything that is copied should be moved into a shared library, either shared-design-system-ui or the data-access library. 
The shared interfaces should be cleaned up - some of them are already addressed in #1205 - Let's double check to make sure this is everything. 

## Styles
We have some global styles that are gonna get problematic - e.g. button or anchor 
Scope them properly to components. 
Go over all styles and check if some properties can be replaced by using the design tokens - e.g. font-size, font-weight, spacings.

PR #1279, opened 2020-07-13
Design Systems: refactor shared things

rowa-audil:
Refactors shared things and uses design tokens wherever they could be used.

Fixes: #1215

### DIFF
--- a/apps/next/src/app/components/nav/nav.scss
+++ b/apps/next/src/app/components/nav/nav.scss
@@ -21,6 +21,7 @@
   padding-left: $fluid-spacing--small;
   display: flex;
   align-items: center;
+  border: none;
 }
 
 .next-nav-item {
--- a/apps/next/src/app/pages/single-page/single-page.ts
+++ b/apps/next/src/app/pages/single-page/single-page.ts
@@ -16,19 +16,19 @@
 
 import { Component } from '@angular/core';
 import { BaSinglePageContent } from '@dynatrace/shared/design-system/interfaces';
-import { BaPageService } from '@dynatrace/shared/data-access-strapi';
+import { DsPageService } from '@dynatrace/shared/data-access-strapi';
 
 @Component({
-  selector: 'next-single-page',
+  selector: 'ds-single-page',
   templateUrl: 'single-page.html',
   styleUrls: ['single-page.scss'],
   host: {
-    class: 'ba-page',
+    class: 'ds-single-page',
   },
 })
-export class NextSinglePage {
+export class SinglePage {
   /** @internal The current page content from the cms */
   _pageContent = this._pageService._getCurrentPage();
 
-  constructor(private _pageService: BaPageService<BaSinglePageContent>) {}
+  constructor(private _pageService: DsPageService<BaSinglePageContent>) {}
 }
--- a/apps/next/src/app/styles/_page.scss
+++ b/apps/next/src/app/styles/_page.scss
@@ -1,3 +1,5 @@
+@import '~@dynatrace/fluid-design-tokens/global/spacing';
+
 .next-page-grid {
   display: grid;
   grid-template-columns: minmax(0, 1fr);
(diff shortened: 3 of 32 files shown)


apps/next/src/styles.scss
```diff
@@ -26,16 +27,16 @@ html {
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
   text-rendering: optimizeLegibility;
-  font-size: 16px;
+  font-size: $fluid-spacing--medium;
```
tomheller, line 30: No, please don't use spacing tokens to assign font sizes, just because their units match by happenstance. 
There are tokens for body text, please use these.

apps/barista-design-system/src/pages/.search-page/search-page.module.ts
```diff
@@ -17,15 +17,15 @@
 import { CommonModule } from '@angular/common';
```
tomheller, line 17: search page is deprecated. You could delete it.

rowa-audil: Here: #1285

apps/next/src/app/components/nav/nav.scss
```diff
@@ -60,5 +65,5 @@
 }
 
 .next-nav-link-active {
-  border-bottom: var(--color-primary-100) solid 2px;
+  border-bottom: var(--color-primary-100) solid $fluid-spacing--3x-small;
```
tomheller, line 68: Please do not use spacing tokens for border sizes. Eventually we will have tokens for border widths as well.

As a general rule, spacing tokens should only be used for things that create spacing, i.e.:
* margin
* padding
* top / left / right / bottom
* transforms
* grid gaps

apps/next/src/app/styles/_page.scss
```diff
@@ -46,7 +48,8 @@
       'sidenav title .'
       'sidenav main meta'
       'sidenav footer .';
-    padding: 1rem 1rem 1rem 0;
+    padding: $fluid-spacing--medium $fluid-spacing--medium
+      $fluid-spacing--medium 0;
```
tomheller, line 52: Nitpick:
```suggestion
    padding: $fluid-spacing--medium $fluid-spacing--medium
      $fluid-spacing--medium $fluid-spacing--0;
```

apps/next/src/app/components/nav/nav.scss
```diff
@@ -21,6 +21,11 @@
   padding-left: $fluid-spacing--small;
   display: flex;
   align-items: center;
+  border: none;
+
+  &:focus {
+    outline: none;
+  }
```
ffriedl89, line 28: Please don't remove the outline for focus - this is crucial for a11y. I can live with the default browser focus style.

apps/next/src/app/pages/single-page/single-page.ts
```diff
@@ -16,19 +16,19 @@
 
 import { Component } from '@angular/core';
 import { BaSinglePageContent } from '@dynatrace/shared/design-system/interfaces';
-import { BaPageService } from '@dynatrace/shared/data-access-strapi';
+import { DsPageService } from '@dynatrace/shared/data-access-strapi';
 
 @Component({
   selector: 'next-single-page',
```
ffriedl89, line 22: Can we also change the selector to `ds-single-page`?

apps/next/src/app/pages/single-page/single-page.ts
```diff
@@ -16,19 +16,19 @@
...
   templateUrl: 'single-page.html',
   styleUrls: ['single-page.scss'],
   host: {
-    class: 'ba-page',
+    class: 'next-page',
```
ffriedl89, line 26:
```suggestion
    class: 'ds-single-page',
```

apps/next/src/app/pages/single-page/single-page.ts
```diff
@@ -16,19 +16,19 @@
...
   },
 })
 export class NextSinglePage {
```
ffriedl89, line 29: Remove the prefix on classes? This should be the default from the angular cli
@tomheller what do you think? 
```suggestion
export class SinglePage {
```

apps/next/src/app/pages/single-page/single-page.ts
ffriedl89: I think we should get rid of the next entirely except for the deployUrl being /next
This is going to be the ds going forward

tomheller:
Cherry picked manually to 7.x and 7.4.x

PR #1285, opened 2020-07-13
barista: Removes the deprecated search-page

rowa-audil:


tomheller:
Please reword the commit, this is not a feature concerning the barista component library rather than a design system commit.

PR #1292, opened 2020-07-15
style(select): Fixes border issue when select is used inside form-field

bartoszbobin:

Can be reproduce here:
https://barista.dynatrace.com/components/select#form-field

Problem is more visible on Firefox
image

#### Checklist

### DIFF
--- a/libs/barista-components/select/src/select.scss
+++ b/libs/barista-components/select/src/select.scss
@@ -34,7 +34,6 @@ $dt-select-panel-min-width: 112px !default;
 
   .dt-form-field-infix > & {
     border: none;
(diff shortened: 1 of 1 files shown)


thomaspink:
Thank you for this fix. Looks good to me.

Just on minor thing, please change the type of your commit message from `style` to `fix`. Because the type style is not intended for the styling of a component. It's there for the code style itself. See: https://barista.dynatrace.com/components/contribute#type

> **style:** Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)

tomheller:
Cherry picked to 7.x and 7.4.x

PR #1294, opened 2020-07-15
style(input): Fixes extra outlines when places inside form-field

### LINKED ISSUE #999: [Formfield]: Focus management and outlines are weird
# Bug Report

The focus management of the form field, especially when used with infix and/or suffix inlays looks a little weird:
image

## Expected Behavior

I would expect that the input field within the form field would not get an extra outline around it.

## Current Behavior

See the image above

## Possible Solution

Adjust the focusmonitor if the input is used within a formfield.

## Steps to Reproduce

1. Navigate to https://barista.dynatrace.com/components/form-field
2. Scroll to the Prefix and suffix example
3. Focus the input within the example by clicking into it.

### Used Versions:

- **@dynatrace/barista-components**: 6.5.0

bartoszbobin:

Fixes #999

Before fix:
https://barista.dynatrace.com/components/form-field#prefix-and-suffix

After fix:
barista-form-field-no-nested-outlines

#### Checklist

### DIFF
--- a/apps/dev/src/form-field/form-field-demo.component.html
+++ b/apps/dev/src/form-field/form-field-demo.component.html
@@ -27,3 +27,18 @@
     </dt-select>
   </dt-form-field>
 </div>
+
+<dt-form-field>
+  <dt-label>No nested focus outlines</dt-label>
+  <input
+    type="text"
+    dtInput
+    placeholder="Please insert amount"
+    aria-label="Please insert amount"
+  />
+  <dt-icon dtPrefix name="filter">$</dt-icon>
+  <dt-loading-spinner dtPrefix></dt-loading-spinner>
+  <button dt-icon-button dtSuffix variant="nested" aria-label="Submit changes">
+    <dt-icon name="checkmark"></dt-icon>
+  </button>
+</dt-form-field>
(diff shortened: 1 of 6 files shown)


libs/barista-components/input/src/input.scss
```diff
@@ -56,6 +56,12 @@
   }
 }
 
+.dt-focused {
```
ffriedl89, line 59: This might be a bit dangerous, I'd suggest adding a .dt-form-field selector here as well. 
```suggestion
.dt-form-field.dt-focused {
```

bartoszbobin: Wasn't sure if `dtInput` should be aware of `dt-form-field`.
Maybe this should be placed in `dt-form-field` styles?

What do you think @ffriedl89 ?

ffriedl89: Yeah i agree, it makes sense to include those styles in the form-field.scss instead.

thomaspink:
Wow so many contributions. Holy moly. Looks good to me!

Just on minor thing, please change the type of your commit message from `style` to `fix`. 
Because the type style is not intended for the styling of a component. It's there for the code style itself. For reference see: https://barista.dynatrace.com/components/contribute#type

> **style:** Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)

ffriedl89:
cherry-pick done

Issue #866, opened 2020-04-13
Make select placeholder more obviously a placeholder


tcarrio:
# Feature Request

Stylized placeholder in DtSelect component

## Summary

The Barista select component currently allows for a placeholder when an option has not been selected. However, the styling of the placeholder is the same as the actual options, which can cause confusion for users because there is no visual queue to select an option.

## Feature Description

Problem: Users do not have a visual queue from the placeholder value that an option must be selected.

Solution: Provide a styling differentiation between the placeholder value and option values in the select element. As an example, it could be an italicized placeholder where the actual options would be normal.

## Attachments

Current design:

![image](https://user-images.githubusercontent.com/8659099/79147979-cca0aa00-7d92-11ea-8400-bafe67d36291.png)

Proposed Design:
![image](https://user-images.githubusercontent.com/8659099/79148170-138e9f80-7d93-11ea-81f2-e5fa2cc20b36.png)

lukasholzer:
@tcarrio Hey thanks for submitting this issue! Indeed there is a problem with the current UX. But I don't think that the italic fix would solve the problem, as there is no italic in the product as far as I can remember. Maybe we should poke our UX Team about their ideas. 

Maybe we should use a lighter grey but then I'm not sure that the colour contrast meets the a11y standards.

@dynatrace-oss/ux-core any input on this topic would be appreciated!

ursula-wieshofer:
Hi @tcarrio and @lukasholzer from a UX perspective I think this is a very good and helpful idea. In general I would go with our watermark color as this is just another watermark.
Now every watermark in our components (Select, input field,...) is #898989 = Gray500, however this color unfortunately does not yet follow [our a11y guidelines](https://barista.lab.dynatrace.org/guidelines/accessibility#distinguishable), but it would be more consistent.

So imho we should
- Change it to our current watermark color, which would solve your problem, but not the a11y standards
- Change the watermark color on all our components to follow the a11y guidelines

SkaillZ:
@tcarrio We would suggest going for the second option, since we already have an issue for that: https://github.com/dynatrace-oss/barista/issues/880. Could you align with @beeme1mr and consider helping out or contributing this yourself?

bartoszbobin:
@ursula-wieshofer what about disabled input, select with placeholder?

For disabled input with text we use gray-300

Here is a compariston for disabled input with placeholder gray-300 vs gray-500
[...]

PR #1295, opened 2020-07-15
style: Unifying placeholder color for input, select and combobox

bartoszbobin:

Before:
image

After:
image

#### Checklist

### DIFF
--- a/libs/barista-components/experimental/combobox/src/combobox.scss
+++ b/libs/barista-components/experimental/combobox/src/combobox.scss
@@ -41,7 +41,7 @@
 
   .dt-combobox-input,
   .dt-combobox-input::placeholder {
-    color: $disabledcolor;
+    @include dt-form-control-disabled-placeholder();
   }
 }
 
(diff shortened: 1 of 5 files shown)


libs/barista-components/experimental/combobox/src/combobox.scss
```diff
@@ -41,7 +41,7 @@
 
   .dt-combobox-input,
   .dt-combobox-input::placeholder {
-    color: $disabledcolor;
+    @include dt-form-control-disabled-placeholder();
```
tomheller, line 44: I'm not entirely sure if we really need a mixin for this, but it's fine.

bartoszbobin: At this moment we have at least 3 components which used it:
- input
- select
- combobox

so, decided to introduce 2 mixins, and according to https://github.com/dynatrace-oss/barista/issues/866#issuecomment-616968503

> Now every watermark in our components (Select, input field,...) is #898989 = Gray500, **however this color unfortunately does not yet follow our a11y guidelines**, but it would be more consistent.

I think that would be nice to change it in one place in the future ;)

tomheller:
Cherry picked to 7.x and 7.4.x

PR #1298, opened 2020-07-16
ds(sidenav): Added sidenav.

NinaKammerlander:

This PR adds a sidenav for the design system app. The sidenav is generated according to the data from the cms.
<hr>


apps/next/src/app/pages/single-page/components/sidenav/sidenav.ts
```diff
@@ -24,7 +24,11 @@ import { Component, Input } from '@angular/core';
...
 })
-export class BaSidenav {
+export class DsSidenav {
   /** all sidenav items */
-  @Input() sidenavData;
+  @Input()
+  set sidenavData(sidenavData) {
+    this._sections = sidenavData.sections;
+  }
+  _sections;
```
tomheller, line 33: And please type it.
```suggestion
  private: _sections;
```

apps/next/src/app/pages/single-page/components/sidenav/sidenav.ts
```diff
@@ -24,7 +24,11 @@ import { Component, Input } from '@angular/core';
     class: 'next-sidenav',
   },
 })
-export class BaSidenav {
+export class DsSidenav {
   /** all sidenav items */
-  @Input() sidenavData;
+  @Input()
```
tomheller, line 29: Please add a getter to the input as well.

libs/tools/barista/src/generators/ds-navigation.ts
```diff
@@ -0,0 +1,108 @@
...
+      sections: [],
+    };
+
+    for (const file of files) {
+      const content = JSON.parse(readFileSync(join(path, file)).toString());
+
+      sideNav = addNavSection(sideNav, content.navGroup);
+
+      for (const section of sideNav.sections) {
+        const filepath = join(directory, file.replace(/\.[^/.]+$/, ''));
```
ffriedl89, line 92: I am not sure what is going on here - i am really bad at reading regex. Could you please explain why you added this here? :)

libs/tools/barista/src/generators/ds-navigation.ts
```diff
@@ -0,0 +1,108 @@
...
+    let sideNav: DsSideNavContent = {
+      sections: [],
+    };
+
+    for (const file of files) {
+      const content = JSON.parse(readFileSync(join(path, file)).toString());
+
+      sideNav = addNavSection(sideNav, content.navGroup);
+
+      for (const section of sideNav.sections) {
```
ffriedl89, line 91: You are looping over the sections twice now - once in the addNavSection and once in this loop. We can combine those 2 loops into one and fill up the items right away.

libs/tools/barista/src/generators/ds-navigation.ts
NinaKammerlander: This is used to remove the file extension. The variable is used as a link to the page so the link would be e.g. components/button while the file name is button.json

ffriedl89: I'd rather use https://nodejs.org/api/path.html#path_path_basename_path_ext and dirname to extract the filename and the path. This is way safer and battle tested :)

libs/tools/barista/src/generators/ds-navigation.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+}
+
+/** Builds navigation */
+export const navigationBuilder = async () => {
+  const allDirectories = readdirSync(distDir).filter((dirPath) =>
+    isDirectory(dirPath),
+  );
+
+  const pages = allDirectories.map(async (directory) => {
```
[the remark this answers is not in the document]
tomheller: Sorry @NinaKammerlander, but I missed that in our liveShare. Could we please change the loop here to sth like this. Async functions in maps are not the best the most senisble approach: 
```ts
for (const directory of allDirectories) {
// ...
}
```

No need to return `Promise.all` at the end then.

libs/tools/barista/src/generators/ds-navigation.ts
```diff
@@ -0,0 +1,97 @@
...
+
+/** Builds navigation */
+export const navigationBuilder = async () => {
+  const allDirectories = readdirSync(distDir).filter((dirPath) =>
+    isDirectory(dirPath),
+  );
+
+  for (const directory of allDirectories) {
+    const currentPath = join(distDir, directory);
+    const files = readdirSync(currentPath);
```
lukasholzer, line 60: NIT: wont detect nested files in folders you can use `sync('**/*.extension' )` imported from `import { sync } from 'glob';`

libs/tools/barista/src/generators/ds-navigation.ts
```diff
@@ -0,0 +1,97 @@
...
+    const fileMap = new Map<string, any>();
+
+    let sideNav: DsSideNavContent = {
+      sections: [],
+    };
+
+    for (const file of files) {
+      const fileContent = await fs.readFile(join(currentPath, file), { encoding: 'utf-8' });
```
lukasholzer, line 68: in this case the sync version would be way faster then the async version if you want to perform it async then do something like

```typescript
const fileContents = await Promise.all(files.map(file => fs.readFile(join(currentPath, file), { encoding: 'utf-8' }));
fileContents.forEach(fileContent => {
  // do the stuff
})
```

libs/tools/barista/src/generators/ds-navigation.ts
```diff
@@ -0,0 +1,97 @@
...
+            });
+        }
+      }
+    }
+
+    for (const file of files) {
+      const content = fileMap.get(file);
+      // add sidenav to the json file
+      content.sidenav = sideNav;
+      await fs.writeFile(join(currentPath, file), JSON.stringify(content, null, 2), {
```
lukasholzer, line 92: same here create an array of promises and await it in a `Promise.all` otherwise, if you only perform a single operation the sync version is way faster

lukasholzer:
formatting has to be done `npm run format:write` should do the trick!

PR #1301, opened 2020-07-16
feat(formatters): Added maxPrecision parameter to count formatters to…

dominikmessner:
… control the amount of decimals places

<hr>

I need to configure the maximum amount of digits for the count formatter

The percent formatter can already be configured like that using the maxPrecision argument: https://github.com/dynatrace-oss/barista/blob/b2f11f8db6569c20c81f55ba72d9ae78a058c817/libs/barista-components/formatters/src/percent/percent.ts#L31

I would like to add the same argument to the count formatter.

#### Type of PR

#### Checklist

### DIFF
--- a/libs/barista-components/formatters/src/number-formatter.ts
+++ b/libs/barista-components/formatters/src/number-formatter.ts
@@ -60,7 +60,7 @@ function adjustPrecision(value: number, maxPrecision?: number): string {
   const calcValue = Math.abs(value);
   const minValue =
     1 /
-    Math.pow(10, Math.max(maxPrecision || DEFAULT_PRECISION_FOR_MIN_VALUE, 0));
+    Math.pow(10, Math.max(maxPrecision ?? DEFAULT_PRECISION_FOR_MIN_VALUE, 0));
 
   let digits = 0;
   if (calcValue === 0) {
@@ -71,7 +71,7 @@ function adjustPrecision(value: number, maxPrecision?: number): string {
     } else {
       return `< ${minValue}`;
     }
-  } else if (maxPrecision) {
+  } else if (maxPrecision !== undefined) {
     digits = maxPrecision;
   } else if (calcValue < 1) {
     digits = 3;
--- a/libs/barista-components/formatters/README.md
+++ b/libs/barista-components/formatters/README.md
@@ -69,19 +69,21 @@ fields, which are described in a table below:
 The `formatCount` function provides a way to format numbers as abbreviations
 outside the template. The function takes the following parameters:
 
-| Name        | Type                        | Default      | Description                                                                                                                     |
-| ----------- | --------------------------- | ------------ | ------------------------------------------------------------------------------------------------------------------------------- |
-| `input`     | `DtFormattedValue | number` |              | numeric value to be transformed by the pipe                                                                                     |
-| `inputUnit` | `DtUnit | string`           | `Unit.COUNT` | input unit, if not default - displayed together with the formatted value; does not yet support plurals and internationalization |
+| Name           | Type                        | Default      | Description                                                                                                                     |
+| -------------- | --------------------------- | ------------ | ------------------------------------------------------------------------------------------------------------------------------- |
+| `input`        | `DtFormattedValue | number` |              | numeric value to be transformed by the pipe                                                                                     |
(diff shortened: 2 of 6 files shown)


dominikmessner:
I saw that there are unit tests suggesting that the formatter would support negative values for the precision: https://github.com/dynatrace-oss/barista/blob/e62a79ac59a7fc480cf643783c03b0184b405504/libs/barista-components/formatters/src/number-formatter.spec.ts#L188
This is actually not the case, it will throw errors with most values. Not sure if this was a requirement, or if negative precision values should be caught and cause an exception.

ffriedl89:
> I saw that there are unit tests suggesting that the formatter would support negative values for the precision:
> 
> https://github.com/dynatrace-oss/barista/blob/e62a79ac59a7fc480cf643783c03b0184b405504/libs/barista-components/formatters/src/number-formatter.spec.ts#L188
> 
> 
> This is actually not the case, it will throw errors with most values. Not sure if this was a requirement, or if negative precision values should be caught and cause an exception.

@dominikmessner Do you have example values that currently throw an error? We usually want to avoid throwing errors, but instead handle bad input gracefully, since all those values can change at runtime it might be the case that at some point the function is called with bad input and in that case we don't want to crash the entire screen.

libs/barista-components/formatters/src/number-formatter.ts
```diff
@@ -60,7 +60,7 @@ function adjustPrecision(value: number, maxPrecision?: number): string {
   const calcValue = Math.abs(value);
   const minValue =
     1 /
-    Math.pow(10, Math.max(maxPrecision || DEFAULT_PRECISION_FOR_MIN_VALUE, 0));
+    Math.pow(10, Math.max(maxPrecision ?? DEFAULT_PRECISION_FOR_MIN_VALUE, 0));
```
ffriedl89, line 63: 🤩  i like it

ffriedl89:
cherry picked already

tomheller:
Cherry pick done by @ffriedl89

dominikmessner:
> > I saw that there are unit tests suggesting that the formatter would support negative values for the precision:
> > https://github.com/dynatrace-oss/barista/blob/e62a79ac59a7fc480cf643783c03b0184b405504/libs/barista-components/formatters/src/number-formatter.spec.ts#L188
> > 
> > This is actually not the case, it will throw errors with most values. Not sure if this was a requirement, or if negative precision values should be caught and cause an exception.
> 
> @dominikmessner Do you have example values that currently throw an error? We usually want to avoid throwing errors, but instead handle bad input gracefully, since all those values can change at runtime it might be the case that at some point the function is called with bad input and in that case we don't want to crash the entire screen.

@ffriedl89 
```      
{
  input: 4.123456789,
  maxPrecision: -1,
  output: '< 1',
}
```
This causes an exception in `formatNumber`:
`Error: 0.0--1 is not a valid digit info`
https://angular.io/api/common/DecimalPipe

I think any `input > 1` with a `maxPrecision < 0` will cause an exception in `adjustNumber`.

ffriedl89:
@rowa-audil 

> > > I saw that there are unit tests suggesting that the formatter would support negative values for the precision:
> > > https://github.com/dynatrace-oss/barista/blob/e62a79ac59a7fc480cf643783c03b0184b405504/libs/barista-components/formatters/src/number-formatter.spec.ts#L188
> > > 
> > > This is actually not the case, it will throw errors with most values. Not sure if this was a requirement, or if negative precision values should be caught and cause an exception.
> > 
> > 
> > @dominikmessner Do you have example values that currently throw an error? We usually want to avoid throwing errors, but instead handle bad input gracefully, since all those values can change at runtime it might be the case that at some point the function is called with bad input and in that case we don't want to crash the entire screen.
> 
> @ffriedl89
> 
> ```
> {
>   input: 4.123456789,
>   maxPrecision: -1,
>   output: '< 1',
> }
> ```
> 
> This causes an exception in `formatNumber`:
> `Error: 0.0--1 is not a valid digit info`
> https://angular.io/api/common/DecimalPipe
> 
> I think any `input > 1` with a `maxPrecision < 0` will cause an exception in `adjustNumber`.

@rowa-audil could you check this out please and create an issue to fix this :)

PR #1312, opened 2020-07-17
chore(bazel): Bazelify Alert Component

furtnerl:

<hr>

  often helpful).
  guidelines.

#### Type of PR
<!-- Feature (non-breaking change which adds functionality)

#### Checklist


libs/barista-components/alert/BUILD.bazel
```diff
@@ -1,7 +1,30 @@
...
+    ),
+    angular_assets = [
+        ":styles",
+        ":src/alert.html",
+    ],
+    module_name = "@dynatrace/barista-components/alert",
+    tsconfig = "tsconfig_lib",
+    deps = [
+        "//libs/barista-components/icon:compile",
+        "@npm//@angular/core",
```
ffriedl89, line 24: Do we have to list the common module here as well? 
It's imported inside the alert.module.ts

libs/barista-components/alert/BUILD.bazel
```diff
@@ -16,7 +39,54 @@ filegroup(
     ]),
 )
 
+sass_library(
+    name = "scss_lib",
```
lukasholzer, line 43: I would rename to `theme` as it is more descriptive
```suggestion
    name = "theme",
```

libs/barista-components/alert/BUILD.bazel
```diff
@@ -16,7 +39,54 @@ filegroup(
...
+    srcs = glob(
+        ["src/**/_*.scss"],
+    ),
```
lukasholzer, line 46: I would specify the file explicitly
```suggestion
    srcs = ":_alert-theme.scss"
```

libs/barista-components/alert/BUILD.bazel
furtnerl: good catch
Once we do not have to use the view engine rule we get strict dependency checks where a missing third party dep would break the build. 
To be complete: The most important deps are the ones from within the codebase, because those actually determine what needs to be built. If we get those "relative" deps wrong, the build will fail (also with this version already)

lukasholzer: @furtnerl I created the `ng_module_view_engine` macro exactly with the same API as the `ng_module` (the ts_library with the angular plugin). So to check for the dependencies just switch the `ng_module_view_engine` with the `ng_module` and you have checks. 

After the deps checks passed then revert to the view engine one.

Issue #1314, opened 2020-07-17
Next: Headline transformer


rowa-audil:
# Feature Request

Since not all headlines fit into the document structure (e.g headlines in a card component), we need classes which would be assigned to headline elements because the current styling uses the element selectors.

The **scope of this PR** would be a _transformer_ for next pages which adds the headlines corresponding class to the element and changes the element selectors to classes.

PR #1316, opened 2020-07-20
chore(transformer): Adds transformation step adding classes to headlines

rowa-audil:

Adds a transformer that adds headline classes to headline elements
E.g.: `<h1>headline 1</h1>  =>  <h1 class="h1">headline 1</h1>`

P.S: The transformer is only transforming next pages!

Fixes: #1314 

#### Type of PR
Feature (non-breaking change which adds functionality

#### Checklist

### DIFF
--- a/libs/tools/barista/src/builder/strapi.ts
+++ b/libs/tools/barista/src/builder/strapi.ts
@@ -43,6 +43,7 @@ import {
   copyHeadlineTransformer,
   relativeUrlTransformer,
   tableOfContentGenerator,
+  headlineClassTransformer,
 } from '../transform';
 
 const TRANSFORMERS: BaPageTransformer[] = [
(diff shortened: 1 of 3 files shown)


libs/tools/barista/src/builder/strapi.ts
```diff
@@ -64,6 +65,10 @@ export const strapiBuilder: BaPageBuilder = async (
     return [];
   }
 
+  if (next) {
+    TRANSFORMERS.push(headlineClassTransformer);
+  }
```
tomheller, line 70: We don't necessarily need to do this only for next. We could do this for the current barista outlet as well.

lukasholzer: Agree with tom! :)

rowa-audil: Will there be changes in the styling of barista concerning headlines being styled with classes instead of the current usage of the element selectors?

tomheller: yes, there should be changes to the stylings of headlines in barista. 
We want to decouple the semantics from the styling eventually.

lukasholzer: @tomheller but if we want to decuple styling from semantics we should not name the classes `h1` I would go then with something different that is not that much tied to the semantics

PR #1336, opened 2020-07-20
Adds the value property to the checkboxes and a README to the switch webcomponent

rowa-audil:

The native checkbox support a attribute value.
This PR adds that to both the fluid-checkbox and -switch.
Also adds the missing README to the Switch

#### Type of PR

#### Checklist

### DIFF
--- a/apps/fluid-dev/src/pages/checkbox/checkbox-page.component.html
+++ b/apps/fluid-dev/src/pages/checkbox/checkbox-page.component.html
@@ -3,6 +3,7 @@
   [disabled]="disabled"
   [checked]="_isChecked()"
   [indeterminate]="_isIndeterminate()"
+  value="checkbox"
   (change)="changeAll($event)"
 >
   I am the label
--- a/apps/fluid-dev/src/pages/switch/switch-page.component.html
+++ b/apps/fluid-dev/src/pages/switch/switch-page.component.html
@@ -1,6 +1,7 @@
 <fluid-switch
   [disabled]="disabled"
   [checked]="checked"
+  value="switch"
   (change)="handleChange($event)"
   >Switch</fluid-switch
 >
--- a/libs/fluid-elements/checkbox/src/lib/checkbox.spec.ts
+++ b/libs/fluid-elements/checkbox/src/lib/checkbox.spec.ts
@@ -218,6 +218,27 @@ describe('Fluid checkbox', () => {
     });
   });
 
+  describe('value', () => {
+    it('should have the correct default set', async () => {
+      const shadowedInput = fixture.shadowRoot?.querySelector('input');
+      expect(shadowedInput?.getAttribute('value')).toBe('on');
+    });
+
+    it('should set the value to customvalue when the property is set', async () => {
+      fixture.value = 'customvalue';
+      await tick();
+      const shadowedInput = fixture.shadowRoot?.querySelector('input');
+      expect(shadowedInput?.getAttribute('value')).toBe('customvalue');
+    });
+
+    it('should set the value to customvalue when the attribute is set', async () => {
+      fixture.setAttribute('value', 'customvalue');
+      await tick();
+      const shadowedInput = fixture.shadowRoot?.querySelector('input');
+      expect(shadowedInput?.getAttribute('value')).toBe('customvalue');
+    });
+  });
+
   describe('tabindex', () => {
     it('should have a tabindex on the triggerable element when enabled', () => {
       const checkbox = fixture.shadowRoot?.querySelector('svg');
--- a/libs/fluid-elements/checkbox/src/lib/checkbox.ts
+++ b/libs/fluid-elements/checkbox/src/lib/checkbox.ts
@@ -87,6 +87,11 @@ export class FluidCheckbox extends LitElement {
   })
   disabled = false;
 
+  /** The value attribute of the native input element */
+  @property({ type: String, reflect: true })
+  // Default string for the value is 'on' from MDN (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/checkbox#Additional_attributes)
+  value = 'on';
+
   /**
    * Defines if the checkbox is checked or not.
    * @attr
@@ -193,6 +198,7 @@ export class FluidCheckbox extends LitElement {
           type="checkbox"
           class="fluid-checkbox-input"
           id="${this._unique}"
+          value="${this.value}"
           ?checked="${this.checked}"
           ?disabled="${this.disabled}"
           @change="${this._handleInputChange}"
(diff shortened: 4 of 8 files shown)


libs/fluid-elements/checkbox/src/lib/checkbox.spec.ts
```diff
@@ -218,6 +218,27 @@ describe('Fluid checkbox', () => {
...
+    });
+
+    it('should set the value to customvalue when the property is set', async () => {
+      fixture.value = 'customvalue';
+      await tick();
+      const shadowedInput = fixture.shadowRoot?.querySelector('input');
+      expect(shadowedInput?.getAttribute('value')).toBe('customvalue');
+    });
+
+    it('should set the value to customvalue when the attribute is set', async () => {
```
lukasholzer, line 234: Please add a test case that is setting the value via clicking? Then you cover all the scenarios for how to set the value.

libs/fluid-elements/switch/src/lib/switch.spec.ts
```diff
@@ -150,6 +150,27 @@ describe('Fluid switch', () => {
...
+      expect(shadowedInput?.getAttribute('value')).toBe('on');
+    });
+
+    it('should set the value to customvalue when the property is set', async () => {
+      fixture.value = 'customvalue';
+      await tick();
+      const shadowedInput = fixture.shadowRoot?.querySelector('input');
+      expect(shadowedInput?.getAttribute('value')).toBe('customvalue');
+    });
+
```
lukasholzer, line 165: please add a test with clicking here as well.

libs/fluid-elements/switch/src/lib/switch.spec.ts
```diff
@@ -150,6 +150,27 @@ describe('Fluid switch', () => {
     });
   });
 
+  describe('value', () => {
+    it('should set the default value to on', async () => {
```
lukasholzer, line 154: please add a test scenario where no value is present

libs/fluid-elements/checkbox/src/lib/checkbox.spec.ts
```diff
@@ -218,6 +218,27 @@ describe('Fluid checkbox', () => {
     });
   });
 
+  describe('value', () => {
+    it('should set the default value to on', async () => {
```
lukasholzer, line 222: please add a test scenario where no value is present

apps/fluid-dev/src/pages/checkbox/checkbox-page.component.html
```diff
@@ -3,6 +3,7 @@
   [disabled]="disabled"
   [checked]="_isChecked()"
   [indeterminate]="_isIndeterminate()"
+  [value]="'checkbox'"
```
tomheller, line 6: No need to bind that property, if you don't bind it to a variable.
```suggestion
  value="checkbox"
```

apps/fluid-dev/src/pages/switch/switch-page.component.html
```diff
@@ -1,6 +1,7 @@
 <fluid-switch
   [disabled]="disabled"
   [checked]="checked"
+  [value]="'switch'"
```
tomheller, line 4: No need to bind that property, if you don't bind it to a variable.
```suggestion
  value="switch"
```

libs/fluid-elements/checkbox/src/lib/checkbox.spec.ts
tomheller: Value of a checkbox cannot be set vie clicking. The value does not represent the state of the checkbox (as in checked / unchecked). Example from mdn

> A DOMString representing the value of the checkbox. This is never seen on the client-side, but on the server this is the value given to the data submitted with the checkbox's name. Take the following example:

```html
<form>
  <div>
    <input type="checkbox" id="subscribeNews" name="subscribe" value="newsletter">
    <label for="subscribeNews">Subscribe to newsletter?</label>
  </div>
  <div>
    <button type="submit">Subscribe</button>
  </div>
</form>
```

libs/fluid-elements/checkbox/src/lib/checkbox.ts
```diff
@@ -86,6 +86,10 @@ export class FluidCheckbox extends LitElement {
   })
   disabled = false;
 
+  /** The value attribute of the native input element */
+  @property({ type: String, reflect: false })
```
tomheller, line 90: The value can be reflected, it is also when done natively.
```suggestion
  @property({ type: String, reflect: true })
```

libs/fluid-elements/checkbox/src/lib/checkbox.ts
```diff
@@ -86,6 +86,10 @@ export class FluidCheckbox extends LitElement {
...
+  value = 'on'; // Default string for the value is 'on' from MDN (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/checkbox#Additional_attributes)
```
tomheller, line 91: Nitpick:

```suggestion
    // Default string for the value is 'on' from MDN (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/checkbox#Additional_attributes)
    value = 'on';
```

libs/fluid-elements/checkbox/src/lib/checkbox.spec.ts
tomheller: This is the test where no value is present. The default based on the MdnDocumentation is `on`. 
@rowa-audil maybe reword the tests `it` to `should have the correct default set`.

libs/fluid-elements/switch/src/lib/switch.spec.ts
tomheller: This is the test where no value is present. The default based on the MdnDocumentation is on.
@rowa-audil maybe reword the tests it to should have the correct default set.

libs/fluid-elements/switch/src/lib/switch.spec.ts
tomheller: Same here as on the checkbox, there is no way to set the value by clicking.

libs/fluid-elements/switch/src/lib/switch.ts
```diff
@@ -46,7 +62,11 @@ export class FluidSwitch extends LitElement {
   @property({ type: Boolean, reflect: true })
   disabled = false;
 
-  /** Whether the switch is active or not */
+  /** The value attribute of the native input element */
+  @property({ type: String, reflect: false })
```
tomheller, line 66: Value can be reflected.
```suggestion
  @property({ type: String, reflect: true })
```

PR #1342, opened 2020-07-21
chore: Extend demos sidebar filter to example-names and recovery of filtered key.

tomheller:

- filtered value is saved in session storage, to keep it alive during reloading
- filter now checks example names instead of component group names
- expandable sections are automatically expanded when filtered.

#### Type of PR

Other

#### Checklist

### DIFF
--- a/apps/demos/src/side-nav.component.ts
+++ b/apps/demos/src/side-nav.component.ts
@@ -28,8 +28,12 @@ import { Subscription } from 'rxjs';
 
 interface ComponentItem {
   name: string;
+  expanded?: boolean;
   examples: Array<{ name: string; route: string }>;
 }
+
+/** Key under which the current filter is being persisted into the session storage. */
+const FILTER_SESSION_STORAGE_KEY = 'demos-filter';
 
 @Component({
   selector: 'dt-demos-side-nav',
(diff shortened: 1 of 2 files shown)


apps/demos/src/side-nav.component.ts
```diff
@@ -61,13 +63,23 @@ export class DtDemosSideNav implements AfterContentInit, OnDestroy {
     if (this._componentItemsFilterValue !== filterValue) {
       this._componentItemsFilterValue = filterValue;
       this._updateFilteredComponentItems();
+
+      if (window && window.sessionStorage) {
+        window.sessionStorage.setItem('demos-filter', filterValue);
```
lukasholzer, line 68: NIT: would put the key in a const variable as you use it not only once.

apps/demos/src/side-nav.component.ts
```diff
@@ -106,23 +123,43 @@ export class DtDemosSideNav implements AfterContentInit, OnDestroy {
...
+            };
+          }
+        })
+        .filter(Boolean) as ComponentItem[];
     }
 
     this._changeDetectorRef.markForCheck();
   }
+
+  private _createExampleSuggestionsSource(): void {
```
lukasholzer, line 153: NIT: Maybe add some documentation here

PR #1359, opened 2020-07-23
ci: Switch to large instead of x-large containers

ffriedl89:

Since we are currently consuming more and more credits we need to switch to a smaller container. 

Not 100% certain large is sufficient for build though


lukasholzer:
fix for bazel_build is done in #1360

PR #1363, opened 2020-07-23
fix(schematics): Make example schematics compatible

### LINKED ISSUE #1297: Example Schematics fails on new example generation
# Bug Report

## Expected Behavior

Running the example schematics should generate the new example, and not throw any errors.

## Current Behavior

Trying to run: 
```
ng g ./dist/libs/workspace:dt-example --name=custom-filter-values-parser --component=filter-field
```
Will result in an error: 
```
Cannot read property 'initializer' of undefined
```

## Possible Solution

-Extend the schematics capabilities of handling different syntactic solutions.
-Convert all current examples to be compatible with the schematics.
-After the issue is fixed, provide meaningful error message in case of failure.
- Add a test for graceful error message

bardosmisi:
Make example schematics compatible, and remove obsolete schematics.

#### Type of PR

Fixes #1297


lukasholzer:
@ffriedl89 @tomheller I'm not sure if this is the right way to be honest – we had the problem that it could not add it to the array when it is in an extra array. So the only thing would be checking the source nodes of the last for the reference and add it then there – now we are losing extreme flexibility as you should have both possibilities – adding an example by hand or with the schematics.

For the resolving of the ast thing this would be done in half an hour max I could pair programme with @bardosmisi

tomheller:
> I'm not sure if this is the right way to be honest – we had the problem that it could not add it to the array when it is in an extra array.

I would also go the route of actually manipulating the declarations array and not create an extra assignment declaration. Especially thinking about the shareable example generation code for stackblitz, which might be tripped up by these changes.

PR #1364, opened 2020-07-23
fix(quick-filter): Fixes an issue where an ECAC error was thrown.

### LINKED ISSUE #1305: [Quick Filter] Programatically adding active [filters] causes ExpressionChangedAfterItHasBeenCheckedError and prevents handler from firing
# Bug Report

When using the DtQuickFilter component and passing in an array of filters into the [filters] input, the component does not react as required like the DtFilterField does.

# What should happen

The DtQuickFilter component should react appropriately to filters passed into the [filters] input. Then the (filterChanges) event should be fired.

# What happens currently

The DtQuickFilter component accepts the filters passed into [filters], but does not fire the event handler, and also the console shows a ExpressionChangedAfterItHasBeenCheckedError.

# Demo

https://stackblitz.com/edit/github-mpgcfd?file=src%2Fquick-filter%2Fquick-filter-default-example%2Fquick-filter-default-example.ts

1. Note the "fakeFilters" array that is being passed to the [filters] input
2. Note the console log in the "handle(event)" method that is attached to the (filterChanges) event
3. Check the browser console to see that the event was never fired although fake filters were in fact appended to the filter bar
4. Note the expression has changed error in the browser log

Thank you!
Arnaud

tomheller:

Fixes an expression changed after checked error that was thrown in the filter field.
Thank you @areknow for pointing that out and helping us investigate.

Fixes #1305

#### Type of PR
Bugfix

#### Checklist

### DIFF
--- a/libs/barista-components/experimental/quick-filter/src/quick-filter.ts
+++ b/libs/barista-components/experimental/quick-filter/src/quick-filter.ts
@@ -168,9 +168,18 @@ export class DtQuickFilter<T = any> implements AfterViewInit, OnDestroy {
 
   /** Angular life-cycle hook that will be called after the view is initialized */
   ngAfterViewInit(): void {
-    // When the filters changes apply them to the filter field
-    this._activeFilters$
-      .pipe(takeUntil(this._destroy$))
+    // We need to wait for the first on stable call, otherwise the
+    // underlying filterfield will thow an expression changed after checked
+    // error. Deferring the first filter setting.
+    // Relates to a very weird and hard to reproduce bug described in
+    // https://github.com/dynatrace-oss/barista/issues/1305
+    this._zone.onStable
+      .pipe(
+        take(1),
+        switchMap(() => this._activeFilters$),
+        takeUntil(this._destroy$),
+      )
+      // When the filters changes apply them to the filter field
       .subscribe((filters) => {
         if (this._filterField.filters !== filters) {
           this._filterField.filters = filters;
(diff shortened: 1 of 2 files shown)


tomheller:
Seems there are still test failing. I will figure this out tomorrow.

libs/barista-components/experimental/quick-filter/src/quick-filter.ts
```diff
@@ -168,13 +168,22 @@ export class DtQuickFilter<T = any> implements AfterViewInit, OnDestroy {
...
+      // When the filters changes apply them to the filter field
+      .subscribe(() => {
+        this._activeFilters$
+          .pipe(takeUntil(this._destroy$))
+          .subscribe((filters) => {
+            if (this._filterField.filters !== filters) {
+              this._filterField.filters = filters;
+            }
+          });
       });
```
ffriedl89, line 187:
```suggestion
    this._zone.onStable
      .pipe(
        take(1),
        switchMap(() => this._activeFilters$),
        takeUntil(this._destroy$)      
    ).subscribe((filters) => {
            if (this._filterField.filters !== filters) {
              this._filterField.filters = filters;
            }
          });
      });
```

ffriedl89: This should work out the same and avoids the nested subscriptions. 

https://stackblitz.com/edit/rxjs-ck7pba?devtoolsheight=60

tomheller: To my understanding, this would kill the subscription after the first on stable (because of the take(1)). 

I created the nested subscription, because I only want to subscribe after the first onStable, but want to keep the inner subscription around for the duration of the component lifetime. 

Or am I completely missing how rxjs switchmap and unsubscriptions work?

tomheller: Apparently I did blank on this one. Updated to your suggestion in the code.

PR #1370, opened 2020-07-27
ci(bazel): Changes shell command invocation to fix fork skip check fo…

ffriedl89:
…r bazel setup

This PR should fix the failing shell fork check to skip the bazel setup.

Edit:
~~This is not ideal since the steps are still reported as successful although they should be marked as skipped. Could not find anything in the circleci doc though. @lukasholzer I guess still better than having them fail during setup.~~

Moved the credentials now to the pr branch in the shell script. This works out nicely. Additionally fixed the shell that executes the script to have the brackets working for the fork check.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -157,7 +157,7 @@ commands:
           command: echo "import %workspace%/.circleci/bazel.rc" >> ./.bazelrc
       - run:
           name: Setup Bazel Build Cache
-          command: cd ./tools/bazel_build_cache && sh setup_ci_build_cache.sh
+          command: cd ./tools/bazel_build_cache && ./setup_ci_build_cache.sh
 
   run_bazel_stylelint:
     description: >-
--- a/tools/bazel_build_cache/setup_ci_build_cache.sh
+++ b/tools/bazel_build_cache/setup_ci_build_cache.sh
@@ -26,4 +26,5 @@ openssl aes-256-cbc -salt -a -d\
   -in "${DIR}/cache-write-key.json.enc" \
   -pass "pass:${BAZEL_REMOTE_CACHE_PASSWORD}"
 
+echo "build --google_credentials=./tools/bazel_build_cache/cache-write-key.json" >> ../../.circleci/bazel.rc
 echo "build --remote_upload_local_results=true" >> ../../.circleci/bazel.rc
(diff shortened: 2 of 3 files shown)


ffriedl89:
This is not ideal since the steps are still reported as successful although they should be marked as skipped. Could not find anything in the circleci doc though. @lukasholzer I guess still better than having them fail during setup.

tools/bazel_build_cache/setup_ci_build_cache.sh
```diff
@@ -3,15 +3,6 @@
...
 
-PULL_BRANCH_REGEX="^pull\/[0-9]+$"
-
-# Forked pull requests have CIRCLE_BRANCH set to pull/XXX
-if [[ "${CIRCLE_BRANCH}" =~ $PULL_BRANCH_REGEX ]]; then
-  echo "Forked pull request no setup for upload to remote chache!"
-  echo "build --remote_upload_local_results=false" >> ../../.circleci/bazel.rc
-  exit 0
-fi
-
```
lukasholzer, line 14: Why do we remove this? It should run the bazel targets on forks as well only without the remote build cache population. As for pushing to the remote build cache a salt is needed and we cannot pass environment variables to forks.

.circleci/config.yml
```diff
@@ -507,12 +507,6 @@ workflows:
 
   fork-check:
     jobs:
-      - bazel_stylelint:
-          <<: *filter_forks
-      - bazel_unit_test:
-          <<: *filter_forks
-      - bazel_build:
-          <<: *filter_forks
```
lukasholzer, line 515: We cannot remove these targets on forked pull requests. this would mean that we don't run tests and stylelint on forks

ffriedl89: We can't currently - even if the check above works the bazel script fails afterwards. So in the meantime i'd suggest removing them entirely.

ffriedl89: https://app.circleci.com/pipelines/github/dynatrace-oss/barista/8420/workflows/e751e198-fe99-4ecd-aa61-2962371d7635/jobs/50407 see for the error

ffriedl89: Thx to @furtnerl I had some pointers where to go next - have another look, i think this is more down the road that you had in mind.

ffriedl89:
cherry pick done!

Issue #1319, opened 2020-07-20
[tab]: Focus bug on click, Tabindex on wrong tab when programatically removing currently focused tab item


rowa-audil:
# Bug Report

Found some bugs in the tab component.

## Expected Behavior

- Focus style should only apply when the user tabbed to the tab-list/a tab item
- The focusable element should always be an existing tab item

## Current Behavior

- The tab gets the focus outline when clicked.
- The focus stays on the tab which is removed programatically

## Steps to Reproduce

1. Serve the fluid dev
2. Click on a tab item. Results in it getting the focus outline
3. Use the key Tab to navigate to the last tab (saved queries)
4. Click the toggle last tab button
5. Tabbing to a tab item is not possible since the tabindex is '0' on the removed item

## Detailed Description

Set the focus outline only when navigated to using tab. Will remove the code setting the focus on a tab item on click.
Check on removal if a existing tab has a tab index 0. If that is not the case then set the tabindex of the first tab that is not disabled to 0.

PR #1371, opened 2020-07-27
fix(tab): Focus bug fixes

rowa-audil:

Fixes a few bugs for the focus styling of the tabs. Specifically when clicking or removing a focused tab.
fixes: #1319 

#### Type of PR

#### Checklist

### DIFF
--- a/libs/fluid-elements/tab-group/src/lib/tab-events.ts
+++ b/libs/fluid-elements/tab-group/src/lib/tab-events.ts
@@ -16,21 +16,35 @@
 
 /** Custom event implementation fires when the active tab has changes */
 export class FluidTabGroupActiveTabChanged extends CustomEvent<any> {
-  constructor(public activeTab: string) {
+  constructor(public activeTabId: string) {
     super('activeTabChanged', { bubbles: true, composed: true });
   }
 }
 
 /** Custom event implementation that fires when a tab is clicked providing the active tab id  */
 export class FluidTabActivatedEvent extends CustomEvent<any> {
-  constructor(public activeTab: string) {
+  constructor(public tabId: string) {
     super('tabActivated', { bubbles: true, composed: true });
+  }
+}
+
+/** Custom event implementation that fires when the active attribute was set */
+export class FluidTabActiveSetEvent extends CustomEvent<any> {
+  constructor(public tabId: string) {
+    super('activeSet', { bubbles: true, composed: true });
   }
 }
 
 /** Custom event implementation that fires when a tab is disabled providing the disabled tab id  */
 export class FluidTabDisabledEvent extends CustomEvent<any> {
-  constructor(public disableTab: string) {
+  constructor(public tabId: string) {
     super('disabled', { bubbles: true, composed: true });
+  }
+}
+
+/** Custom event implementation that fires when a tab is blurred */
+export class FluidTabBlurredEvent extends CustomEvent<any> {
+  constructor(public tabId: string) {
+    super('blur', { bubbles: true, composed: true });
   }
 }
--- a/libs/fluid-elements/tab-group/src/lib/tab-group/tab-group.spec.ts
+++ b/libs/fluid-elements/tab-group/src/lib/tab-group/tab-group.spec.ts
@@ -15,7 +15,10 @@
  */
 
 import { FluidTabGroup } from './tab-group';
-import { dispatchKeyboardEvent } from '@dynatrace/testing/browser';
+import {
+  dispatchKeyboardEvent,
+  dispatchFakeEvent,
+} from '@dynatrace/testing/browser';
 import { ARROW_RIGHT, SPACE } from '@dynatrace/shared/keycodes';
 import { FluidTab } from '../tab/tab';
 
@@ -26,9 +29,22 @@ function tick(): Promise<void> {
 describe('Fluid tab group', () => {
   let fixture: FluidTabGroup;
   let activeTabChangedSpy: jest.Mock;
-  let keyDownSpy: jest.Mock;
+  let keyupSpy: jest.Mock;
+  let blurSpy: jest.Mock;
 
-  //TODO: Helper functions for getting the correct element to click/keydown
+  /** Get the first tab in the fluid-tab-group */
+  function getFirstSpanElementFromFluidTab(): HTMLSpanElement {
+    return fixture
+      .querySelector('fluid-tab')
+      ?.shadowRoot?.querySelector('span')!;
+  }
+
+  /** Get the last tab in the fluid-tab-group */
+  function getLastSpanElementFromFluidTab(): HTMLSpanElement {
+    return fixture
+      .querySelector('fluid-tab:last-child')
+      ?.shadowRoot?.querySelector('span')!;
+  }
 
   beforeEach(() => {
     // Register the element, if it is not yet registed
@@ -55,8 +71,11 @@ describe('Fluid tab group', () => {
     activeTabChangedSpy = jest.fn();
     fixture.addEventListener('activeTabChanged', activeTabChangedSpy);
 
-    keyDownSpy = jest.fn();
-    fixture.addEventListener('keydown', keyDownSpy);
+    keyupSpy = jest.fn();
+    fixture.addEventListener('keyup', keyupSpy);
+
+    blurSpy = jest.fn();
+    fixture.addEventListener('blur', blurSpy);
   });
 
   afterEach(() => {
@@ -69,66 +88,93 @@ describe('Fluid tab group', () => {
 
   describe('activetabid attribute', () => {
     it('should set the initial active tab', async () => {
-      expect(fixture.getAttribute('activetabid')).toBe('section1'); // Passes on its own
+      expect(fixture.getAttribute('activeTabId')).toBe('section1');
     });
 
     it('should set active tab when property is set', async () => {
-      fixture.activetabid = 'section2';
+      fixture.activeTabId = 'section2';
       await tick();
-      expect(fixture.getAttribute('activetabid')).toBe('section2');
+      expect(fixture.getAttribute('activeTabId')).toBe('section2');
     });
 
     it('should set last activetabid attribute when a tab is clicked', async () => {
-      const tab = fixture
-        .querySelector('fluid-tab:last-child')
-        ?.shadowRoot?.querySelector('span');
+      const tab = getLastSpanElementFromFluidTab();
       tab?.click();
       await tick();
-      expect(fixture.getAttribute('activetabid')).toBe('section2');
+      expect(fixture.getAttribute('activeTabId')).toBe('section2');
     });
 
     it('should set last activetabid attribute when using key events', async () => {
       const tab = fixture.querySelector<FluidTab>('fluid-tab');
       tab?.focus();
       await tick();
-      expect(fixture.getAttribute('activetabid')).toBe('section1');
-      dispatchKeyboardEvent(tab!, 'keydown', ARROW_RIGHT);
+      expect(fixture.getAttribute('activeTabId')).toBe('section1');
+      dispatchKeyboardEvent(tab!, 'keyup', ARROW_RIGHT);
       await tick();
-      expect(keyDownSpy).toBeCalledTimes(1);
-      dispatchKeyboardEvent(document.activeElement!, 'keydown', SPACE);
+      expect(keyupSpy).toBeCalledTimes(1);
+      dispatchKeyboardEvent(document.activeElement!, 'keyup', SPACE);
+      await tick();
       expect(activeTabChangedSpy).toHaveBeenCalledTimes(1);
-      expect(fixture.getAttribute('activetabid')).toBe('section2');
+      expect(fixture.getAttribute('activeTabId')).toBe('section2');
+    });
+
+    it('should represent the correct activetabid if the tab itself is set to active', async () => {
+      await tick();
+      const tab = fixture.querySelector<FluidTab>('fluid-tab:last-child');
+      tab!.active = true;
+
+      await tick();
+      expect(tab?.active).toBeTruthy();
+      expect(fixture.getAttribute('activeTabId')).toBe('section2');
     });
   });
 
   describe('tabindex attribute', () => {
-    // Test if tabindex is set to 0 when clicked or keyed to a tab
     it('should set tabindex to 0 when tab is clicked', async () => {
+      const tab = getFirstSpanElementFromFluidTab();
+      tab?.click();
       await tick();
-      expect(
-        fixture.querySelector('fluid-tab:last-child')?.getAttribute('tabindex'),
-      ).toBe('-1');
-      fixture
-        .querySelector('fluid-tab:last-child')
-        ?.shadowRoot?.querySelector('span')
-        ?.click();
+      fixture.querySelector<FluidTab>('fluid-tab:last-child')!.click();
+      getLastSpanElementFromFluidTab().click();
       await tick();
-      expect(
-        fixture.querySelector('fluid-tab:last-child')?.getAttribute('tabindex'),
-      ).toBe('0');
+      expect(getFirstSpanElementFromFluidTab().getAttribute('tabindex')).toBe(
+        '-1',
+      );
+      expect(getLastSpanElementFromFluidTab().getAttribute('tabindex')).toBe(
+        '0',
+      );
     });
 
-    // Todo: Find out how to dispatch a keyboard event
-    it('should set tabindex to 0 when tab is navigated to using arrow keys', async () => {});
-    // Test if tabindex is set to -1 when clicked or keyed to a tab
-    // Should have tabindex -1 when element is disabled
+    it('should set tabindex to 0 when tab is selected using keys', async () => {
+      const tab = fixture.querySelector<FluidTab>('fluid-tab');
+      tab?.focus();
+      await tick();
+      dispatchKeyboardEvent(tab!, 'keyup', ARROW_RIGHT);
+      await tick();
+      dispatchKeyboardEvent(document.activeElement!, 'keyup', SPACE);
+      await tick();
+      expect(getFirstSpanElementFromFluidTab().getAttribute('tabindex')).toBe(
+        '-1',
+      );
+      expect(getLastSpanElementFromFluidTab().getAttribute('tabindex')).toBe(
+        '0',
+      );
+    });
+
+    it('should set tabindex to -1 when tab is disabled', async () => {
+      fixture
+        .querySelector<FluidTab>('fluid-tab')
+        ?.setAttribute('disabled', 'true');
+      await tick();
+      expect(getFirstSpanElementFromFluidTab().getAttribute('tabindex')).toBe(
+        '-1',
+      );
+    });
   });
 
   describe('activeTabChanged event', () => {
     it('should fire an event when a tab is clicked', async () => {
-      const tab = fixture
-        .querySelector('fluid-tab[tabid="section2"]')
-        ?.shadowRoot?.querySelector('span') as HTMLSpanElement;
+      const tab = getLastSpanElementFromFluidTab();
       tab?.click();
       await tick();
       expect(activeTabChangedSpy).toBeCalledTimes(1);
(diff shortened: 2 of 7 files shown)


libs/fluid-elements/tab-group/src/lib/tab-group.ts
```diff
@@ -51,50 +52,64 @@ export class FluidTabGroup extends LitElement {
...
+  private _toBeActiveTab: string;
 
   /** Sets the active tab on click */
   private handleClick(event: FluidTabActivatedEvent): void {
-    if (this.activetabid != event.activeTab) {
-      this.activetabid = event.activeTab;
-      for (const tab of this.tabChildren) {
-        tab.active = tab.tabid === this.activetabid;
+    // Only continue when the clicked tab is different from the currently active tab
+    if (this.activeTabId != event.activeTab) {
```
tomheller, line 68:
```suggestion
    if (this.activeTabId !== event.activeTab) {
```

libs/fluid-elements/tab-group/src/lib/tab-group.ts
```diff
@@ -51,50 +52,64 @@ export class FluidTabGroup extends LitElement {
...
+      // Resets the tabindexes
+      for (const tabItem of this.tabChildren) {
+        tabItem.tabindex = -1;
       }
-      this.dispatchEvent(new FluidTabGroupActiveTabChanged(this.activetabid));
+      this.activeTabId = event.activeTab;
+      // Contains the tab to be active
+      const tab = this.tabChildren.find(
+        (tabItem) => tabItem.tabid === event.activeTab,
+      )!;
```
tomheller, line 77: Check this one first. 
If the activeTabId is set to something that is not here, we need to be able to deal with it. As the activeTabId is a property, it can be set from the outside to any random string. So check first if there is a activateable tab, before resetting all. 

```ts
const toActivateTab = this.tabChildren.find(
  (tabItem) => tabItem.tabid === event.activeTab,
);

if (toActivateTab) {
  
  // Resets the tabindexes
  for (const tabItem of this.tabChildren) {
    tabItem.tabindex = -1;
   }

  tab.active = true;
  tab.tabindex = 0;
  this.dispatchEvent(new FluidTabGroupActiveTabChanged(this.activeTabId));
}

File: libs/fluid-elements/tab-group/src/lib/tab-group.ts [hunk 3]
```diff
@@ -51,50 +52,64 @@ export class FluidTabGroup extends LitElement {
...
   /** Sets the active tab on keydown (ArrowLeft and ArrowRight to select / Enter and Space to confirm) */
-  private handleKeyDown(event: KeyboardEvent): void {
-    // Enter Space controll (validate selection)
+  private handleKeyUp(event: KeyboardEvent): void {
+    // Sets the focus outline when user tabbed into the fluid element
+    if (event.code === TAB) {
+      const tabbed = this.tabChildren.find((tab) => {
+        return tab.tabindex === 0;
+      });
+      tabbed!.tabbed = true;
```
tomheller, line 91: You cannot really rely on the fact, that you will find a currentlyFocussed tab. Please add a check if you actually found one.

libs/fluid-elements/tab-group/src/lib/tab-group.ts
```diff
@@ -51,50 +52,64 @@ export class FluidTabGroup extends LitElement {
...
+      tab.tabindex = 0;
+      this.dispatchEvent(new FluidTabGroupActiveTabChanged(this.activeTabId));
     }
   }
 
   /** Sets the active tab on keydown (ArrowLeft and ArrowRight to select / Enter and Space to confirm) */
-  private handleKeyDown(event: KeyboardEvent): void {
-    // Enter Space controll (validate selection)
+  private handleKeyUp(event: KeyboardEvent): void {
+    // Sets the focus outline when user tabbed into the fluid element
```
tomheller, line 86:
```suggestion
    // Sets the focus outline when user tabbed into the tab group
```

libs/fluid-elements/tab-group/src/lib/tab/tab.scss
```diff
@@ -56,9 +56,13 @@
...
     content: '';
   }
+
+  &:not(.fluid-tabbed):focus {
+    outline: none;
+  }
 }
 
-.fluid-state--active {
+.fluid-tab-state--active {
```
tomheller, line 65: This was intentionally designed to not contain the element, but be a more generic class name that holds state representation. But there is no guildeline for this one yet, so it's fine.

libs/fluid-elements/tab-group/src/lib/tab-group.ts
```diff
@@ -51,50 +52,64 @@ export class FluidTabGroup extends LitElement {
...
       const activeTab = this.tabChildren.find((tab) => {
-        return this.activetabid === tab.tabid;
+        return this._toBeActiveTab === tab.tabid;
       })!;
 
       activeTab.active = true;
-      this.activetabid = activeTab.tabid;
-      this.dispatchEvent(new FluidTabGroupActiveTabChanged(this.activetabid));
+      this.activeTabId = activeTab.tabid;
+      this._toBeActiveTab = this.activeTabId;
```
tomheller, line 102: This should be already the same value, right?

libs/fluid-elements/tab-group/src/lib/tab-group.ts
```diff
@@ -107,38 +122,58 @@ export class FluidTabGroup extends LitElement {
...
+      // Set the old to normal state
+      this.tabChildren[oldIndex].tabbed = false;
+    }
+  }
+
+  /** In order to prevent the browser to scroll when the user selects a tab using the spacebar we prevent the default behavior. */
+  private handleKeyDown(event: KeyboardEvent): void {
+    if (event.code === SPACE) {
+      event.preventDefault();
     }
```
tomheller, line 138: This is not really a JSDoc annotation about the function, but a code comment that is helpful to the developer of the component to understand why this was necessary. 

JSDoc will often show up for consumers using an IDE (not for private fields obviously), and for them this information would not be very helpful.
```suggestion
  private handleKeyDown(event: KeyboardEvent): void {
     // In order to prevent the browser to scroll when the user selects a tab using the spacebar
     // we prevent the default behavior.
    if (event.code === SPACE) {
      event.preventDefault();
    }
```

libs/fluid-elements/tab-group/src/lib/tab-group.ts
```diff
@@ -107,38 +122,58 @@ export class FluidTabGroup extends LitElement {
...
   private slotchange(): void {
+    // Contains the length of the tabChildren to check if the user has removed tabs
+    let tabChildrenLength;
+    if (this.tabChildren !== undefined) {
+      tabChildrenLength = this.tabChildren.length;
+    }
     this.tabChildren = Array.from(this.querySelectorAll('fluid-tab'));
-    // Initially set the first tab to active
-    if (!this.activetabid && this.tabChildren.length > 0) {
+    // Check if one of the tabs is focusable and if the user removed tabs from the tabs Array if non are focusable
```
tomheller, line 156:
```suggestion
    // Check if one of the tabs is focusable and if the user removed tabs from the tabs array if non are focusable
```

libs/fluid-elements/tab-group/src/lib/tab-group.ts
```diff
@@ -107,38 +122,58 @@ export class FluidTabGroup extends LitElement {
...
+    // If so we set the first available (not disabled) tab active
+    if (
+      !this.tabChildren.find((tab) => tab.tabindex === 0) &&
+      tabChildrenLength > this.tabChildren.length
+    ) {
+      this.setFirstEnabledTabActive();
+    }
```
tomheller, line 163: This feels like something that could be some shareable logic with what happens in L165-L176
Maybe this could be simplified.

libs/fluid-elements/tab-group/src/lib/tab/tab.scss
```diff
@@ -56,9 +56,13 @@
     left: 0;
     content: '';
   }
+
+  &:not(.fluid-tabbed):focus {
```
tomheller, line 60:
```suggestion
  &:not(.fluid-tab-state--tabbed):focus {
```

libs/fluid-elements/tab-group/src/lib/tab/tab.ts
```diff
@@ -78,16 +79,15 @@ export class FluidTab extends LitElement {
...
   get tabindex(): number {
     return this._tabindex;
   }
   set tabindex(value: number) {
     this._tabindex = value;
-    // TODO: Figure out why using the old value does not work. The attribute vanishes in the dom/is not set
     this.requestUpdate('tabindex');
   }
-  private _tabindex = 0;
+  private _tabindex = -1;
```
tomheller, line 90: Hmm, I'm not sure this is a good idea. If the tab would ever be used outside of a tabgroup, the tab is pretty much unusable. 
It would be nice if the tab would not be inaccessible by default. If the tabgroup relies on all tabindexes in the group except for the active one to be set to -1, it should be the one controlling it.

libs/fluid-elements/tab-group/src/lib/tab/tab.ts
```diff
@@ -99,45 +99,77 @@ export class FluidTab extends LitElement {
...
+  get tabbed(): boolean {
+    return this._tabbed;
+  }
+  set tabbed(tabbed: boolean) {
+    const oldTabbed = this.tabbed;
+    this._tabbed = tabbed;
+    this.requestUpdate('tabbed', oldTabbed);
+    this.tabindex = tabbed === true ? 0 : -1;
+  }
+  private _tabbed = false;
```
tomheller, line 121: Nitpick. 
Upon reading this, it was very confusing to have the name of the setter / member be the same as the function parameter. 
Nothing against it, but I think this is easier to grasp.
```suggestion
  set tabbed(value: boolean) {
    const oldTabbed = this.tabbed;
    this._tabbed = value;
    this.requestUpdate('tabbed', oldTabbed);
    this.tabindex = value === true ? 0 : -1;
  }
  private _tabbed = false;
```

libs/fluid-elements/tab-group/src/lib/tab/tab.ts
```diff
@@ -99,45 +99,77 @@ export class FluidTab extends LitElement {
...
   private handleClick(): void {
     if (!this._active) {
       this.dispatchActiveTabEvent();
     }
   }
 
+  /** Called when the tab loses focus. Removes the focus outline and sets the tab up for TAB navigation. */
+  private handleBlur(): void {
+    this.tabbed = false;
+    this.tabindex = 0;
```
tomheller, line 138: The setting of the tabIndex contradicts what happens in the setter of the tabbed. 
The tabbed setter sets the tabindex to -1 if the set value is `false`. 
In line 138 you set it to 0 explicitly. 

I don't think that this is right.

libs/fluid-elements/tab-group/src/lib/tab/tab.ts
```diff
@@ -99,45 +99,77 @@ export class FluidTab extends LitElement {
...
+
   /**
    * Render function of the custom element. It is called when one of the
    * observedProperties (annotated with @property) changes.
    */
   render(): TemplateResult {
     const classes = {
       'fluid-tab': true,
-      'fluid-state--active': this._active,
+      'fluid-tabbed': this.tabbed,
```
tomheller, line 148:
```suggestion
      'fluid-tab-state--tabbed': this.tabbed,
```

libs/fluid-elements/tab-group/src/lib/tab/tab.ts
```diff
@@ -99,45 +99,77 @@ export class FluidTab extends LitElement {
...
+      @blur="${this.handleBlur}"
     >
       <slot></slot>
     </span>`;
   }
+
+  /** return the span element of the template as HTMLSpanElement */
+  private getSpanElement(): HTMLSpanElement {
+    return this.shadowRoot?.querySelector('span')!;
+  }
```
tomheller, line 167:
```suggestion
```

libs/fluid-elements/tab-group/src/lib/tab/tab.ts
```diff
@@ -99,45 +99,77 @@ export class FluidTab extends LitElement {
...
+
+  /** Focuses the span element in the template */
+  focus(): void {
+    this.getSpanElement().focus();
```
tomheller, line 171: You don't need to query this one every time you call focus. Queries on the dom could take a bit longer, and as the span element never changes, you could save it into a member. 

```ts
private _rootElement: HTMLSpanElement;

firstUpdated(props: Map<string | number | symbol, unknown>): void {
  this._rootElement = this.shadowRoot?.querySelector('span');
}
```

```suggestion
    this._rootElement.focus();
```

libs/fluid-elements/tab-group/src/lib/tab-group.ts
rowa-audil: You're right

libs/fluid-elements/tab-group/src/lib/tab/tab.scss
rowa-audil: I'll gladly change it back. When i was reading over it I thought it is more descriptive of what it styles and forgot we talked about these classes to be more generic.

libs/fluid-elements/tab-group/src/lib/tab/tab.ts
rowa-audil: Right

libs/fluid-elements/tab-group/src/lib/tab-group.ts
```diff
@@ -51,50 +52,77 @@ export class FluidTabGroup extends LitElement {
...
-      this.activetabid = event.activeTab;
-      for (const tab of this.tabChildren) {
-        tab.active = tab.tabid === this.activetabid;
+    const toActivateTab = this.tabChildren.find(
+      (tabItem) => tabItem.tabid === event.activeTab,
+    );
+
+    if (toActivateTab) {
+      // Only continue when the clicked tab is different from the currently active tab
+      if (this.activeTabId !== event.activeTab) {
```
tomheller, line 73: The ifs can be merged. 
```suggestion
    // Only continue when the clicked tab is different from the currently active tab
    if (toActivateTab && this.activeTabId !== event.activeTab) {
```

libs/fluid-elements/tab-group/src/lib/tab/tab.spec.ts
```diff
@@ -33,7 +33,7 @@ describe('Fluid tab', () => {
   /** Checks if the current fixture has an active tab */
   function isActive(): boolean {
     return (
-      fixture.shadowRoot?.querySelector('fluid-state--active') !== undefined
+      fixture.shadowRoot?.querySelector('fluid-tab-state--active') !== undefined
```
tomheller, line 36:
```suggestion
      fixture.shadowRoot?.querySelector('fluid-state--active') !== undefined
```

tomheller:
Code looks good now 👍 

Needs a rebase

tomheller:
Sorry about all the conflicts, @rowa-audil Reach out if you need help resolving them. 
They are related to a lot of refactoring done in #1291

rowa-audil:
@tomheller No worries I expected that.

libs/fluid-elements/tab-group/src/lib/tab-group/tab-group.ts
```diff
@@ -135,39 +162,64 @@ export class FluidTabGroup extends LitElement {
...
-      this.tabChildren[0].active = true;
-      this.activetabid = this.tabChildren[0].tabid;
+    // Check if one of the tabs is focusable and if the user removed tabs from the tabs array if non are focusable
+    // If so we set the first available (not disabled) tab active
+    if (
+      !this.tabChildren.find((tab) => tab.tabindex === 0) &&
+      tabChildrenLength > this.tabChildren.length
+    ) {
+      this.setFirstEnabledTabActive();
+    }
```
ffriedl89, line 206: This seems rather complicated. I think we can check whether there is an active tab still in the remaining tabs. If not call setFirstEnabledTabActive().
Am I missing something?

libs/fluid-elements/tab-group/src/lib/tab-group/tab-group.ts
```diff
@@ -135,39 +162,64 @@ export class FluidTabGroup extends LitElement {
...
+
+    const toActivateTab = this.tabChildren.find(
```
ffriedl89, line 208: I am a bit lost - too many checks for that I don't understand. Could you explain what you are doing here.
L208 - L220 is not really clear for me

libs/fluid-elements/tab-group/src/lib/tab-events.ts
```diff
@@ -16,21 +16,27 @@
...
 
 /** Custom event implementation that fires when a tab is disabled providing the disabled tab id  */
 export class FluidTabDisabledEvent extends CustomEvent<any> {
-  constructor(public disableTab: string) {
+  constructor(public tabId: string) {
     super('disabled', { bubbles: true, composed: true });
   }
 }
+
+export class FluidTabBlurredEvent extends CustomEvent<any> {
```
tomheller, line 38: Add jsDoc please

libs/fluid-elements/tab-group/src/lib/tab-group/tab-group.spec.ts
```diff
@@ -26,9 +26,21 @@ function tick(): Promise<void> {
 describe('Fluid tab group', () => {
```
tomheller, line 26: you could add some tests for:
- if the initialization is done correctly, like we have tried in the fluid-dev app.
- if it handles everything correctly if you set `active` on a tab, after it has been initialized
- if it handles everything correctly when tabs are removed / added (slotchange)
- if it deals with it correctly, if there are no longer any tabs in there?
  + Even if the component was interacted with before.

libs/fluid-elements/tab-group/src/lib/tab-group/tab-group.ts
```diff
@@ -136,37 +156,56 @@ export class FluidTabGroup extends LitElement {
...
   private handleDisabled(disableTabEvent: FluidTabDisabledEvent): void {
-    if (this.activetabid === disableTabEvent.disableTab) {
-      const tabToEnable = this.tabChildren.find((tab) => !tab.disabled);
-      if (tabToEnable) {
-        tabToEnable.active = true;
-      }
+    if (this.activeTabId === disableTabEvent.tabId) {
+      this.setFirstEnabledTabActive();
     }
   }
```
tomheller, line 177: Please also add a test scenario for this behavior.

libs/fluid-elements/tab-group/src/lib/tab-group/tab-group.ts
```diff
@@ -136,37 +156,56 @@ export class FluidTabGroup extends LitElement {
...
+    const toDisableTabIndexTab = this.tabChildren.find(
+      (tab) => tab.tabid === event.tabId,
+    );
+    if (toDisableTabIndexTab) {
+      toDisableTabIndexTab.tabindex = -1;
+    }
+    // Sets the active tabs tabindex to 0
+    const toEnableTabIndexTab = this.tabChildren.find((tab) => tab.active);
+    if (toEnableTabIndexTab) {
+      toEnableTabIndexTab.tabindex = 0;
```
tomheller, line 208: Please add tests for this scenario as well.

libs/fluid-elements/tab-group/src/lib/tab-events.ts
```diff
@@ -16,21 +16,27 @@
...
+  constructor(public tabId: string) {
+    super('blurred', { bubbles: true, composed: true });
```
tomheller, line 40: The native event would be called `blur`. 
As long as we can, we should stick to the native namings as close as possible. If it does not work when calling it blur, please use `tabBlurred` (as seen on `tabActivated`).

libs/fluid-elements/tab-group/src/lib/tab/tab.ts
```diff
@@ -168,44 +176,82 @@ export class FluidTab extends LitElement {
...
+  }
+  set tabbed(value: boolean) {
+    const oldTabbed = this.tabbed;
+    this._tabbed = value;
+    this.requestUpdate('tabbed', oldTabbed);
+    this.tabindex = value === true ? 0 : -1;
+  }
+  private _tabbed = false;
+
+  private _rootElement: HTMLSpanElement;
```
tomheller, line 201: jsdoc please.

libs/fluid-elements/tab-group/src/lib/tab/tab.ts
```diff
@@ -148,13 +157,12 @@ export class FluidTab extends LitElement {
...
-  @property({ type: Number, reflect: true })
+  @property({ type: Number, reflect: false })
   get tabindex(): number {
     return this._tabindex;
   }
   set tabindex(value: number) {
     this._tabindex = value;
-    // TODO: Figure out why using the old value does not work. The attribute vanishes in the dom/is not set
     this.requestUpdate('tabindex');
   }
```
tomheller, line 167: I'm curious, do we still need the getter and setter here? this seems like the default behavior.

rowa-audil: We don't need them actually.

libs/fluid-elements/tab-group/src/lib/tab-group/tab-group.ts
rowa-audil: This is not done yet, since the blurring doesn't in a test environment. I need help with this one ^^

lukasholzer:
Please squash your commits

rowa-audil:
@lukasholzer 
I'm going to wait with squashing the commits a until tom has given his review too

lukasholzer:
@rowa-audil Tom is on holiday and will join directly the VCT so he won't have time for that anymore. We don't wait for his review.

rowa-audil:
@lukasholzer Alright. Squashed and rebased :)

lukasholzer (approved):

PR #1373, opened 2020-07-27
feat(show-more): Changed theming to custom properties.

NinaKammerlander:

<hr>
Changed the theming of the show more component to custom properties. Part of #1076 .


libs/barista-components/show-more/src/show-more.scss
```diff
@@ -2,7 +2,6 @@
 @import '../../core/src/style/interactive-common';
 @import '../../core/src/style/variables';
 @import '../../core/src/theming/theming';
```
tomheller, line 4: I don't think we need theming any more.

PR #1385, opened 2020-07-28
fix(theming): Fixes an issue where custom properties were not set to default.

### LINKED ISSUE #1383: [Quick Filter] Radio buttons invisible after upgrade to 7.5.0
# Bug Report

Hi team, 
I noticed that after the 7.5.0 upgrade, the radio buttons in the quick filter are no longer visible. They are however still selectable. I think this is due to the recent change to custom properties styling, but I may be incorrect.

Screen Shot 2020-07-27 at 8 35 07 PM

Thank you!

## Expected Behavior

The radio buttons should still be visible.

## Current Behavior

The radio buttons are not visible.

## Steps to Reproduce

https://stackblitz.com/edit/github-dyfbut

### Used Versions:

- **@dynatrace/barista-components**: 7.5.0

tomheller:

Fixes #1383 

#### Type of PR

Bugfix

#### Checklist

### DIFF
--- a/libs/barista-components/style/themes.scss
+++ b/libs/barista-components/style/themes.scss
@@ -1,5 +1,6 @@
 @import '../core/src/style/variables';
 
+:root,
 .dt-theme-turquoise,
 .dt-theme-blue,
 .dt-theme-royalblue,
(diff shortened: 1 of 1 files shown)


ffriedl89:
As discussed on slack - i think we need to have L63 scoped to the :root selector as well. Since the turquoise one should be the default for the main color :)

tomheller:
@ffriedl89 Added the root selector to the main color theme definitions as well.

PR #1391, opened 2020-07-28
fix(option): Fixes multiple selected bug when using keyboard and mouse

### LINKED ISSUE #1070: [Filter field] First entry always seems to be selected.
# Bug Report

## Expected Behavior

Only the entry where the mouse hovers over should be colored.

## Current Behavior

The entry where the mouse hovers and the first entry are colored

## Possible Solution

Should be fixed, since the user may be confused what he will select when clicking.

## Steps to Reproduce

Just try the first filter field here:
https://barista.dynatrace.com/components/filter-field
* also check autocomplete, select, and combobox

1. Go to https://barista.dynatrace.com/components/filter-field
2. Click on the first filter field
3. Hover over the third entry. You will see that the first entry is also colored as if "selected"

## Context (Environment)
Found on:
Firefox Version: 76.0.1
Chrome Version: 83.0.4103.61

### Used Versions:

- **node**: >=10.13.0
- **yarn**: >=1.13.0 <2
- **typescript**: 3.8.3
- **angular**: 9.1.0
- **@angular/cdk**: 9.2.0
- **@dynatrace/barista-components**: 6.5.2

## Attachments
filter-field-bug

rowa-audil:
@ffriedl89 @thomaspink 

Tom and I have come up with this sollution for the filter field selection "bug" we have.
The DtOption has a subject containing the item that is hovered on mousemove which the autocomplete subscribes to setting the active item.

Any thoughts?

Fixes #1070

### DIFF
--- a/libs/barista-components/autocomplete/src/autocomplete.ts
+++ b/libs/barista-components/autocomplete/src/autocomplete.ts
@@ -36,10 +36,13 @@ import {
   ViewContainerRef,
   ViewEncapsulation,
   OnDestroy,
+  NgZone,
+  Optional,
 } from '@angular/core';
 
 import { DtOption } from '@dynatrace/barista-components/core';
-import { Subscription } from 'rxjs';
+import { Subscription, Subject, merge } from 'rxjs';
+import { switchMap, startWith, takeUntil } from 'rxjs/operators';
 
 let _uniqueIdCounter = 0;
 
@@ -208,18 +211,39 @@ export class DtAutocomplete<T>
   private _projectedOptionsChangeSubscription: Subscription =
     Subscription.EMPTY;
 
+  /** Subject used for unsubscribing */
+  private _destroy$ = new Subject<void>();
+
   constructor(
     private _changeDetectorRef: ChangeDetectorRef,
     private _elementRef: ElementRef<HTMLElement>,
     private _viewContainerRef: ViewContainerRef,
     @Inject(DT_AUTOCOMPLETE_DEFAULT_OPTIONS)
     defaults: DtAutocompleteDefaultOptions,
+    /**
+     * @deprecated
+     * @breaking-change Will be mandatory with verion 9.0.0
+     */
+    @Optional() private _ngZone?: NgZone,
   ) {
     this._autoActiveFirstOption = !!defaults.autoActiveFirstOption;
   }
 
   ngAfterViewInit(): void {
     this._portal = new TemplatePortal(this._template, this._viewContainerRef);
+    this._options.changes
+      .pipe(
+        startWith(null),
+        switchMap(() =>
+          merge(...this._options.map((option) => option._optionHovered)),
+        ),
+        takeUntil(this._destroy$),
+      )
+      .subscribe((option) => {
+        this._ngZone?.run(() => {
+          this._keyManager.setActiveItem(option);
+        });
+      });
   }
 
   ngAfterContentInit(): void {
--- a/libs/barista-components/core/src/option/option.ts
+++ b/libs/barista-components/core/src/option/option.ts
@@ -28,12 +28,14 @@ import {
   Optional,
   Output,
   ViewEncapsulation,
+  NgZone,
 } from '@angular/core';
-import { Subject } from 'rxjs';
+import { Subject, fromEvent } from 'rxjs';
 
 import { _readKeyCode } from '../util/index';
 import { DtOptgroup } from './optgroup';
 import { Highlightable } from '@angular/cdk/a11y';
+import { takeUntil } from 'rxjs/operators';
 
 let _uniqueId = 0;
 
@@ -99,6 +101,9 @@ export class DtOption<T> implements Highlightable, AfterViewChecked, OnDestroy {
    */
   readonly _stateChanges = new Subject<void>();
 
+  /** Subject used for unsubscribing */
+  private _destroy$ = new Subject<void>();
+
   /** The unique ID of the option. */
   get id(): string {
     return this._id;
@@ -126,6 +131,11 @@ export class DtOption<T> implements Highlightable, AfterViewChecked, OnDestroy {
     private _element: ElementRef,
     private _changeDetectorRef: ChangeDetectorRef,
     @Optional() readonly group?: DtOptgroup,
+    /**
+     * @deprecated
+     * @breaking-change Will be mandatory with verion 9.0.
+     */
+    @Optional() private _ngZone?: NgZone,
   ) {}
 
   ngAfterViewChecked(): void {
@@ -142,10 +152,21 @@ export class DtOption<T> implements Highlightable, AfterViewChecked, OnDestroy {
         this._stateChanges.next();
       }
     }
+    if (this._ngZone) {
+      this._ngZone.runOutsideAngular(() => {
+        fromEvent(this._element.nativeElement, 'mouseover')
+          .pipe(takeUntil(this._destroy$))
+          .subscribe(() => {
+            this._handleMouseOver();
+          });
+      });
+    }
   }
 
   ngOnDestroy(): void {
     this._stateChanges.complete();
+    this._destroy$.next();
+    this._destroy$.complete();
   }
 
   /** Selects the option. */
@@ -223,6 +244,16 @@ export class DtOption<T> implements Highlightable, AfterViewChecked, OnDestroy {
     event.stopImmediatePropagation();
 
     this._selectViaInteraction();
+  }
+
+  /** Contains the hovered option as a stream */
+  _optionHovered = new Subject<DtOption<T>>();
+
+  /** @internal Handles mousemove on the option and selects the not active option */
+  _handleMouseOver(): void {
+    if (!this.active) {
+      this._optionHovered.next(this);
+    }
   }
 
   /**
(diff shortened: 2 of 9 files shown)


libs/barista-components/core/src/option/option.ts
```diff
@@ -62,6 +62,7 @@ export class DtOptionSelectionChange<T> {
     '[class.dt-option-disabled]': 'disabled',
     '(click)': '_handleClick($event)',
     '(keydown)': '_handleKeydown($event)',
+    '(mousemove)': '_handleMouseMove()',
```
thomaspink, line 65: Attaching a `mousemove` this way will trigger a CD on every move. Please use `zone.runOutsideAngular` and attach the event handler in that callback (maybe via `fromEvent`). Note, When the correct event has been captured and you emit the hovered output you need to wrap the emitting again in a `zone.run`

libs/barista-components/autocomplete/src/autocomplete.ts
```diff
@@ -220,6 +226,19 @@ export class DtAutocomplete<T>
 
   ngAfterViewInit(): void {
     this._portal = new TemplatePortal(this._template, this._viewContainerRef);
+    this._options.changes
+      .pipe(
+        // startWith(null),
```
tomheller, line 231: Remove commented code.
```suggestion
```

libs/barista-components/experimental/combobox/src/combobox.spec.ts
```diff
@@ -116,10 +116,10 @@ describe('Combobox', () => {
     fixture.detectChanges();
     focusInput(input);
 
-    expect(trigger.classList.contains('dt-combobox-open')).toBeTruthy();
     expect(
       overlayContainerElement.querySelectorAll('dt-option').length,
     ).toEqual(3);
+    expect(trigger.classList.contains('dt-combobox-open')).toBeTruthy();
```
tomheller, line 122: I don't think this change is necessary.

libs/barista-components/experimental/combobox/src/combobox.ts
```diff
@@ -337,6 +337,7 @@ export class DtCombobox<T> extends _DtComboboxMixinBase
       this._templatePortalContent,
       this._viewContainerRef,
     );
+
```
tomheller, line 340: I don't think this change is necessary. 
```suggestion
```

libs/barista-components/select/src/select.ts
```diff
@@ -524,10 +524,18 @@ export class DtSelect<T> extends _DtSelectMixinBase
...
+          merge(...this.options.map((option) => option._optionHovered)),
+        ),
+        takeUntil(this._destroy),
+      )
+      .subscribe((a) => {
+        this._ngZone.run(() => {
+          this._keyManager.setActiveItem(a);
+        });
         this._resetOptions();
-        this._initializeSelection();
```
tomheller, line 530: Why do we remove the initializeSelection here. I guess this should still be the behaviour to set the first item active.

tomheller:
Some e2e tests would be very nice, otherwise this looks fairly good.

libs/barista-components/autocomplete/src/autocomplete.ts
```diff
@@ -220,6 +226,18 @@ export class DtAutocomplete<T>
 
   ngAfterViewInit(): void {
     this._portal = new TemplatePortal(this._template, this._viewContainerRef);
+    this._options.changes
+      .pipe(
+        takeUntil(this._destroy$),
```
lukasholzer, line 231: The `takeUntil` has to be the last operator – when you perform a `switchMap` you switch your stream to a new source and the new source does not get completed with the takeUntil. 
In your code the `takeUntil` only handles the `option.changes` subscription that is triggered by the `switchMap`.

In your case I would go with a simple store the result as subscription and unsubscribe no need for an extra subject.

libs/barista-components/select/src/select.ts
```diff
@@ -942,11 +957,13 @@ export class DtSelect<T> extends _DtSelectMixinBase
   }
 
   private _resetOptions(): void {
+    console.log('');
```
lukasholzer, line 960:
```suggestion
```

libs/barista-components/select/src/select.ts
```diff
@@ -942,11 +957,13 @@ export class DtSelect<T> extends _DtSelectMixinBase
...
     const changedOrDestroyed = merge(this.options.changes, this._destroy);
 
     this.optionSelectionChanges
       .pipe(takeUntil(changedOrDestroyed))
       .subscribe((event) => {
+        console.log(event.isUserInput);
```
lukasholzer, line 966:
```suggestion
```

libs/barista-components/core/src/option/option.ts
```diff
@@ -225,13 +240,24 @@ export class DtOption<T> implements Highlightable, AfterViewChecked, OnDestroy {
...
+  }
+
   /**
    * @internal
    * Selects the option while indicating the selection came from the user. Used to
    * determine if the select's view -> model callback should be invoked.
    */
   _selectViaInteraction(): void {
     if (!this.disabled) {
+      // debugger;
```
lukasholzer, line 260:
```suggestion
```

libs/barista-components/core/src/option/option.ts
```diff
@@ -142,10 +148,19 @@ export class DtOption<T> implements Highlightable, AfterViewChecked, OnDestroy {
         this._stateChanges.next();
       }
     }
+    this._ngZone.runOutsideAngular(() => {
+      fromEvent(this._element.nativeElement, 'mousemove')
```
ffriedl89, line 152: could we use mouseover instead? We only want to know when somebody enters the option and not all consecutive move events as well.

libs/barista-components/core/src/option/option.ts
```diff
@@ -125,6 +130,7 @@ export class DtOption<T> implements Highlightable, AfterViewChecked, OnDestroy {
   constructor(
     private _element: ElementRef,
     private _changeDetectorRef: ChangeDetectorRef,
+    private _ngZone: NgZone,
```
ffriedl89, line 133: Same here regarding the breaking change.

libs/barista-components/autocomplete/src/autocomplete.ts
```diff
@@ -208,10 +210,14 @@ export class DtAutocomplete<T>
...
     Subscription.EMPTY;
 
+  /** Subject used for unsubscribing */
+  private _destroy$ = new Subject<void>();
+
   constructor(
     private _changeDetectorRef: ChangeDetectorRef,
     private _elementRef: ElementRef<HTMLElement>,
     private _viewContainerRef: ViewContainerRef,
+    private _ngZone: NgZone,
```
ffriedl89, line 220: @lukasholzer isn't this considered a breaking change? 
How should we handle this - we can't really go without the zone here - so there is no way around not having this functionality and no zone. One option would be to make the zone optional and not have the functionality when no zone is given. And make it mandatory in the next major version. - Should we add this to the breaking changes for 8? 
What do you think?

lukasholzer: Yea as it comes from the DI it would be ok only if you would extend from the autocomplete then it would break so we have to make it optional and mark it as deprecated to be non optional with version 9 I would say when we release 8 these days. 

I'm really not sure about how we should handle this (Can we mark the component class itself as internal?) so using it via DI is ok but if you want to extend it is on your own risk?

Because to be honest I don't want to support extending from the component.

libs/barista-components/core/src/option/option.ts
lukasholzer: mousemove with takeuntil destroy is really long –> you would listen the whole component lifecycle for mousemoves.

rowa-audil:
I rebased the master into the branch in order for the e2e test to run through. Otherwise it would throw an error because the deprecated empty-state class doesn't have a decorator

libs/barista-components/select/src/select.ts
```diff
@@ -526,8 +526,24 @@ export class DtSelect<T> extends _DtSelectMixinBase
...
+      .pipe(
+        startWith(null),
+        switchMap(() =>
+          merge(...this.options.map((option) => option._optionHovered)),
+        ),
+        takeUntil(this._destroy),
+      )
+      .subscribe((option) => {
+        this._ngZone.run(() => {
+          console.log('set');
```
lukasholzer, line 544: please cleanup
```suggestion
```

libs/barista-components/core/src/option/option.ts
```diff
@@ -225,6 +246,17 @@ export class DtOption<T> implements Highlightable, AfterViewChecked, OnDestroy {
     this._selectViaInteraction();
   }
 
+  /** Contains the hovered option as a stream */
+  _optionHovered = new Subject<DtOption<T>>();
+
+  /** @internal Handles mousemove on the option and selects the not active option */
+  _handleMouseOver(): void {
+    if (!this.active) {
+      console.log('setting');
```
lukasholzer, line 255: cleanup

```suggestion
```

lukasholzer (approved):

heartdisease (approved):

thomaspink (approved):

PR #1416, opened 2020-08-03
fix(chart): Fixes an issue where series visibility was not retained over new data set.

### LINKED ISSUE #1412: Chart filtering resets when new data arrives.
# Bug Report

## Expected Behavior

Column filtering shouldn't be affected by new data.

## Current Behavior

After data changes all columns are visible again.

## Steps to Reproduce
- https://stackblitz.com/edit/github-vilm5e?file=src/chart/chart-stream-example/chart-stream-example.html

### Used Versions:

- **typescript**: 3.8.3
- **@angular/cdk**: 9.2.4
- **@dynatrace/barista-components**: 7.6.0

## Attachments

barista_visability_reseting

tomheller:

When new seriesdata was set into the chart, the visibility settings for each series were not carried
over, as the chart is completely destroyed and reinstantiated every time new data is set.
To mitigate this issue, whenever a new data is set, we now look in the old series dataset for a
series with the same name and carry over the `visible` setting from the previous source.

Fixes #1412

#### Type of PR

Bugfix

#### Checklist

### DIFF
--- a/libs/barista-components/chart/src/chart.ts
+++ b/libs/barista-components/chart/src/chart.ts
@@ -102,7 +102,7 @@ import {
 import { DtChartRange } from './range/range';
 import { DtChartTimestamp } from './timestamp/timestamp';
 import { DtChartTooltip } from './tooltip/chart-tooltip';
-import { getPlotBackgroundInfo } from './utils';
+import { getPlotBackgroundInfo, retainSeriesVisibility } from './utils';
 import { DtChartOptions, DtChartSeries } from './chart.interface';
 
 const HIGHCHARTS_PLOT_BACKGROUND = '.highcharts-plot-background';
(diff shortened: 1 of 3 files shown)


libs/barista-components/chart/src/chart.ts
```diff
@@ -231,11 +231,19 @@ export class DtChart
...
       });
     } else {
-      this._currentSeries = series;
+      if (series) {
+        this._currentSeries = series.map(
+          retainSeriesVisibility(this._chartObject?.series),
+        );
+      } else {
+        this._currentSeries = series;
+      }
```
lukasholzer, line 246: NIT: This would reduce duplication and can eliminate an unnecessary variable declaration.
I thought we have the `prefer-conditional-expression` lint rule in place for that... 🤔

```suggestion
      this._currentSeries = !series
          ? series
          : series.map(retainSeriesVisibility(this._chartObject?.series));
```

PR #1433, opened 2020-08-04
fix(highlight): Fixes an issue where the highlight would break layouts.

### LINKED ISSUE #1420: [Filter field / Autocomplete] Virtualized elements change width of overlay when scrolled into view.
# Bug Report

## Expected Behavior

Filter suggestions panel width should consider max with of their content from the beginning.

## Current Behavior

Right now, the filter suggestions panel width only takes into account the data displayed when expanded, but if some wider content appears on scroll, the width of the panel is adjusted (causing an unsightly "jump" effect).

## Steps to Reproduce

See live code: https://stackblitz.com/edit/github-g3mxzb

1. Click in filter component
2. Select `item` filter key
3. Scroll down suggestions until `some even much longer item` is displayed

tomheller:
When the highlight was wrapped in an option in the filter field and out of view, it was not
initially rendered, but broke the layout of the autocomplete overlay by being too small.

When running and forcing highlight after the init a first time, the value will be mirrored into the
visible span, therefor giving it the size it should eventually have.

Fixes #1420

#### Type of PR

Bugfix

#### Checklist

### DIFF
--- a/libs/barista-components/highlight/src/highlight.ts
+++ b/libs/barista-components/highlight/src/highlight.ts
@@ -30,6 +30,7 @@ import {
   Optional,
   ViewChild,
   ViewEncapsulation,
+  AfterContentInit,
 } from '@angular/core';
 import { Subscription } from 'rxjs';
 import { take } from 'rxjs/operators';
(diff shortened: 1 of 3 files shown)


tomheller:
I understand why the tests on the ci are failing with the results that they are. Interestingly enough I cannot get the tests to fail locally. 
I already tried to clear the jest cache and nx cache to no avail. I also tried to run the tests in a virtual windows machine to eliminate either resource or OS differences. There all tests also passed. This is not making me feel comfortable about our test setup at the moment.
@lukasholzer @ffriedl89

tomheller:
Depends on #1441

lukasholzer:
Rebased with the current master to see if the changes are passing with Angular 10

PR #1446, opened 2020-08-05
fix(filter-field): Fixes an issue with overflowing suggestions.

### LINKED ISSUE #1440: [Filter field] The suggestion list can not display the full text of a long suggestion name
# Bug Report
The suggestion list of the filter component can not handle very long filter values. If the filter value is too long the text is truncated/hidden. The user has no possibility to see the full text of the suggestion.

## Expected Behavior
If the user hover over such an entry, the full suggestion text should be shown in an overlay or something similar. Truncating the text is not an option, because the values would not be unique anymore. 
## Current Behavior

image

## Steps to Reproduce

- insert a very long name in suggestion array in the data source.
- open the suggestion list

## Context (Environment)
- https://stackblitz.com/edit/github-xo8n6f

tomheller:

Fixes #1440

#### Type of PR

Bugfix

#### Checklist

### DIFF
--- a/libs/barista-components/filter-field/src/filter-field.scss
+++ b/libs/barista-components/filter-field/src/filter-field.scss
@@ -178,3 +178,13 @@ button.dt-filter-field-clear-all-button-hidden {
 .dt-filter-field-error p {
   margin: 0;
 }
+
+::ng-deep .dt-filter-field-panel .dt-option {
+  max-width: 100%;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
+::ng-deep .dt-filter-field-panel .dt-highlight {
+  display: inline;
+}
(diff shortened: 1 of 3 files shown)


libs/barista-components/filter-field/src/filter-field.scss
```diff
@@ -178,3 +178,13 @@ button.dt-filter-field-clear-all-button-hidden {
 .dt-filter-field-error p {
   margin: 0;
 }
+
+::ng-deep .dt-filter-field-panel .dt-option {
```
lukasholzer, line 182: NIT: please prefix the `::ng-deep` with a `:host` as you break out of the encapsulation of the component.

tomheller: We need to break encapsulation, as this is styled in the cdk overlay portal, hence the `.dt-filter-field-panel` scope

lukasholzer:
I rebased the branch to see if the checks are passing now with Angular 10 as well.

Issue #1403, opened 2020-07-30
[Barista icons]: Include metadata in published package


tomheller:
To prepare for #1404 and #1405, we should discuss if it is feasible to include the .json files that are holding the metadata of the barista-icons with the released version. 

The information stored in there is already public on the barista outlet pages and we could simplify the build process in the long run, to not rely onto a checked out version of the barista-icons repository.

If more information is needed about this, please reach out.

ffriedl89:
I definitely see this as a big improvement - this switch to have a installed version but check out master only to generate the pages containing the metadata is odd.
I don't see a reason why this metadata could not be included in the bundle. 
So I am 💯  for go!

tomheller:
Solved in deployment strategies for the icons.

PR #1448, opened 2020-08-05
ds: Use installed instead of checked-out icons as icon-page source

### LINKED ISSUE #1404: [Barista tools]: Replace the generation of icon pages source
Currently we rely on the barista-icons sources being checked out to generate the icon pages.
https://github.com/dynatrace-oss/barista/blob/master/libs/tools/barista/src/builder/icons.ts

To simplify this and unify the sourcing of files between barista.dynatrace.com/resources/icons and barista.dynatrace.com/components/icon we should move the generation of the files to source the installed data from @dynatrace/barista-icons npm package.

tomheller:

Fixes #1404

#### Type of PR

Other

#### Checklist

### DIFF
--- a/.deployment/barista/internal/Jenkinsfile
+++ b/.deployment/barista/internal/Jenkinsfile
@@ -10,7 +10,6 @@ pipeline {
   environment {
     WORKSPACE_DIR = "/home/dynatrace"
     BARISTA_PATH = "barista"
-    ICONS_PATH = "barista-icons"
     FRONTEND_IMAGE_NAME = "designops/barista-app-frontend"
     BITBUCKET_URL = credentials("bitbucket-url")
     DOCKER_REGISTRY = credentials("docker-registry")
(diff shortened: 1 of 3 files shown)


lukasholzer:
We need the checked out icons for generating the icon page I thought? That was the reason why we could not use the one from the node_modules

tomheller:
> We need the checked out icons for generating the icon page as we need the matching json file.
That was the reason why we could not use the one from the node_modules – the jsons are only in the repo source folder.

You are correct, but the circumstances have changed. And considering that the json files are now part of the published package since #1403 is resolved, we no longer need to checkout the repo source.

lukasholzer:
> > We need the checked out icons for generating the icon page as we need the matching json file.
> > That was the reason why we could not use the one from the node_modules – the jsons are only in the repo source folder.
> 
> You are correct, but the circumstances have changed. And considering that the json files are now part of the published package since #1403 is resolved, we no longer need to checkout the repo source.

Nice! That simplifies the situation :) Cool that it is possible now 👍  Have you checked the internal build (Jenkins) with this branch?

tomheller:
> Have you checked the internal build (Jenkins) with this branch?

Not yet, but it should be fine as we already source node_modules icons for the icon demo page under the components root page. 

I will start a test-run on jenkins for that branch a little later.

tomheller:
I tried a test run and it failed with the following: 
```
09:35:31  [Error: EISDIR: illegal operation on a directory, read] {
09:35:31    errno: -21,
09:35:31    code: 'EISDIR',
09:35:31    syscall: 'read'
09:35:31  }
```

Probably this is because the node_modules are somehow symlinked. This seems to be more tricky than expected.

lukasholzer:
> I tried a test run and it failed with the following:
> 
> ```
> 09:35:31  [Error: EISDIR: illegal operation on a directory, read] {
> 09:35:31    errno: -21,
> 09:35:31    code: 'EISDIR',
> 09:35:31    syscall: 'read'
> 09:35:31  }
> ```
> 
> Probably this is because the node_modules are somehow symlinked. This seems to be more tricky than expected.

there is the https://nodejs.org/api/cli.html#cli_preserve_symlinks `--preserve-symlinks` flag in nodejs maybe you have to add it to the node script.

lukasholzer:
Pipeline successfully passed and all the icons are the same 

image

SkaillZ (approved):

heartdisease (approved):

PR #1450, opened 2020-08-06
chore: Fixes an issue where the package version was not propagated to the package.

tomheller:

This fixes the packaging until the fluid elements will be bazelified as one of the next steps. 

#### Type of PR

Other 

#### Checklist

### DIFF
--- a/.deployment/publish-next-packages/Jenkinsfile
+++ b/.deployment/publish-next-packages/Jenkinsfile
@@ -12,25 +12,6 @@ pipeline {
 
   stages {
 
-    stage('Checkout specified branch') {
-      steps {
-        checkout([
-          $class: 'GitSCM',
-          branches: [[name: "master"]],
-          doGenerateSubmoduleConfigurations: false,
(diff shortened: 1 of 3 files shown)


.deployment/publish-next-packages/Jenkinsfile
```diff
@@ -12,25 +12,6 @@ pipeline {
 
   stages {
 
-    stage('Checkout specified branch') {
```
lukasholzer, line 15: Is it now declared in the jenkins task? To be honest I like the declarative approach in the Jenkins file.

tomheller: yeah, it was defined in the jenkins task already. we made a second checkout for no reason.

tomheller:
> Can you please add a test scenario for this special case?

I don't really understand which special case you mean. This was already fine, it needed to fail if there is no packageVersion.

Also this will be rewritten again when the components will be bazelified.

PR #1455, opened 2020-08-06
feat(combobox): Introduce injection token to configure option height

heartdisease:
#### Type of PR

Bugfix: Height of options can be configured now using an injection token.


libs/barista-components/autocomplete/src/autocomplete-trigger.ts
```diff
@@ -297,6 +316,16 @@ export class DtAutocompleteTrigger<T>
           }
         });
     }
+
+    if (_optionConfig) {
+      this._optionHeight = _optionConfig.height;
+      this._panelMaxHeight = Math.ceil(
+        _optionConfig.height * OPTION_TO_PANEL_HEIGHT_RATIO,
+      );
```
ffriedl89, line 324: What if the option height is really large - lets say 100px - this is not capped to any value. 
We need to cap this - I'd rather have a logic that tries to fit as many options as possible into the AUTOCOMPLETE_PANEL_MAX_HEIGHT, and in the case that the optionheight is a multiple of the AUTOCOMPLETE_PANEL_MAX_HEIGHT decrease the panel height by half an option height.
This would not let the overlay grow indefinitely and solve the problem that if the height is a multiple of the option height - it looks like all options are shown.

libs/barista-components/experimental/combobox/src/combobox.ts
```diff
@@ -15,56 +15,56 @@
  */
 
 import {
+  AfterContentInit,
   AfterViewInit,
+  Attribute,
   ChangeDetectionStrategy,
```
ffriedl89, line 21: Same here - the file actually has no changes, but due to the imports reordering it pops up in the changeset

apps/dev/src/combobox/combobox-demo.component.scss
```diff
@@ -0,0 +1,11 @@
+.oversized-option {
+  line-height: 40px;
+
+  &:nth-child(n + 1) {
+    background-color: #ff8 !important;
```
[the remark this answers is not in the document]
heartdisease: Aaah interesting, didn't know that!
I usually avoid `!important` whenever possible, but since this is just in the dev app I thought it didn't matter.

libs/barista-components/autocomplete/src/autocomplete-trigger.ts
heartdisease: Good idea, I'll do that!

libs/barista-components/experimental/combobox/src/combobox.ts
heartdisease: I'll fix the import order in a fixup commit.

apps/dev/src/combobox/combobox-demo.component.scss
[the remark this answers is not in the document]
lukasholzer: please remove even in the dev-app as we don't want to have anywhere an example where it could be copied.

apps/dev/src/combobox/combobox-demo.component.scss
```diff
@@ -0,0 +1,11 @@
...
+  }
+
+  &:nth-child(2n) {
+    background-color: #998 !important;
```
lukasholzer, line 9: please remove `!important` here as well.

apps/dev/src/combobox/combobox-demo.component.ts
```diff
@@ -26,43 +26,56 @@ import {
...
   { name: 'Value 4', value: '[value: Value 4]' },
+  { name: 'Value 5', value: '[value: Value 5]' },
+  { name: 'Value 6', value: '[value: Value 6]' },
+  { name: 'Value 7', value: '[value: Value 7]' },
+  { name: 'Value 8', value: '[value: Value 8]' },
+  { name: 'Value 9', value: '[value: Value 9]' },
+  { name: 'Value 10', value: '[value: Value 10]' },
 ];
 
+function optionFilter(filter: string): (o: ComboboxDemoOption) => boolean {
```
lukasholzer, line 49: please provide a small JsDoc that outlines the behaviour of the function (in this case it is a factory as more it is important to describe why) – especially in the dev app or examples it can benefit to understand those function better. 

Please avoid abbrevations as we have this in the [coding guidelines](https://github.com/dynatrace-oss/barista/blob/master/CODING_STANDARDS.md#naming) as it makes hard to understand the code
 ```suggestion
function optionFilter(filter: string): (option: ComboboxDemoOption) => boolean {
```

File: apps/dev/src/combobox/combobox-demo.component.ts [hunk 6]
```diff
@@ -26,43 +26,56 @@ import {
...
+    return option.value.toLowerCase().indexOf(filter.toLowerCase()) >= 0;
+  };
+}
+
 @Component({
   selector: 'combobox-dev-app-demo',
   templateUrl: 'combobox-demo.component.html',
+  styleUrls: ['combobox-demo.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush,
+  providers: [{ provide: DT_OPTION_CONFIG, useValue: { height: 40 } }],
```
lukasholzer, line 60: please put the height of `40` in a constant in the top of the file with a description why exactly 40

`const DT_OPTION_DEFAULT_HEIGHT = 40`

libs/barista-components/autocomplete/src/autocomplete-trigger.ts
```diff
@@ -319,9 +317,10 @@ export class DtAutocompleteTrigger<T>
 
     if (_optionConfig) {
       this._optionHeight = _optionConfig.height;
-      this._panelMaxHeight = Math.ceil(
-        _optionConfig.height * OPTION_TO_PANEL_HEIGHT_RATIO,
-      );
+      this._panelMaxHeight =
+        _optionConfig.height >= AUTOCOMPLETE_PANEL_MAX_HEIGHT
```
ffriedl89, line 321: This checks whether one single option is the same height as the panel max height - please check whether the option height is a multiple of the panel height instead. and then deduct half a option height. 
Lets say you have 3 options that fit exactly into the panelheight - i'd want to have the panel height adjusted to 2.5 option heights in that case

lukasholzer:
Your pipelines cannot run please login to circleci and connect with your github account. as you are a Maintainer this is needed to get the credentials for running the actions. Otherwise you have to create a fork where we run in a workflow without credentials.

image

libs/barista-components/autocomplete/src/autocomplete-trigger.ts
```diff
@@ -270,6 +284,9 @@ export class DtAutocompleteTrigger<T>
     @Optional()
     @Inject(DT_UI_TEST_CONFIG)
     private _config?: DtUiTestConfiguration,
+    @Optional()
+    @Inject(DT_OPTION_CONFIG)
+    _optionConfig?: DtOptionConfiguration,
```
lukasholzer, line 289: NIT: if it is not a property in the class please don't prefix with an underscore (for consistency). We only prefix properties that are internal or private and accessible from outside with an underscore.

```suggestion
    optionConfig?: DtOptionConfiguration,
```

libs/barista-components/autocomplete/src/autocomplete-trigger.ts
```diff
@@ -297,6 +314,22 @@ export class DtAutocompleteTrigger<T>
...
+      const height =
+        _optionConfig.height > AUTOCOMPLETE_OPTION_HEIGHT
+          ? _optionConfig.height
+          : AUTOCOMPLETE_OPTION_HEIGHT;
+
+      this._optionHeight = height;
+      this._panelMaxHeight =
+        AUTOCOMPLETE_PANEL_MAX_HEIGHT % height === 0
+          ? AUTOCOMPLETE_PANEL_MAX_HEIGHT - height / 2
+          : AUTOCOMPLETE_PANEL_MAX_HEIGHT;
```
lukasholzer, line 328: Write some unit tests for it to test the calculation.

NIT: it would be nice and easier to test to put it in a pure function.

libs/barista-components/autocomplete/src/autocomplete-trigger.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+  return {
+    height,
+    maxPanelHeight:
+      AUTOCOMPLETE_PANEL_MAX_HEIGHT % height === 0
+        ? AUTOCOMPLETE_PANEL_MAX_HEIGHT - height / 2
+        : AUTOCOMPLETE_PANEL_MAX_HEIGHT,
+  };
+}
+
+fdescribe.each([
```
lukasholzer, line 46: please remove the `f` in describe

heartdisease: But then I can't use parameterized tests...

lukasholzer (approved):
Thx for adding the test scenario! :)

lukasholzer (approved):

PR #1480, opened 2020-08-12
chore(bazel): Bazelifies the confirmation-dialog

rowa-audil:

Bezelifies the confirmation dialog


libs/barista-components/confirmation-dialog/BUILD.bazel
```diff
@@ -1,7 +1,42 @@
...
+    srcs = glob(
+        include = ["**/*.ts"],
+        exclude = [
+            "**/*.spec.ts",
+            "src/test-setup.ts",
+        ],
+    ),
+    angular_assets = [
+        ":styles",
+        ":src/confirmation-dialog.html",
```
lukasholzer, line 18: you are missing the `confirmation-dialog-state.html` here. 

But what you can try is to bazelify the confirmation dialog state as own dependency.
You can DM me for this :)

libs/barista-components/confirmation-dialog/tsconfig.lib.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "./tsconfig.json",
+  "extends": "../tsconfig.json",
```
lukasholzer, line 2: this was correct please undo

PR #1512, opened 2020-08-20
button group webcomponent

### LINKED ISSUE #1399: Button group as a webcomponent
# Feature Request

## Summary

Create the button group component as a webcomponent. 

## Feature Description

Keep in mind that the button group + items are basically a radio-group + radios. Use the correct roles for the components to be accessible. 

- [x] Create component boilerplate
- [x] Add properties for button-group-items (value, disabled, name, checked)
- [x] Add properties for button-group (value, disabled, name) 
- [x] Style using css proxy custom properties similar to the other web components
- [x] Tests

Think about sharable code between tabs and button-groups - the selection and focus handling using tab and arrow keys is the same. 

image

rowa-audil:

Here's the button-group web component.

fixes #1399 

#### Type of PR

#### Checklist

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -2913,6 +2913,45 @@
         }
       }
     },
+    "fluid-button-group": {
+      "root": "libs/fluid-elements/button-group",
+      "sourceRoot": "libs/fluid-elements/button-group/src",
+      "projectType": "library",
+      "schematics": {},
+      "architect": {
+        "build": {
+          "builder": "./dist/libs/workspace:typescript",
+          "options": {
+            "tsConfig": "libs/fluid-elements/button-group/tsconfig.lib.json",
+            "outDir": "dist/out-tsc"
+          },
+          "configurations": {
+            "production": {}
+          }
+        },
+        "package": {
+          "builder": "./dist/libs/workspace:package-element",
+          "options": {
+            "outputPath": "dist/libs/fluid-elements/button-group",
+            "entryFile": "dist/out-tsc/libs/fluid-elements/button-group/src/index.js",
+            "project": "libs/fluid-elements/button-group/package.json"
+          }
+        },
+        "lint": {
+          "builder": "@angular-devkit/build-angular:tslint",
+          "options": {
+            "tsConfig": [
+              "libs/fluid-elements/button-group/tsconfig.lib.json",
+              "libs/fluid-elements/button-group/tsconfig.spec.json"
+            ],
+            "exclude": [
+              "**/node_modules/**",
+              "!libs/fluid-elements/button-group/**/*"
+            ]
+          }
+        }
+      }
+    },
     "fluid-checkbox": {
       "root": "libs/fluid-elements/checkbox",
       "sourceRoot": "libs/fluid-elements/checkbox/src",
@@ -2955,6 +2994,32 @@
           "options": {
             "jestConfig": "libs/fluid-elements/checkbox/jest.config.js",
             "passWithNoTests": true
+          }
+        }
+      }
+    },
+    "fluid-core": {
+      "root": "libs/fluid-elements/core",
+      "sourceRoot": "libs/fluid-elements/core/src",
+      "projectType": "library",
+      "schematics": {},
+      "architect": {
+        "build": {
+          "builder": "./dist/libs/workspace:typescript",
+          "options": {
+            "tsConfig": "libs/fluid-elements/core/tsconfig.lib.json",
+            "outDir": "dist/out-tsc"
+          },
+          "configurations": {
+            "production": {}
+          }
+        },
+        "package": {
+          "builder": "./dist/libs/workspace:package-element",
+          "options": {
+            "outputPath": "dist/libs/fluid-elements/core",
+            "entryFile": "dist/out-tsc/libs/fluid-elements/core/src/index.js",
+            "project": "libs/fluid-elements/core/package.json"
           }
         }
       }
--- a/libs/fluid-elements/button-group/src/lib/button-group-item/button-group-item.spec.ts
+++ b/libs/fluid-elements/button-group/src/lib/button-group-item/button-group-item.spec.ts
@@ -1,0 +1,276 @@
+/**
+ * @license
+ * Copyright 2020 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { FluidButtonGroupItem } from './button-group-item';
+import { SPACE, ENTER } from '@dynatrace/shared/keycodes';
+import {
+  dispatchKeyboardEvent,
+  dispatchFakeEvent,
+  tick,
+} from '@dynatrace/testing/browser';
+
+/** Checks if the current fixture is displaying the checkmark */
+function showsNob(fixture: FluidButtonGroupItem): boolean {
+  const svg = fixture.shadowRoot?.querySelector('svg');
+  return Boolean(svg?.classList.contains('fluid-state--checked'));
+}
+
+describe('Fluid button-group', () => {
+  let fixture: FluidButtonGroupItem;
+  let changeSpy;
+
+  beforeEach(() => {
+    // Register the element, if it is not yet registered
+    if (!customElements.get('fluid-button-group-item')) {
+      customElements.define('fluid-button-group-item', FluidButtonGroupItem);
+    }
+    // create the fixture
+    document.body.innerHTML =
+      '<fluid-button-group-item>I am the label</fluid-button-group-item>';
+    fixture = document.querySelector<FluidButtonGroupItem>(
+      'fluid-button-group-item',
+    )!;
+
+    // Add spied eventListeners
+    changeSpy = jest.fn();
+    fixture.addEventListener('checkedChange', changeSpy);
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it('should create the button-group-item', () => {
+    expect(fixture).not.toBe(null);
+  });
+
+  describe('Checked Attribute', () => {
+    it('should set the state to checked when the attribute is set', async () => {
+      fixture.setAttribute('checked', '');
+      await tick();
+      expect(fixture.checked).toBeTruthy();
+      expect(fixture.hasAttribute('checked')).toBeTruthy();
+      expect(showsNob(fixture)).toBeTruthy();
+    });
+
+    it('should set the state to checked when the attribute is set true', async () => {
+      fixture.setAttribute('checked', 'true');
+      await tick();
+      expect(fixture.checked).toBeTruthy();
+      expect(fixture.getAttribute('checked')).toBe('true');
+      expect(showsNob(fixture)).toBeTruthy();
+    });
+
+    it('should set the state to checked when the property is set to true', async () => {
+      fixture.checked = true;
+      await tick();
+      expect(fixture.checked).toBeTruthy();
+      expect(fixture.hasAttribute('checked')).toBeTruthy();
+      expect(showsNob(fixture)).toBeTruthy();
+    });
+
+    it('should set the state to unchecked when the attribute is set to true then to false', async () => {
+      fixture.setAttribute('checked', '');
+      await tick();
+      expect(fixture.checked).toBeTruthy();
+      expect(fixture.hasAttribute('checked')).toBeTruthy();
+      expect(showsNob(fixture)).toBeTruthy();
+      fixture.removeAttribute('checked');
+      await tick();
+      expect(fixture.checked).toBeFalsy();
+      expect(fixture.getAttribute('checked')).toBeNull();
+      expect(showsNob(fixture)).toBeFalsy();
+    });
+
+    it('should set the state to unchecked when the property is set to true then to false', async () => {
+      fixture.checked = true;
+      await tick();
+      expect(fixture.checked).toBeTruthy();
+      expect(fixture.hasAttribute('checked')).toBeTruthy();
+      expect(showsNob(fixture)).toBeTruthy();
+      fixture.checked = false;
+      await tick();
+      expect(fixture.checked).toBeFalsy();
+      expect(fixture.getAttribute('checked')).toBeNull();
+      expect(showsNob(fixture)).toBeFalsy();
+    });
+
+    it('should set the state to checked if the label is clicked', async () => {
+      fixture.shadowRoot?.querySelector('label')?.click();
+      await tick();
+      expect(fixture.checked).toBeTruthy();
+      expect(fixture.hasAttribute('checked')).toBeTruthy();
+      expect(showsNob(fixture)).toBeTruthy();
+    });
+
+    it('should set the state to checked when using the SPACE key', async () => {
+      const svg = fixture.shadowRoot?.querySelector('svg');
+      dispatchKeyboardEvent(svg!, 'keyup', SPACE);
+      await tick();
+      expect(fixture.checked).toBeTruthy();
+      expect(fixture.hasAttribute('checked')).toBeTruthy();
+      expect(showsNob(fixture)).toBeTruthy();
+    });
+
+    it('should set the state to checked when using the TAB key', async () => {
+      const svg = fixture.shadowRoot?.querySelector('svg');
+      dispatchKeyboardEvent(svg!, 'keyup', ENTER);
+      await tick();
+      expect(fixture.checked).toBeTruthy();
+      expect(fixture.hasAttribute('checked')).toBeTruthy();
+      expect(showsNob(fixture)).toBeTruthy();
+    });
+  });
+
+  describe('Disabled attribute', () => {
+    it('should set disabled state when the attribute is set', async () => {
+      fixture.setAttribute('disabled', '');
+      await tick();
+      expect(fixture.disabled).toBeTruthy();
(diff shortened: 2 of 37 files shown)


angular.json
```diff
@@ -2925,6 +2925,52 @@
...
+            ]
+          }
+        },
+        "test": {
+          "builder": "@nrwl/jest:jest",
+          "options": {
+            "jestConfig": "libs/fluid-elements/button-group/jest.config.js",
+            "passWithNoTests": true
+          }
+        }
```
lukasholzer, line 2971: please perform linting and testing with bazel and remove it here

libs/fluid-elements/button-group/jest.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
lukasholzer, line 1: Do testing with Bazel and create a `jest.config.json` with following content:

```json
{ 
  "name":"fluid-button-group"
}
```

libs/fluid-elements/button-group/src/lib/button-group-item/button-group-item.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+import { FluidButtonGroupItem } from './button-group-item';
+import { SPACE, ENTER } from '@dynatrace/shared/keycodes';
+import {
+  dispatchKeyboardEvent,
+  dispatchFakeEvent,
+} from '@dynatrace/testing/browser';
+
+function tick(): Promise<void> {
+  return Promise.resolve();
+}
```
lukasholzer, line 26: ca you move this to the testing browser and add some jsdoc to it? What it does and why it is needed?

libs/fluid-elements/button-group/src/lib/button-group-item/button-group-item.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+
+describe('Fluid button-group', () => {
+  let fixture: FluidButtonGroupItem;
+  let changeSpy: jest.Mock;
+
+  /** Checks if the current fixture is displaying the checkmark */
+  function showsNob(): boolean {
+    const svg = fixture.shadowRoot?.querySelector('svg');
+    return Boolean(svg?.classList.contains('fluid-state--checked'));
+  }
```
lukasholzer, line 36: please moove function out of describe closure

libs/fluid-elements/button-group/src/lib/button-group-item/button-group-item.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+    // Add spied eventListeners
+    changeSpy = jest.fn();
+    fixture.addEventListener('checkedChange', changeSpy);
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it('should create the checkbox', async () => {
```
lukasholzer, line 59: why does it have to be async?

```suggestion
  it('should create the checkbox', () => {
```

libs/fluid-elements/button-group/src/lib/button-group-item/button-group-item.ts
```diff
@@ -0,0 +1,393 @@
...
+  static get styles(): CSSResult {
+    return css`
+      :host {
+        /**
+        * Legibility definitions should probably be
+        * shipped or imported from a core
+        */
+        -webkit-font-smoothing: antialiased;
+        -moz-osx-font-smoothing: grayscale;
+        text-rendering: optimizeLegibility;
```
lukasholzer, line 75: why do we specify this on a per component level?

libs/shared/fluid-utils/src/lib/group-navigation.ts
```diff
@@ -0,0 +1,37 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ARROW_LEFT, ARROW_RIGHT } from '@dynatrace/shared/keycodes';
+
+/** Calculate based on a length and the direction (arrow key) what the next index is supporting rotationing */
+export function getNextIndex(
```
lukasholzer, line 20: please provide a more meaningful name that reflects what the function does.

Maybe `getTabGroupNavigationIndex`. (please name the file always in a kebab-cased variant of the method so it is easier to find)

Please update the jsdoc to reflect the problem more took me a while to understand what it is doing

- [ ] Please add a simple unit test for it :) Pure functions are easy to test

libs/fluid-elements/button-group/src/lib/button-group/button-group.ts
```diff
@@ -0,0 +1,221 @@
...
+  private _setFirstTabIndex(): void {
+    for (const item of this._buttonGroupItems) {
+      item.tabIndex = -1;
+    }
+    const toSetTabindexItem = this._buttonGroupItems.find(
+      (item) => !item.disabled,
+    );
+    if (toSetTabindexItem) {
+      toSetTabindexItem.tabIndex = 0;
+    }
```
lukasholzer, line 210: you can combine this in the worst case you are looping through 2 times

libs/fluid-elements/button-group/src/lib/button-group/button-group.ts
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  /** Disables every item in the group */
+  private _setDisabledOnEveryItem(): void {
+    if (this._buttonGroupItems.length > 0) {
+      for (const item of this._buttonGroupItems) {
+        item.disabled = this._disabled;
+      }
+    }
```
lukasholzer, line 219: you don't need the if check because there won't be an iteration if it is an empty array

```suggestion
    for (const item of this._buttonGroupItems) {
      item.disabled = this._disabled;
    }
```

libs/fluid-elements/button-group/src/lib/button-group/button-group.ts
```diff
@@ -0,0 +1,221 @@
...
+      ) {
+        buttonGroupItem.checked = false;
+      }
+    }
+    const checkedButtonGroupItem = this._buttonGroupItems.find(
+      (item) => item.checked,
+    );
+    if (checkedButtonGroupItem) {
+      this.checkedId = checkedButtonGroupItem.id;
+    }
```
lukasholzer, line 174: same here in the worst case you loop 2 times through please combine

libs/fluid-elements/button-group/src/lib/button-group/button-group.ts
```diff
@@ -0,0 +1,221 @@
...
+    const oldValue = this._checkedId;
+    this._checkedId = value;
+    this.requestUpdate('checkedId', oldValue);
+  }
+  private _checkedId: string | null;
+
+  /**
+   * Whether the children are disabled or not.
+   * @attr
+   * @type string
```
lukasholzer, line 89: you specify here string but 2 lines below you have `type: Boolean` 🧐

libs/fluid-elements/button-group/src/lib/button-group/button-group.ts
```diff
@@ -0,0 +1,221 @@
...
+      .fluid-button-group {
+        background-color: var(--fluid-button-group-background-color);
+        width: max-content;
+        padding-left: ${unsafeCSS(FLUID_SPACING_X_SMALL)};
+      }
+    `;
+  }
+
+  /**
+   * Currently checked button-group-item id.
```
lukasholzer, line 71: please explain to me what it does

libs/fluid-elements/button-group/src/lib/button-group/button-group.ts
```diff
@@ -0,0 +1,221 @@
...
+  static get styles(): CSSResult {
+    return css`
+      :host {
+        /**
+        * Legibility definitions should probably be
+        * shipped or imported from a core
+        */
+        -webkit-font-smoothing: antialiased;
+        -moz-osx-font-smoothing: grayscale;
+        text-rendering: optimizeLegibility;
```
lukasholzer, line 57: I don't like to ship this with every component! Furthermore, I want to have CSS processed by an autoprefixer to not have to write code that can be auto-generated.

lukasholzer (changes requested):
Please build, test and lint everything with bazel. (Linting won't be executed on the ci anymore without bazel)

- [ ] remove empty `button-group-page.component.scss`
- [ ] move the `fluid-utils` into `libs/fluid-comopnents/core` - there is a own nx move schematic that should adapt all paths and do this for you I would do it the same like [ing-bank/lion](https://github.com/ing-bank/lion/tree/master/packages/core/src). They have some mixins and util code in there. As we want to have all code that will be shipped for the fluid components inside the fluid components folder (like for angular components)

libs/fluid-elements/button-group/src/lib/button-group-item/button-group-item.ts
```diff
@@ -0,0 +1,393 @@
...
+ * @cssprop --fluid-button-group-item--radio-hover-color - Customize the color of the button-group-item nob when hovering.
+ * @cssprop --fluid-button-group-item--radio-active-color - Customize the color of the button-group-item nob when active.
+ */
+@customElement('fluid-button-group-item')
+export class FluidButtonGroupItem extends LitElement {
+  /**
+   * Unique identifier used for the id and label connection
+   * within the checkbox.
+   */
+  private _unique = `fluid-checkbox-${uniqueCounter++}`;
```
tomheller, line 56:
```suggestion
  private _unique = `fluid-button-group-item-${uniqueCounter++}`;
```

libs/fluid-elements/button-group/src/lib/button-group-item/button-group-item.ts
```diff
@@ -0,0 +1,393 @@
...
+ * to user interaction.
+ * @cssprop --fluid-button-group-item--label-color - Customize the label color.
+ * @cssprop --fluid-button-group-item--radio-hover-color - Customize the color of the button-group-item nob when hovering.
+ * @cssprop --fluid-button-group-item--radio-active-color - Customize the color of the button-group-item nob when active.
+ */
+@customElement('fluid-button-group-item')
+export class FluidButtonGroupItem extends LitElement {
+  /**
+   * Unique identifier used for the id and label connection
+   * within the checkbox.
```
tomheller, line 54:
```suggestion
   * within the button group item.
```

libs/fluid-elements/button-group/src/lib/button-group-item/button-group-item.ts
```diff
@@ -0,0 +1,393 @@
...
+   */
+  private _unique = `fluid-checkbox-${uniqueCounter++}`;
+
+  /** Reference to the input container */
+  private _inputElement: HTMLInputElement;
+
+  constructor() {
+    super();
+  }
```
tomheller, line 63: If you only call `super()` in the constructor, you can omit it.
```suggestion
```

libs/fluid-elements/button-group/src/lib/button-group-item/button-group-item.ts
tomheller: Because at this point there is no core yet (if there ever will be one for css). It's more likely that this will be provided by some sort of wrapper custom-element (similar to the design system provider from fast https://fast.design/docs/components/design-system-provider)

libs/fluid-elements/button-group/src/lib/button-group-item/button-group-item.ts
```diff
@@ -0,0 +1,393 @@
...
+        --fluid-button-group-item--radio-hover-color: var(--color-neutral-100);
+        --fluid-button-group-item--radio-active-color: var(--color-primary-100);
+      }
+
+      :host([disabled]) {
+        pointer-events: none;
+      }
+
+      :host([disabled]) .fluid-label {
+        opacity: 0.5;
```
tomheller, line 89:
```suggestion
        // TODO: Should be replaced by an opacity token at some point.
        opacity: 0.5;
```

libs/fluid-elements/button-group/src/lib/button-group-item/button-group-item.ts
```diff
@@ -0,0 +1,393 @@
...
+      }
+
+      :host([disabled]) .fluid-svg-radio-button {
+        opacity: 0.5;
```
tomheller, line 93:
```suggestion
        // TODO: Should be replaced by an opacity token at some point.
        opacity: 0.5;
```

libs/fluid-elements/button-group/src/lib/button-group-item/button-group-item.ts
```diff
@@ -0,0 +1,393 @@
...
+
+      :host([disabled]) {
+        pointer-events: none;
+      }
+
+      :host([disabled]) .fluid-label {
+        opacity: 0.5;
+      }
+
+      :host([disabled]) .fluid-svg-radio-button {
```
tomheller, line 92: Why is this class called `fluid-svg-radio-button`? This could be fairly confusing, as this is a part of the button-group-item.

libs/shared/fluid-utils/src/lib/group-navigation.ts
peter-affenzeller: it is not only for tab groups, but for grouped items that you can navigate with the arrow keys in general, right?
so maybe something like `getNextGroupItemIndex`?

libs/fluid-elements/button-group/src/lib/button-group-item/button-group-item.ts
```diff
@@ -0,0 +1,393 @@
...
+   */
+  @property({
+    type: String,
+    reflect: true,
+    attribute: 'aria-checked',
+  })
+  ariaChecked: string = 'false';
+
+  /** First updated lifecycle */
+  firstUpdated(props: Map<string | number | symbol, unknown>): void {
```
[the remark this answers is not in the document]
rowa-audil: Thank you!

SkaillZ (dismissed):

lukasholzer (changes requested):
Build and test is failing on the ci

heartdisease (approved):

SkaillZ (approved):

lukasholzer (changes requested):
please remove the test target from the `angular.json` and adapt the jset.config to be a json file (like the fluid button)

rowa-audil:
@lukasholzer I did remove the test target and renamed the `jest.config` of the fluid-button-group. Is it needed for the core and the keycodes aswell?

libs/shared/keycodes/jest.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
```
lukasholzer, line 1: why do you added this it is not needed as there are no tests

libs/fluid-elements/tab-group/src/lib/tab-group/tab-group.ts
```diff
@@ -46,6 +46,8 @@ import {
   FLUID_SPACING_0,
   FLUID_SPACING_MEDIUM,
 } from '@dynatrace/fluid-design-tokens';
+// TODO: Ask Lukaz or someone how to solve this
+import { getNextGroupItemIndex } from '@dynatrace/fluid-elements/core';
```
lukasholzer, line 50: @rowa-audil to solve what? :D

libs/fluid-elements/core/package.json
```diff
@@ -0,0 +1,27 @@
...
+  },
+  "dependencies": {
+    "lit-element": "",
+    "lit-html": ""
+  },
+  "peerDependencies": {
+    "@dynatrace/shared/keycodes": "",
+    "lit-element": "",
+    "lit-html": ""
+  }
```
lukasholzer, line 26: if it is a peer dependency then it is not a dependency.

libs/fluid-elements/button-group/.stylelintrc
```diff
@@ -0,0 +1,6 @@
+{
```
lukasholzer, line 1: this file is not needed as there are no styles.

lukasholzer (changes requested):

libs/fluid-elements/tab-group/src/lib/tab-group/tab-group.ts
rowa-audil: @lukasholzer This is solved actually. I need to get back on looking over the PR's. 😠

rowa-audil: This is already solved 😄

heartdisease (approved):

libs/fluid-elements/core/src/lib/keyboard/get-next-group-item-index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/** Calculate based on a length and the direction (arrow key) what the next index is supporting rotationing */
+export function getNextGroupItemIndex(
+  index: number,
+  length: number,
+  key: string,
+): number {
+  if (key === ARROW_RIGHT) {
+    index += 1;
+  }
```
peter-affenzeller, line 27: I think a switch-case would make sense here, we will probably also need this for the up and down arrow at some point (e.g. navigating through combo-box options)
then we could just do:
```
switch (key) {
  case ARROW_RIGHT:
  case ARROW_DOWN:
    index += 1;
    break;
  case ARROW_LEFT:
  case ARROW_UP:
    index -= 1;
    break;
  default:
    break;
}
```

peter-affenzeller (dismissed):
just one suggestion, overall lgtm 👍

libs/fluid-elements/core/src/lib/keyboard/get-next-group-item-index.ts
rowa-audil: I'd do this in a small follow up #1584

lukasholzer (changes requested):

lukasholzer (dismissed):

lukasholzer:
@rowa-audil please resolve conflicts and squash commits :)

rowa-audil:
@lukasholzer Did.

lukasholzer (approved):

PR #1523, opened 2020-08-25
fix(button): Fixes an issue with secondary theme hover and active state

JanuszLis:
According to the UX guidelines hover/active states should not enforce strict background colors, but just some opacity.

#### Checklist

- [*] I have read the CONTRIBUTING doc and I follow the PR guidelines
- [*] Lint and unit tests pass locally with my changes
- [*] I have added tests that prove my fix is effective or that my feature works
- [*] I have added necessary documentation (if appropriate)

@ffriedl89 @lukasholzer FYI

### DIFF
--- a/libs/barista-components/button/src/_button-theme.scss
+++ b/libs/barista-components/button/src/_button-theme.scss
@@ -87,7 +87,7 @@ a.dt-button-secondary:visited {
   }
 }
 .dt-button-secondary:hover:not([disabled]) {
-  background: $gray-100;
+  background: rgba(255, 255, 255, 0.2);
   border-color: var(--dt-button-hover-color);
   color: var(--dt-button-hover-color);
 
@@ -96,7 +96,7 @@ a.dt-button-secondary:visited {
   }
 }
 .dt-button-secondary:active:not([disabled]) {
-  background: $gray-100;
+  background: rgba(255, 255, 255, 0.2);
   border-color: var(--dt-button-active-color);
   color: var(--dt-button-active-color);
 


libs/barista-components/button/src/_button-theme.scss
```diff
@@ -87,7 +87,7 @@ a.dt-button-secondary:visited {
   }
 }
 .dt-button-secondary:hover:not([disabled]) {
-  background: $gray-100;
+  background: rgba(255, 255, 255, 0.2);
```
lukasholzer, line 90: @tomheller should we put this in a design token?

tomheller: I don't think it is necessary to put this into a design token. What would make sense is to create opacity (or transparency ones, whatever feels better) tokens, and use the transparentize function from sass to create the color. Could look something like:
```scss
background: transparentize($white, $opacity-low);
```

JanuszLis: Same pattern we use few lines above :) for 
`.dt-button-primary[disabled] {`

Will change it when transparentize(...) fits here more,
@tomheller is @white +@opacity-low just an example howto use it or we could go with that?

tomheller: @JanuszLis this was just an example of how fixed color variables from our tokens could work with transparencies. I don't think you need to go with it. I guess this is something we should tackle on a more global scale (as you mentioned this appears in other places as well). 
Furthermore there are no opacity tokens set up at this moment. 

This suggestion was more a outlook into the future and how this could be solved.

JanuszLis: ok now I get it ;) so leaving current changes as they are. Feel free for any more feedback or review decisions ;)

lukasholzer (approved):

JanuszLis:
thx @lukasholzer 👍

PR #1530, opened 2020-08-27
feat(select): Remove deprecated function in select component

heartdisease:
### <strong>Remove deprecated function in select component</strong>

<hr>

Removed deprecated function `getDtSelectNonFunctionValueError()` in select component and replaced it with an error log.

> I also do not understand why this was deprecated in the first place. Because these changes can lead to a broken component, if the user provides something other than a function. (In my opinion it should rather be changed in the combobox to the solution we have here in the select).

@lukasholzer Any input on this? I haven't touched the select component before, so I cannot answer this question.

### DIFF
--- a/libs/barista-components/experimental/combobox/src/combobox.ts
+++ b/libs/barista-components/experimental/combobox/src/combobox.ts
@@ -201,8 +201,10 @@ export class DtCombobox<T> extends _DtComboboxMixinBase
     // tslint:disable-next-line:strict-type-predicates
     if (typeof fn !== 'function') {
       LOG.error(DT_COMPARE_WITH_NON_FUNCTION_VALUE_ERROR_MSG);
+    } else {
+      this._compareWith = fn;
     }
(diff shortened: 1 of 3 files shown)


thomaspink:
Please change the commit type as it is not a `chore`. Would rather go for a `feat` here. See [contribution guidelines](https://barista.dynatrace.com/components/contribute#type) for reference.

Also please change the text of the commit so it describes what actually changes for the consumer.

I also do not understand why this was deprecated in the first place. Because these changes can lead to a broken component, if the user provides something other than a function. (In my opinion it should rather be changed in the combobox to the solution we have here in the select)

thomaspink:
I also do not understand why this was deprecated in the first place. Because these changes can lead to a broken component, if the user provides something other than a function. (In my opinion it should rather be changed in the combobox to the solution we have here in the select).

lukasholzer (approved):

PR #1548, opened 2020-09-02
chore: Try yarn instead of npm

lukasholzer:
## Try to use yarn instead of npm

As the npm postinstall script does not work properly with`npm_install` from bazel try to use yarn.

### Obstackls

- [ ] release script is parsing the package-lock.json (needs to be updated and adapted to work with `yarn.lock` file.
- [ ] Sha1Sum generation on the ci scripts to speedup everything is with package-lock


lukasholzer:
Not needed as we fixed the bazel npm install problem with #1562

PR #1562, opened 2020-09-07
build(bazel): Fix npm install on windows.

lukasholzer:

Currently `npm_install` cannot access npm binaries in postinstall script on windows. With this small patch the postinstall script is using the global npm executable that is set via nodejs. 

Patch for https://github.com/bazelbuild/rules_nodejs/issues/2144.


SkaillZ (approved):
🎉

thomaspink (approved):

PR #1564, opened 2020-09-07
fix(tree-table): Fixes overflow when info-group contains long texts

### LINKED ISSUE #1502: Tree-table doesn't hide overflow of child nodes
# Bug Report

## Expected Behavior

Tree-table component should hide the overflow of child node cells that contain text which is too long

## Current Behavior

The table's width gets bigger than the viewport and / or its parent element

## Steps to Reproduce

https://stackblitz.com/edit/github-wjgnuy?file=src/app/app.component.ts

## Context (Environment)

The tree-table component will be used to build the Angular version of the waterfall component which is part of Dynatrace DEM.

### Used Versions:

- **node**: >=10.13.0
- **npm**: 6.14.4
- **typescript**: 3.8.3
- **angular**: 9.1.12
- **@angular/cdk**: 9.2.4
- **@dynatrace/barista-components**: 7.5.1

## Detailed Description

The tree-table component should hide text which is too wide for the table in a similar fashion to the Barista table component (with "...")
image

rowa-audil:

This fixes the overflow issue when putting long texts into the info-group part of the tree-table by setting a width and showing the whole text when hovering over the text/cell.

fixes #1502

#### Type of PR

#### Checklist

### DIFF
--- a/apps/dev/src/tree-table/tree-table-demo.component.ts
+++ b/apps/dev/src/tree-table/tree-table-demo.component.ts
@@ -27,7 +27,8 @@ import {
 
 const TESTDATA: ThreadNode[] = [
   {
-    name: 'hz.hzInstance_1_cluster.thread',
+    name:
+      'hz.hzInstance_1_cluster.threadhz.hzInstance_1_cluster.threadhz.hzInstance_1_cluster.threadhz.hzInstance_1_cluster.thread',
     icon: 'airplane',
     threadlevel: 'S0',
     totalTimeConsumption: 150,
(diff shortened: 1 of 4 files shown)


lukasholzer (approved):
@rowa-audil can you please add a small test scenario in the e2e app that checks for this behaviour?

the condition would be `element.offsetWidth < element.scrollWidth` So when the offset is smaller than the scroll width the element gets ellipsis.

heartdisease (approved):

rowa-audil:
@lukasholzer Will do in a follow up issue.

PR #1565, opened 2020-09-07
Feat/fluid input as webcomponent

peter-affenzeller:

FluidInput web-component

#### Type of PR

#### Checklist

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -4,6 +4,7 @@
     "esbenp.prettier-vscode",
     "ms-vscode.vscode-typescript-tslint-plugin",
     "angular.ng-template",
-    "hex-ci.stylelint-plus"
+    "hex-ci.stylelint-plus",
+    "runem.lit-plugin"
   ]
 }
--- a/libs/fluid-elements/input/src/lib/input.spec.ts
+++ b/libs/fluid-elements/input/src/lib/input.spec.ts
@@ -14,15 +14,21 @@
  * limitations under the License.
  */
 
-/*
- * THIS FILE IS GENERATED BASED ON THE DESIGN TOKENS DEFINED IN THE src/**.yml
- * FILES, DO NOT CHANGE MANUALLY.
- * TO GENERATE THESE FILES RUN 'ng build shared-design-tokens'
- */
+import { FluidInput } from './input';
 
-export * from './global/palette';
-export * from './global/spacing';
-export * from './global/theme';
-export * from './global/typography';
-export * from './patterns/button';
-export * from './patterns/input';
+describe('Fluid input', () => {
+  let fixture: FluidInput;
+
+  beforeEach(() => {
+    if (!customElements.get('fluid-input')) {
+      customElements.define('fluid-input', FluidInput);
+    }
+    document.body.innerHTML =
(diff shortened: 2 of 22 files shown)


ffriedl89:
I've tried to use the input and I think it does a bit too much right now. I'd argue that the hint & label & potential validation messages are not bound to the input alone but rather to any form control. 
Having a `fluid-input` that maybe just proxies the HTMLInputElement specific properties seems better after i had a chance to try it out. The markup for the hint, errors and label should in my opinion be moved to a form-field component.

E.g. 
```
<fluid-form-field name="filterby">
    <label>Filter by</label>
    <fluid-input type="text"></dt-input>
</fluid-form-field>
```

heartdisease (approved):

.vscode/extensions.json
```diff
@@ -4,6 +4,7 @@
     "esbenp.prettier-vscode",
     "ms-vscode.vscode-typescript-tslint-plugin",
     "angular.ng-template",
-    "hex-ci.stylelint-plus"
+    "hex-ci.stylelint-plus",
+    "runem.lit-plugin"
```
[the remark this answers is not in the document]
peter-affenzeller: for syntax-highlighting:

libs/fluid-elements/input/src/lib/input.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+    document.body.innerHTML =
+      '<fluid-input hint="FluidInput hint."><label slot="label">Fluid Input Label</label><input type="text" placeholder="Fluid Input" /></fluid-input>';
+    fixture = document.querySelector<FluidInput>('fluid-input')!;
+  });
+
+  it('should create the input', async () => {
+    expect(fixture).not.toBe(null);
+  });
+});
```
lukasholzer, line 34: please add some more test cases :D or create follow up issue for it

libs/fluid-elements/input/src/lib/input.ts
```diff
@@ -0,0 +1,183 @@
...
+  private _handleSlotChange({ target }: { target: HTMLSlotElement }): void {
+    const slottedElements = target.assignedElements({
+      flatten: true,
+    });
+
+    // Only accept one element per slot
+    if (slottedElements.length > 1) {
+      throw new Error(
+        `The fluid-input only takes one input element in the default slot.`,
+      );
```
lukasholzer, line 115: @ffriedl89 do we really want to throw an error or is logging enough?

lukasholzer (dismissed):

libs/fluid-elements/input/src/lib/input.ts
ffriedl89: I think with a changed API where the form field is extracted we can add a warning/error log. Throwing an error is usually not the go to i'd say. For now I wouldn't bother.

heartdisease (approved):

lukasholzer (approved):

bardosmisi (approved):
I think the input component should be added to the fluid-dev app in order to present what it can do and how does it look like. The other, already done fluid components are also in there (button, checkbox, tab, switch).

Also I guess the component should have a name and a value property in order to make it serializable, but that can be done in the future as well.

Otherwise LGTM 🦛 !

lukasholzer:
Manual cherry-picked into `8.x` and `8.0.x`

peter-affenzeller:
thx @bardosmisi, I added this to the follow-up issue (#1596)

PR #1585, opened 2020-09-10
fix(filter-field): Fixed UX problem where text that wasn't applied to the filter would remain after losing focus.

### LINKED ISSUE #1572: Filter field: Clear free text on lost focus
# Bug Report<!---

## Expected Behavior

Random text in filter field should be cleared on focus loss

## Current Behavior

Text remains as it is.
In the filter field, users can type in some text that does not match any available option. This text remains in the field even after it looses focus which can wrongly give the impression that some filter is applied when it is not

## Possible Solution

Remove any text in the field on focus loss and keep only the selected filter pills. 

image

SkaillZ:

Fixes #1572 

<hr>

#### Type of PR

Bugfix

### DIFF
--- a/libs/barista-components/filter-field/src/filter-field.spec.ts
+++ b/libs/barista-components/filter-field/src/filter-field.spec.ts
@@ -252,6 +252,20 @@ describe('DtFilterField', () => {
     });
   });
 
+  describe('focus loss', () => {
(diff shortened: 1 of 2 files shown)


lukasholzer (approved):
THX :) 🐝

thomaspink (approved):

PR #1588, opened 2020-09-10
build(context-dialog): Bazelify context-dialog component

SkaillZ:
#### Type of PR

Feature


libs/barista-components/switch/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "../tsconfig.json",
+  "extends": "../../../tsconfig.base.json",
```
lukasholzer, line 2: please extend from the `libs/barista-components/tsconfig.json`.

lukasholzer (changes requested):

libs/barista-components/switch/tsconfig.json
SkaillZ: Thanks for catching this.

lukasholzer (approved):

PR #1599, opened 2020-09-11
build(quick-filter): Bazelify quick-filter component

SkaillZ:

Build quick-filter component with Bazel.

#### Type of PR

Other


libs/barista-components/experimental/quick-filter/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "../../../../tsconfig.base.json",
+  "extends": "../../tsconfig.json",
```
lukasholzer, line 2: This is the wrong tsconfig please use the one from the components

lukasholzer (changes requested):

libs/barista-components/experimental/quick-filter/tsconfig.json
lukasholzer: Never mind this is fine! overseen that it is nested in a folder

lukasholzer (approved):

lukasholzer (approved):

PR #1620, opened 2020-09-15
chore: Remove the customizable tooltip references from the micro-chart doc

bardosmisi:

Remove the customizable tooltip references from the micro-chart documentation. It will not be possible in the future to have customizable tooltip for the micro chart, only the default one. The references to hat approach has to be removed from the documentation.

#### Type of PR


lukasholzer (approved):
Redacting ✏️__________ 🤓

heartdisease (approved):

PR #1621, opened 2020-09-15
fix(filter-field): Fixes an issue where custom tag parsers are not applied on filters set initially.

### LINKED ISSUE #1591: [filter-field] customTagParser doesn't work with pre-initialized filters
# Bug Report

It seems that the `customTagParser` is just called for new items, or when re-rendering existing ones. It would be expected, that it is also applied after the first loading .

## Steps to Reproduce
https://stackblitz.com/edit/github-z7swbn?file=src/filter-field/filter-field-programmatic-filters-example/filter-field-programmatic-filters-example.ts
* when loading the page, an Emoji should be visible near the items, but it isn't
* it only applies, when changing some items (eg. focusing)

image

### My customTagParser
```
customParser(
    filterValues: DtFilterValue[],
    editable?: boolean,
    deletable?: boolean,
  ): DtFilterFieldTagData | null {
    const tagData = defaultTagDataForFilterValuesParser(
      filterValues,
      editable,
      deletable,
    );
    if (tagData) {
      tagData.key = '❤ ' + tagData.key;
    }
    return tagData;
  }
```

thomaspink:

fix(filter-field): Fixes an issue where custom tag parsers are not applied on filters set initially.

Fixes #1591

#### Type of PR

#### Checklist

### DIFF
--- a/apps/dev/src/filter-field/filter-field-demo.component.html
+++ b/apps/dev/src/filter-field/filter-field-demo.component.html
@@ -4,6 +4,8 @@
   [disabled]="_disabled"
   label="Filter by"
   clearAllLabel="Clear all"
+  [filters]="filters"
+  [customTagParser]="customParser"
   (inputChange)="inputChange($event)"
   (filterChanges)="filterChanges($event)"
   (currentFilterChanges)="currentFilterChanges($event)"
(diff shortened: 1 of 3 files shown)


thomaspink:
Note: Could not find a way to proper test this because there are already tests for exact this issue in place that never failed because at some point you need to trigger change detection manually in the test which makes it always work

heartdisease (approved):

lukasholzer (approved):

PR #1627, opened 2020-09-16
WebComponents: Adds PropertyValue to aria labeling and fixes disabled in button-group

rowa-audil:

Adds PropertyValue instead of a Map and fixes the disabled state of the button-group removing comments in the CSS

#### Type of PR

### DIFF
--- a/apps/fluid-dev/src/pages/button-group/button-group-page.component.html
+++ b/apps/fluid-dev/src/pages/button-group/button-group-page.component.html
@@ -1,11 +1,11 @@
 <fluid-button-group
   (checkedChange)="handleChange($event)"
-  [disabled]="disableall"
+  [disableAll]="disableall"
 >
-  <fluid-button-group-item name="btngrp">
+  <fluid-button-group-item name="btngrp" [disabled]="disable">
     Options 1
   </fluid-button-group-item>
-  <fluid-button-group-item name="btngrp" [disabled]="disable">
+  <fluid-button-group-item name="btngrp">
     Buildingsteps 2
   </fluid-button-group-item>
   <fluid-button-group-item *ngIf="show" name="btngrp">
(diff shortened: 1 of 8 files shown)


peter-affenzeller (dismissed):
lgtm 🙂

apps/fluid-dev/src/pages/button-group/button-group-page.component.html
```diff
@@ -2,10 +2,10 @@
   (checkedChange)="handleChange($event)"
   [disabled]="disableall"
 >
-  <fluid-button-group-item name="btngrp">
+  <fluid-button-group-item name="btngrp" [disabled]="disable">
```
bardosmisi, line 5: NIT: Personally I feel this is  bit off :). I think this should be "disabled", and "disableAll".

bardosmisi (dismissed):
One NIT comment added, other than that, LGTM.

libs/fluid-elements/button-group/src/lib/button-group-item/button-group-item.ts
```diff
@@ -83,12 +83,10 @@ export class FluidButtonGroupItem extends LitElement {
       }
 
       :host([disabled]) .fluid-label {
-        // TODO: Should be replaced by an opacity token at some point.
         opacity: 0.5;
       }
 
       :host([disabled]) .fluid-button-group-item-svg {
-        // TODO: Should be replaced by an opacity token at some point.
```
lukasholzer, line 91: Why do you removed it?

rowa-audil: Very weird. It doesn't throw any errors anywhere but the opacity is not set when there's a comment above it. 🤔

apps/fluid-dev/src/pages/button-group/button-group-page.component.html
rowa-audil: @bardosmisi One disabled property is on the wrapper so I thought maybe users would understand that this disables all inside items. I agree that `disableAll` is more clear and easy to understand

bardosmisi (dismissed):

peter-affenzeller (dismissed):

bardosmisi (dismissed):

lukasholzer (approved):

lukasholzer:
@rowa-audil please squash your commits

lukasholzer (approved):

PR #1630, opened 2020-09-16
feat(alert): Introduce `info` severity type.

bartoszbobin:
image

#### Type of PR

#### Checklist

- [x] I have added necessary documentation

### DIFF
--- a/libs/barista-components/alert/src/alert.ts
+++ b/libs/barista-components/alert/src/alert.ts
@@ -20,6 +20,13 @@ import {
   Input,
   ViewEncapsulation,
 } from '@angular/core';
+import { DtIconType } from '@dynatrace/barista-icons';
+
+const ALERT_SEVERITY_TO_ICON_MAP: Record<string, DtIconType> = {
+  error: 'criticalevent',
+  warning: 'incident',
+  info: 'information',
+};
 
 @Component({
   selector: 'dt-alert',
@@ -29,6 +36,7 @@ import {
     role: 'alert',
     class: 'dt-alert',
     '[class.dt-alert-warning]': 'severity === "warning"',
+    '[class.dt-alert-info]': 'severity === "info"',
   },
   preserveWhitespaces: false,
   changeDetection: ChangeDetectionStrategy.OnPush,
@@ -36,5 +44,12 @@ import {
 })
 export class DtAlert {
   /** The severity type of the alert. */
-  @Input() severity: 'error' | 'warning' = 'error';
+  @Input() severity: 'error' | 'warning' | 'info' = 'error';
+
+  get severityIcon(): DtIconType {
+    return (
+      ALERT_SEVERITY_TO_ICON_MAP[this.severity] ??
+      ALERT_SEVERITY_TO_ICON_MAP.error
+    );
+  }
 }
--- a/apps/dev/src/alert/alert-demo.component.html
+++ b/apps/dev/src/alert/alert-demo.component.html
@@ -1,1 +1,4 @@
-<dt-alert severity="warning">This is an error message!</dt-alert>
+<dt-alert>This is a default severity message!</dt-alert>
+<dt-alert severity="error">This is an error message!</dt-alert>
+<dt-alert severity="warning">This is a warning message!</dt-alert>
+<dt-alert severity="info">This is an info message!</dt-alert>
--- a/apps/dev/src/alert/alert-demo.component.scss
+++ b/apps/dev/src/alert/alert-demo.component.scss
@@ -1,3 +1,7 @@
 :host {
   display: block;
 }
+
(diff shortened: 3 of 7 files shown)


libs/barista-components/alert/src/alert.ts
```diff
@@ -29,12 +30,27 @@ import {
...
+    '[class.dt-alert-success]': 'severity === "success"',
   },
   preserveWhitespaces: false,
   changeDetection: ChangeDetectionStrategy.OnPush,
   encapsulation: ViewEncapsulation.Emulated,
 })
 export class DtAlert {
   /** The severity type of the alert. */
-  @Input() severity: 'error' | 'warning' = 'error';
+  @Input() severity: 'error' | 'warning' | 'info' | 'success' = 'error';
```
jaroslawsawicki, line 42: Don't we use Object literals for types like this?

libs/barista-components/alert/src/alert.ts
```diff
@@ -29,12 +30,27 @@ import {
...
+
+  get severityIcon(): DtIconType {
```
jaroslawsawicki, line 44: It could be stored as a map and used as
```ts
return severityToIconType[this.severity];
```
and the method could be named as
```ts
  get severityIconType(): DtIconType {
```

libs/barista-components/alert/src/alert.ts
bartoszbobin: Did you mean TypeScript types?
```
type AlertSeverityType = 'error' | 'warning' | 'info' | 'success';
```

libs/barista-components/alert/src/alert.ts
bartoszbobin: 👍 for adding a map but will leave `severityIcon`

libs/barista-components/alert/src/alert.ts
jaroslawsawicki: ie string enums
```ts
enum Severity {
  ERROR = 'error',
  WARNING = 'warning',
}
```

lukasholzer: No, we use tuple types without a separate type as it is more consumer-friendly. When you hide the values in a `type` the consumer will get the `type AlertServityType` in his autocompletion. If you provide the tuple directly the consumer has first-class autocompletion with the appropriate values.

lukasholzer:
@bartoszbobin Hey thank you very much for taking the time to contribute something for the library. 

I'm very curious where the requirement comes from to support those two additional types of alert, as my short sync with @dynatrace-oss/ux-core resulted in, that they are not aware of this. So I would ask you kindly to open an issue where you elaborate your use case and describe your problem.

To avoid declining a PR or changing it a lot, we recommend opening an issue first so everybody can see what the intent behind a new feature is. Furthermore, you provide a platform for discussion where you can gather a lot of opinions.

kathrin-aigner:
@bartoszbobin the Info alert is approved from the ux side. For the success message please use depending on your use case our confirmation dialog [https://barista.dynatrace.com/components/confirmation-dialog](url) or a toast [https://barista.dynatrace.com/components/toast](url).

bartoszbobin:
Due to some doubts about `success` state, PR was limited to `info` severity type

libs/barista-components/alert/src/alert.ts
```diff
@@ -29,12 +38,20 @@ import {
...
+    '[class.dt-alert-info]': 'severity === "info"',
   },
   preserveWhitespaces: false,
   changeDetection: ChangeDetectionStrategy.OnPush,
   encapsulation: ViewEncapsulation.Emulated,
 })
 export class DtAlert {
   /** The severity type of the alert. */
-  @Input() severity: 'error' | 'warning' = 'error';
+  @Input() severity: AlertSeverityType = 'error';
```
lukasholzer, line 49: Please don't put the union type in a type declaration. if you inline it the consumer has better IDE support for the value of the input.

```suggestion
  @Input() severity: 'error' | 'warning' | 'info' = 'error';
```

lukasholzer (changes requested):

libs/barista-components/alert/src/alert.ts
bartoszbobin: Fixed

lukasholzer (approved):
Awesome! 🎉 

Thank you very much for your contribution! Sorry for the inconvenience with UX but we always have to get UX approvals for our barista components, as they might not only be used in the cluster – they can be used in the whole dynatrace domain.

jaroslawsawicki (approved):

heartdisease (approved):

lukasholzer:
manual cherry-picked into `8.x`

Issue #1632, opened 2020-09-17
[Filter field] suggestions list disappear after clicking on one of suggested entries


pawel-bugalski-dynatrace:
# Bug Report

## Expected Behavior

After clicking on an item in a list it should be field into the filter field

## Current Behavior

Suggestions list disappear (note that pressing enter does work).

## Steps to Reproduce

![Screen Recording 2020-09-17 at 14 49 51](https://user-images.githubusercontent.com/67259865/93474698-005e9d80-f8f8-11ea-932b-2792e7b0e7b9.gif)

This is reproduced in a demo barista app but not in stackblitz (as on the time of writing it has a too old varsion)
http://localhost:4200/filter-field-default-example
### Used Versions:

- **node**: 
- **npm**: 
- **typescript**: 
- **angular**: 
- **@angular/cdk**: 
- **@dynatrace/barista-components**: 

## Detailed Description

This seems to be a regression caused by https://github.com/dynatrace-oss/barista/pull/1625

samuelfahrngruber:
When trying to reproduce this reliably, I found out that this happens more frequently, when there are so few filter items displayed after filtering by text, that the scrolling bar disappears.

#### e.g. This works: 
![grafik](https://user-images.githubusercontent.com/35682879/93578308-7537e300-f99d-11ea-9742-eb2d60b87757.png)
#### This doesn't:
![grafik](https://user-images.githubusercontent.com/35682879/93578442-a0bacd80-f99d-11ea-8492-8a3ccb2cf109.png)

Also it seems that the issue happened after #1585, as commenting out `this._writeInputValue('');` solved the problem in the barista demo app for me.

PR #1650, opened 2020-09-22
Revert "fix(filter-field): Fixed UX problem where text that wasn't applied to the filter would remain after losing focus."

samuelfahrngruber:
Reverts dynatrace-oss/barista#1585
Fixes dynatrace-oss/barista#1632

---

### Explaination
The code caused the filter field to not properly select suggestion values when typing in filter text any more.
This seems to be caused by the fact, that the filter field does not only loose focus when the user clicks outside of the filter field but also when the user selects a prefiltered value from the suggestion dropdown.

IMHO clearing the text input when the filter field looses focus is an ux improvement and should not be seen as a bug, thus not being included in any patch version.

I would propose to re-open #1572 and postpone this change to `8.1.x`, keeping this regression in mind when re-implementing.

### DIFF
--- a/libs/barista-components/filter-field/src/filter-field.spec.ts
+++ b/libs/barista-components/filter-field/src/filter-field.spec.ts
@@ -252,20 +252,6 @@ describe('DtFilterField', () => {
     });
   });
 
-  describe('focus loss', () => {
-    it('should clear the current input value', fakeAsync(() => {
-      const input = fixture.debugElement.query(By.css('.dt-filter-field-input'))
-        .nativeElement;
-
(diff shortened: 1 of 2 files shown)


thomaspink:
Thank you for the revert. Please reword the commit to fit our [guidelines](https://barista.dynatrace.com/components/contribute#commit-message-guidelines) otherwise the tooling will miss it.

lukasholzer (approved):

lukasholzer:
Successful cherry picked into `8.x` and `8.0.x`

PR #1663, opened 2020-09-25
chore(fluid-virtual-scroll): Virtual scroll container as web-component.

peter-affenzeller:

Overall, the virtual scroll container is doing what it's supposed to do, although there are still a couple of issues left to tackle (fast scrolling, items with different height slightly jumping, more tests), but since the combo-box is relying on the virtual-scroll-container and we want to get a first version out, please take a look at the current state and let me know what you think.

#### Type of PR

#### Checklist

- [x] I have **partially** added tests that prove my fix is effective or that my feature works

### DIFF
--- a/libs/fluid-elements/virtual-scroll-container/src/lib/virtual-scroll-container.ts
+++ b/libs/fluid-elements/virtual-scroll-container/src/lib/virtual-scroll-container.ts
@@ -1,0 +1,616 @@
+/**
+ * @license
+ * Copyright 2020 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  CSSResult,
+  LitElement,
+  PropertyValues,
+  TemplateResult,
+  css,
+  customElement,
+  html,
+  property,
+  query,
+  unsafeCSS,
+} from 'lit-element';
+import { render } from 'lit-html';
+import { ifDefined } from 'lit-html/directives/if-defined';
+
+import { FLUID_INPUT_PADDING } from '@dynatrace/fluid-design-tokens';
+
+import { FluidVirtualScrollState } from './virtual-scroll-state';
+import { FluidVirtualScrollContainerRenderedItemsChange } from './virtual-scroll-events';
+import {
+  virtualScrollAddInitialItemsBatch,
+  virtualScrollCalculatePlaceholderSize,
+  virtualScrollResetState,
+  virtualScrollUpdateEndScrollBuffer,
+  virtualScrollUpdateStartScrollBuffer,
+} from './virtual-scroll-utils';
+
+let _unique = 0;
+
+// Threshold to dertermine when the observer should trigger the callback
+const VIRTUAL_SCROLL_THRESHOLD = 0.5;
+
+/**
+ * This is an experimental virtual-scroll-container element built with lit-elements and
+ * web-components. It registers itself as `fluid-virtual-scroll-container` custom element.
+ * @element fluid-virtual-scroll-container
+ */
+@customElement('fluid-virtual-scroll-container')
+export class FluidVirtualScrollContainer<T> extends LitElement {
+  /** Styles for the virtual-scroll-container component */
+  static get styles(): CSSResult {
+    return css`
+      :host {
+        display: block;
+        width: 100%;
+        height: 100%;
+      }
+
+      *,
+      *::before,
+      *::after {
+        box-sizing: border-box;
+      }
+
+      .fluid-virtual-scroll-container {
+        --virtual-scroll-item-height: 0;
+
+        width: 100%;
+        height: 100%;
+        overflow-x: hidden;
+        overflow-y: scroll;
+      }
+
+      .fluid-virtual-scroll-empty {
+        padding: ${unsafeCSS(FLUID_INPUT_PADDING)};
+      }
+
+      .fluid-virtual-scroll-item--fixed-height {
+        height: var(--virtual-scroll-item-height);
+        overflow: hidden;
+      }
+    `;
+  }
+
+  /**
+   * Defines the tab element with an id attribute
+   * @attr
+   */
+  @property({ type: String, reflect: true })
+  scrollcontainerid = `fluid-virtual-scroll-container-${_unique++}`;
+
+  /**
+   * Defines the class(es) to add to the rendered virtual scroll items
+   * @attr
+   */
+  @property({ type: Boolean, reflect: true })
+  equalitemheight: boolean = false;
+
+  /**
+   * Defines the class(es) to add to the rendered virtual scroll items
+   * @attr
+   */
+  @property({ type: String, reflect: true })
+  itemclass: string;
+
+  /**
+   * Defines what to display if there are no items to render
+   * @attr
+   */
+  @property({ type: String, reflect: true })
+  noitemsmessage = `No items to display.`;
+
+  /**
+   * Defines the max height of the virtual scroll container
+   * If this is not set, a parent of the virtual scroll container should
+   * have an explicit height, else it will grow until all items are rendered
+   * @attr
+   */
+  @property({ type: String, reflect: true })
+  maxheight: string;
+
+  /**
+   * Array of items to display in the scroll container.
+   * @attr
+   * @type array
+   */
+  @property({ type: Array, attribute: false })
+  get items(): T[] {
+    return this._items;
+  }
+  set items(value: T[]) {
+    if (value !== this._items) {
+      const oldItems = this._items;
+      this._items = value;
+      this._resetVirtualScroll();
+      this.requestUpdate(`items`, oldItems);
+    }
+  }
+  private _items: T[] = [];
+
+  /**
+   * Defines the function to use for rendering items
+   * @attr
+   * @type function
+   */
+  @property({ attribute: false })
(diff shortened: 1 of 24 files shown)


heartdisease (dismissed):

libs/fluid-elements/virtual-scroll-container/src/lib/virtual-scroll-state.ts
```diff
@@ -0,0 +1,291 @@
...
+  first: number;
+  last: number;
+}
+
+// Number of items to render each time until the container is filled up
+const VIRTUAL_SCROLL_BATCH_SIZE = 5;
+// Number of items to render in addition to the initially rendered items
+const VIRTUAL_SCROLL_BUFFER_SIZE = 3;
+
+export class VirtualScrollState {
```
thomaspink, line 32: General question, do we prefix exported symbols like on the angular side (`dt`, `fluid`)?

libs/fluid-elements/virtual-scroll-container/src/lib/virtual-scroll-state.ts
```diff
@@ -0,0 +1,291 @@
...
+  endItemsBufferSize = 0;
+
+  /** Index of the item to observe for adjusting the start buffer */
+  bufferStartObservedItemIndex = -1;
+
+  /** Index of the item to observe for adjusting the end buffer */
+  bufferEndObservedItemIndex = -1;
+
+  /** Total number of items in the virtual scroll panel */
+  set itemsCount(value: number) {
```
thomaspink, line 70: Would recommend always having a getter/setter (adding the getter here) pair, unless it is a special case that requires it.

libs/fluid-elements/virtual-scroll-container/src/lib/virtual-scroll-state.ts
```diff
@@ -0,0 +1,291 @@
...
+  private _firstRenderedItemIndex = -1;
+
+  /** Index of the last rendered item */
+  private _lastRenderedItemIndex = -1;
+
+  constructor({
+    itemsBatchSize,
+    itemsBufferSize,
+  }: FluidVirtualScrollOptions = {}) {
+    this._itemsBatchSize = itemsBatchSize
```
thomaspink, line 100: NIT: suggestion
```ts
    this._itemsBatchSize = itemsBatchSize ?? VIRTUAL_SCROLL_BATCH_SIZE;
    this. _itemsBufferSize = itemsBufferSize ?? VIRTUAL_SCROLL_BUFFER_SIZE;
```

thomaspink (changes requested):

libs/fluid-elements/virtual-scroll-container/src/lib/virtual-scroll-state.ts
peter-affenzeller: yep, we should

libs/fluid-elements/virtual-scroll-container/src/lib/virtual-scroll-container.ts
```diff
@@ -0,0 +1,582 @@
...
+            entry.rootBounds!.bottom
+        ) {
+          // Observed item is at the end of the list, the item is already above the
+          // bottom edge of the container, hence the observer would not trigger again
+          // when scrolling further down
+          this._clearVirtualScrollBufferObserver();
+          this._scrollState.updateEndScrollBuffer();
+          this._renderVirtualScrollItems();
+          this._updatePlaceholders();
+        } else if (
```
lukasholzer, line 450: Instead of all the `else if` statements try to use early exists as it makes the code way more readable. Otherwise, you get a deep nesting and you don't need in which part of the if you are.

libs/fluid-elements/virtual-scroll-container/src/lib/virtual-scroll-container.ts
```diff
@@ -0,0 +1,582 @@
...
+
+  /** Sets the height of the placeholder */
+  private _setPlaceholders(): void {
+    if (this._items.length) {
+      this._scrollStartPlaceholder.style.height = `${this._scrollState.startPlaceholderSize}px`;
+      this._scrollEndPlaceholder.style.height = `${this._scrollState.endPlaceholderSize}px`;
+    } else {
+      this._scrollStartPlaceholder.style.height = `0`;
+      this._scrollEndPlaceholder.style.height = `0`;
+    }
```
lukasholzer, line 531: This would be an example for an early exit
```suggestion
    if (this._items.length) {
      this._scrollStartPlaceholder.style.height = `${this._scrollState.startPlaceholderSize}px`;
      this._scrollEndPlaceholder.style.height = `${this._scrollState.endPlaceholderSize}px`;
      return;
    } 
    
    this._scrollStartPlaceholder.style.height = `0`;
    this._scrollEndPlaceholder.style.height = `0`;
```

libs/fluid-elements/virtual-scroll-container/src/lib/virtual-scroll-state.ts
```diff
@@ -0,0 +1,291 @@
...
+   * this getter nor rely on its state after {@link #updateStartScrollBuffer()}
+   * or {@link #updateEndScrollBuffer()} have been called again!
+   * In case you need to store that information past this point, always create
+   * a copy right away instead.
+   * @ ChMa: pls explain again why it should be copied? I forgot what the potential issue was...
+   */
+  renderedItemsRange: FluidVirtualScrollRenderedItemRange = Object.seal({
+    first: -1,
+    last: -1,
+  });
```
lukasholzer, line 49: 1. Why does this have to be an object? Can we just have 2 properties? `firstItemIndex` and `lastItemIndex` same as:
image

2. The whole class should be marked as internal and not be exported to the user as this is an internal state.

3. Why do we need an object seal here? It puts only overhead in the code that is not needed. Please remove
```suggestion
  renderedItemsRange: FluidVirtualScrollRenderedItemRange = {
    first: -1,
    last: -1,
  };
```

libs/fluid-elements/virtual-scroll-container/src/lib/virtual-scroll-state.ts
```diff
@@ -0,0 +1,291 @@
...
+    this._lastRenderedItemIndex = this.renderedItemsRange.last;
+
+    // Calculate number of items of the end buffer
+    this.endItemsBufferSize =
+      this._renderedItemsCount() -
+      this._initialBatches * this._itemsBatchSize -
+      this.startItemsBufferSize;
+
+    this._calculateObservedItemsIndices();
+  }
```
lukasholzer, line 138: A state should not have methods in my opinion. A state should be an object that holds a state not more.

libs/fluid-elements/virtual-scroll-container/src/lib/virtual-scroll-state.ts
```diff
@@ -0,0 +1,291 @@
...
+        this._lastRenderedItemIndex - Math.floor(this._itemsBufferSize * 0.5);
+      this.bufferStartObservedItemIndex =
+        this.bufferEndObservedItemIndex -
+        this._initialBatches * this._itemsBatchSize -
+        this._itemsBufferSize;
+    }
+  }
+
+  /** Calculates the number of currently rendered items */
+  private _renderedItemsCount(): number {
```
lukasholzer, line 288: this can be a getter

libs/fluid-elements/virtual-scroll-container/src/lib/virtual-scroll-state.ts
```diff
@@ -0,0 +1,291 @@
...
+        this._lastRenderedItemIndex - Math.floor(this._itemsBufferSize * 0.5);
+    } else {
+      // Observe start and end of the rendered items
+      this.bufferEndObservedItemIndex =
+        this._lastRenderedItemIndex - Math.floor(this._itemsBufferSize * 0.5);
+      this.bufferStartObservedItemIndex =
+        this.bufferEndObservedItemIndex -
+        this._initialBatches * this._itemsBatchSize -
+        this._itemsBufferSize;
+    }
```
lukasholzer, line 284: please use early exits for the else if madness. And move into pure function that returns an object with the indexes for easier testing later on.

lukasholzer (changes requested):
Awesome component! 🐝

But a little bit of general feedback for the component:
1. Please try to avoid long changing of `if else` and nested `if` statements as it makes the code really hard to understand and read.

Early exits have better readability.

2. Try to extract parts into pure functions as the classes are really long and sometimes hard to read. If you extract parts in functions you can test them better and encapsulate logic so it makes readability hard.

3. A state should always be an object that should be updated on one single place. No big methods inside the state. I would separate the two things.
The best solution would be you have a state and it gets modified by a pure function that returns a new state. Immutal and side-effect free.

rowa-audil (dismissed):
Other than the suggestions in the stand-up it LGTM

libs/fluid-elements/virtual-scroll-container/src/lib/virtual-scroll-container.ts
```diff
@@ -0,0 +1,582 @@
...
+      } else if (
+        this._scrollState.renderedItemsRange.last !==
+        this.items.length - 1
+      ) {
+        // Call scroll state update end buffer to add the buffer items
+        this._scrollState.updateEndScrollBuffer();
+
+        this._renderVirtualScrollItems();
+
+        requestAnimationFrame(() => {
```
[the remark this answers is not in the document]
peter-affenzeller: had another `requestAnimationFrame` in place that made this one necessary, but is actually unnecessary with some slight changes... unfortunately not related to scrolling issue...

libs/fluid-elements/virtual-scroll-container/src/lib/virtual-scroll-container.ts
peter-affenzeller: I didn't do a lot of cleanup for this one, since this is most certainly the problem when it comes to the container stopping to render items and will be subject to change.

rowa-audil (approved):

rowa-audil (approved):

lukasholzer (approved):

lukasholzer:
@peter-affenzeller can you please squash your commits to one for merging?

thomaspink (approved):

rowa-audil (approved):

lukasholzer (approved):

PR #1673, opened 2020-09-29
feat(radial-chart): Add interaction to legend

### LINKED ISSUE #1478: Radial chart: Add legend filtering capabilities
# Bug Report

We need, the **radial chart** to have the filtering capabilities, as in other barista's charts, by clicking in the legends.

## Expected Behavior

When the user clicks on any of the legend item, in the chart, should toggle the visibility of this one in chart.

**_You can use the following chart as an example:_**
https://barista.dynatrace.com/components/chart#donut-or-pie-chart

| Shown             |  Hidden |
:-------------------------:|:-------------------------:
image | image

## Current Behavior

For the current implementation, if the user clicks, in one in the legend, nothing happens.

https://barista.dynatrace.com/components/radial-chart

## Steps to Reproduce

1. Go to https://barista.dynatrace.com/components/radial-chart
2. Click on any legend

subarroca:
Bugfix

Closes #1478 
Add interaction to radial chart.
The legend now hides and shows the paths.
The selection is not affected. It could be in the future

#### Checklist

### DIFF
--- a/libs/barista-components/radial-chart/src/radial-chart-path.ts
+++ b/libs/barista-components/radial-chart/src/radial-chart-path.ts
@@ -16,6 +16,7 @@
 
 import { AfterContentInit, Component, Input, TemplateRef } from '@angular/core';
 import { DtOverlayTrigger } from '@dynatrace/barista-components/overlay';
+import { DtColors } from '@dynatrace/barista-components/theming';
 import { DtRadialChartRenderData } from './utils/radial-chart-interfaces';
 
 export interface DtRadialChartOverlayData {
@@ -73,5 +74,10 @@ export class DtRadialChartPath
         },
       };
     }
+  }
+
+  /** @internal Get slice color based on its status */
+  get color(): string {
+    return this.series.origin.active ? this.series.color : DtColors.GRAY_100;
   }
 }
--- a/libs/barista-components/radial-chart/src/radial-chart.ts
+++ b/libs/barista-components/radial-chart/src/radial-chart.ts
@@ -377,4 +377,18 @@ export class DtRadialChart implements AfterContentInit, OnDestroy {
       series.origin.selectedChange.emit(series.origin.selected);
     }
   }
+
+  /** @internal Toggle the visibility of an element */
+  _toggleLegend(slice: DtRadialChartRenderData): void {
+    if (
+      // Keep at least one element visible at any time
+      this._renderData.filter((node) => node.origin.active).length > 1 ||
+      // If all but one are inactive, allow the clicked one to be activated again
+      !slice.origin.active
+    ) {
+      slice.origin.active = !slice.origin.active;
+
+      this._updateRenderData();
+    }
+  }
 }
--- a/apps/components-e2e/src/components/radial-chart/radial-chart.e2e.ts
+++ b/apps/components-e2e/src/components/radial-chart/radial-chart.e2e.ts
@@ -16,21 +16,20 @@
 
 import { resetWindowSizeToDefault, waitForAngular } from '../../utils';
 import {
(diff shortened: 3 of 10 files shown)


libs/barista-components/radial-chart/src/radial-chart-path.ts
```diff
@@ -74,4 +75,8 @@ export class DtRadialChartPath
       };
     }
   }
+
+  get color(): string {
```
lukasholzer, line 79: Please add JsDoc comment :)

lukasholzer (approved):
Awesome! 🎉 
Thank you very much for contributing this feature and adding e2e tests to it to help improve our quality! 🐝

libs/barista-components/radial-chart/src/radial-chart.ts
```diff
@@ -377,4 +377,18 @@ export class DtRadialChart implements AfterContentInit, OnDestroy {
...
+
+  /** @internal Toggle the visibility of an element */
+  _toggleLegend(slice: DtRadialChartRenderData): void {
+    // don't allow hiding last element
+    if (
+      this._renderData.filter((node) => node.origin.active).length > 1 ||
+      !slice.origin.active
+    ) {
+      slice.origin.active = !slice.origin.active;
+      // updateNodesVisibility(this._filledSeries, this._legends);
```
lukasholzer, line 389: please remove unused comment

lukasholzer (changes requested):

lukasholzer (dismissed):

libs/barista-components/radial-chart/src/radial-chart.ts
```diff
@@ -377,4 +377,17 @@ export class DtRadialChart implements AfterContentInit, OnDestroy {
       series.origin.selectedChange.emit(series.origin.selected);
     }
   }
+
+  /** @internal Toggle the visibility of an element */
+  _toggleLegend(slice: DtRadialChartRenderData): void {
+    // don't allow hiding last element
+    if (
+      this._renderData.filter((node) => node.origin.active).length > 1 ||
+      !slice.origin.active
```
rowa-audil, line 386: I think you don't need the second statement because it would already be false, right?

rowa-audil (changes requested):
Could you please explain why you did that?

libs/barista-components/radial-chart/src/radial-chart.ts
subarroca: You're right, second condition is not needed

subarroca: No, the condition is needed because if you left only one active legend you would not be able to re activate the rest of them

rowa-audil (approved):
LGTM

bardosmisi (dismissed):

peter-affenzeller (approved):

peter-affenzeller (changes requested):
cool 👍 
there is a bug not directly related to the legend interaction, but rather the overlay-trigger that the legend-items are extending form.
when you focus a legend item and then hit enter, a type error is thrown, since the overlay-trigger would usually pin the overlay on enter, but of course the legend-item does not have an overlay in this case, hence a type error is thrown.
would be great if you could take a look at this before we move on 🙂

peter-affenzeller (approved):
cool 👍  thx 🙂

rowa-audil (approved):

lukasholzer (approved):

lukasholzer:
@subarroca please rebase and squash your changes into one commit so that we can merge it :)

rowa-audil (approved):

lukasholzer (approved):

Issue #1680, opened 2020-10-01
Accessebility tests are failing due to a violation in the combobox


lukasholzer:
# Bug

Fix the a11y tests for the combo box so that the tests are passing again.
![image.png](https://images.zenhubusercontent.com/5ea980b320880f5153072208/d135d1e6-8e4b-44d9-97a7-84347e6e003c)

PR #1690, opened 2020-10-05
fix(a11y): Fixes flaky combobox a11y test and changes aria attribute in button-group-item

rowa-audil:

I've updated axe-core to v4.0.2 and then let the a11y tests run locally and on the CI a few times without errors thrown.
This resulted in the ButtonGroupItem component to fail due to the a11y rule 
`Elements must only use allowed ARIA attributes`, so I changed the attribute from `aria-selected` to `aria-checked`.

Fixes: #1680 

#### Type of PR

#### Checklist

### DIFF
--- a/libs/barista-components/button-group/README.md
+++ b/libs/barista-components/button-group/README.md
@@ -54,14 +54,24 @@ To apply the button group component, use the `<dt-button-group>` and
 
 ## Button group item inputs
 
-| Name           | Type                  | Default     | Description                                                                                            |
-| -------------- | --------------------- | ----------- | ------------------------------------------------------------------------------------------------------ |
-| `<ng-content>` |                       |             | The content which is displayed inside of the item. This should only be text.                           |
-| `value`        | `T | undefined`       | `undefined` | The associated value of this item                                                                      |
-| `disabled`     | `boolean | undefined` | `undefined` | Sets disabled state if property is set and the value is truthy or undefined                            |
-| `tabIndex`     | `number`              | `0`         | Sets and gets the tabIndex property                                                                    |
-| `selected`     | `boolean`             | `false`     | Sets or gets the selected state of this item                                                           |
-| `color`        | `'main' | 'error'`    | `main`      | Sets color. Possible options: <ul><li><code>main</code> (default)</li><li><code>error</code></li></ul> |
+| Name                    | Type                  | Default     | Description                                                                                            |
+| ----------------------- | --------------------- | ----------- | ------------------------------------------------------------------------------------------------------ |
+| `<ng-content>`          |                       |             | The content which is displayed inside of the item. This should only be text.                           |
+| `value`                 | `T | undefined`       | `undefined` | The associated value of this item                                                                      |
+| `disabled`              | `boolean | undefined` | `undefined` | Sets disabled state if property is set and the value is truthy or undefined                            |
+| `tabIndex`              | `number`              | `0`         | Sets and gets the tabIndex property                                                                    |
+| _deprecated_ `selected` | `boolean`             | `false`     | Sets or gets the selected state of this item                                                           |
+| `checked`               | `boolean`             | `false`     | Sets or gets the checked state of this item                                                            |
+| `color`                 | `'main' | 'error'`    | `main`      | Sets color. Possible options: <ul><li><code>main</code> (default)</li><li><code>error</code></li></ul> |
+
+The property `selected` is deprecated and will be renamed to `checked` in
+version 9.0.0
+
+## Button group item outputs
+
+| Name            | Type                                       | Description                                        |
+| --------------- | ------------------------------------------ | -------------------------------------------------- |
+| `checkedChange` | `event<DtButtonGroupItemCheckedChange<T>>` | Emits an event when the checked attribute changed. |
 
 ## Button group item methods
 
--- a/libs/barista-components/button-group/src/button-group.ts
+++ b/libs/barista-components/button-group/src/button-group.ts
@@ -83,7 +83,7 @@ export class DtButtonGroup<T> extends _DtButtonGroup
   set value(newValue: T | null) {
     if (this._value !== newValue) {
       this._value = newValue;
-      this._updateSelectedItemFromValue();
+      this._updateCheckedItemFromValue();
       this._changeDetectorRef.markForCheck();
     }
   }
@@ -114,11 +114,11 @@ export class DtButtonGroup<T> extends _DtButtonGroup
       // find if there is a selection
       // defer to next CD run - this is needed because we cannot update the item right away when there is no value set
       Promise.resolve().then(() => {
-        const selected = this._items.find((item) => item.selected);
+        const checked = this._items.find((item) => item.checked);
         if (this.value === null) {
-          this.value = selected ? selected.value : this._items.first.value;
+          this.value = checked ? checked.value : this._items.first.value;
         } else {
-          this._updateSelectedItemFromValue();
+          this._updateCheckedItemFromValue();
         }
       });
     }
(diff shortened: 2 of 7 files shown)


libs/barista-components/button-group/src/button-group.ts
```diff
@@ -194,6 +194,11 @@ export class DtButtonGroupItem<T> extends _DtButtonGroupItem
     HasTabIndex,
     AfterContentInit,
     OnDestroy {
+  /**
+   * @deprecated
+   * @breaking-change Will be renamed from 'selected' to 'checked' with v 9.0.0.
+   */
+  /* tslint:disable:no-unused-variable */
```
lukasholzer, line 201: the deprecation has to be on the `@Inpu`t there you deprecate a private property a private does not have to be deprecated.

lukasholzer (changes requested):

libs/barista-components/button-group/src/button-group.ts
```diff
@@ -203,7 +203,11 @@ export class DtButtonGroupItem<T> extends _DtButtonGroupItem
     DtButtonGroupItemSelectionChange<T>
   >();
 
-  /** Whether the button-group item is selected. */
+  /**
+   * @deprecated
+   * @breaking-change Will be renamed from 'selected' to 'checked' with v 9.0.0.
+   * Whether the button-group item is selected.
+   */
```
lukasholzer, line 210: please provide the new input for `checked` and add the behaviour so that when we remove this one it already works. so that they can already use the new version.

lukasholzer (changes requested):

libs/barista-components/button-group/src/button-group.ts
rowa-audil: We shouldn't do that as it changes the public API. Let's hold off with updating this in a minor version and fix it in version 9.

libs/barista-components/button-group/src/button-group.ts
```diff
@@ -194,24 +205,50 @@ export class DtButtonGroupItem<T> extends _DtButtonGroupItem
...
+
+  /**
+   * @deprecated
+   * @breaking-change Will be renamed from 'selected' to 'checked' with v 9.0.0.
+   * Use checked instead.
+   */
   @Input()
   get selected(): boolean {
-    return this._selected && !this.disabled;
+    return this._checked && !this.disabled;
```
lukasholzer, line 233:
```suggestion
    return this._checked;
```

libs/barista-components/button-group/src/button-group.ts
```diff
@@ -194,24 +205,50 @@ export class DtButtonGroupItem<T> extends _DtButtonGroupItem
...
   }
   set selected(value: boolean) {
     const newValue = coerceBooleanProperty(value);
-    if (this._selected !== newValue) {
-      this._selected = newValue;
+    if (this._checked !== newValue) {
+      this._checked = newValue;
+      this._changeDetectorRef.markForCheck();
+    }
+  }
```
lukasholzer, line 241:
```typescript
set selected(value: boolean) {
  this.checked(value);
}
```

libs/barista-components/button-group/src/button-group.ts
```diff
@@ -194,24 +205,50 @@ export class DtButtonGroupItem<T> extends _DtButtonGroupItem
...
-  private _selected = false;
+  private _checked = false;
   private _value: T;
   private _disabled = false;
 
-  /** Emits a stream when this item is selected or deselected. */
+  /**
+   * @deprecated
+   * @breaking-change will be renamed to checkedChange
+   * Emits a stream when this item is selected or deselected.
```
lukasholzer, line 215:
```suggestion
   * @breaking-change will be removed with version 9.0.0 please use the `checkedChange` instead.
```

libs/barista-components/button-group/src/button-group.ts
```diff
@@ -194,24 +205,50 @@ export class DtButtonGroupItem<T> extends _DtButtonGroupItem
...
-  /** Whether the button-group item is selected. */
+  /** Emits a stream when this item is checked or unchecked. */
+  @Output() readonly checkedChange = new EventEmitter<
+    DtButtonGroupItemCheckedChange<T>
+  >();
+
+  /**
+   * @deprecated
+   * @breaking-change Will be renamed from 'selected' to 'checked' with v 9.0.0.
+   * Use checked instead.
```
lukasholzer, line 229:
```suggestion
   * @breaking-change will be removed with version 9.0.0 please use the `checked` property instead.
```

libs/barista-components/button-group/src/button-group.ts
```diff
@@ -138,22 +139,32 @@ export class DtButtonGroup<T> extends _DtButtonGroup
...
       });
     }
   }
 }
 
-/** Change event object emitted by DtRadioButton */
+/**
+ * @deprecated
+ * @breaking-change Will be renamed to `DtButtonGroupItemCheckedChange`
+ * Change event object emitted by DtRadioButton
```
lukasholzer, line 155:
```suggestion
 * @breaking-change Will be removed with version 9.0.0. Please use the `DtButtonGroupItemCheckedChange` instead.
```

libs/barista-components/button-group/src/button-group.ts
```diff
@@ -264,24 +301,25 @@ export class DtButtonGroupItem<T> extends _DtButtonGroupItem
...
   }
 
-  /** @internal Selects this item. */
-  _onSelect(event: Event): void {
+  /** @internal Checks this item. */
+  _onChecked(event: Event): void {
     event.stopPropagation();
     const groupValueChanged =
       this._buttonGroup && this.value !== this._buttonGroup.value;
     this.selectionChange.emit({ source: this, value: this.value });
```
lukasholzer, line 309: NIT: Add a comment for removing with version 9.0.0
```
// TODO: Version 9.0.0 remove this event emission.
this.selectionChange.emit({ source: this, value: this.value });
```

libs/barista-components/button-group/src/button-group.ts
```diff
@@ -48,6 +48,7 @@ export const _DtButtonGroup = mixinTabIndex(
   class {} as Constructor<CanDisable>,
 );
 
+// ! TODO: Remove selected attribute and methods and use checked instead
```
lukasholzer, line 51:
```suggestion
```

lukasholzer (changes requested):

libs/barista-components/button-group/src/button-group.ts
```diff
@@ -194,24 +203,44 @@ export class DtButtonGroupItem<T> extends _DtButtonGroupItem
...
+  >();
+
+  /**
+   * @deprecated
+   * @breaking-change will be removed with version 9.0.0 please use the `checked` property instead.
+   */
   @Input()
   get selected(): boolean {
-    return this._selected && !this.disabled;
+    return this._checked;
```
lukasholzer, line 229: Please proxy the existing getter. Otherwise you wont have the full fledged check with the disabled

```suggestion
    return this.checked;
```

lukasholzer (changes requested):

libs/barista-components/button-group/README.md
```diff
@@ -54,14 +54,24 @@ To apply the button group component, use the `<dt-button-group>` and
...
+| `disabled`              | `boolean | undefined` | `undefined` | Sets disabled state if property is set and the value is truthy or undefined                            |
+| `tabIndex`              | `number`              | `0`         | Sets and gets the tabIndex property                                                                    |
+| _deprecated_ `selected` | `boolean`             | `false`     | Sets or gets the selected state of this item                                                           |
+| `checked`               | `boolean`             | `false`     | Sets or gets the checked state of this item                                                            |
+| `color`                 | `'main' | 'error'`    | `main`      | Sets color. Possible options: <ul><li><code>main</code> (default)</li><li><code>error</code></li></ul> |
+
+The property `selected` is deprecated and will be renamed to `checked` in
+version 9.0.0
+
+## Button group item inputs
```
peter-affenzeller, line 70:
```suggestion
## Button group item outputs
```

peter-affenzeller (changes requested):

peter-affenzeller (dismissed):

lukasholzer (approved):

lukasholzer:
@rowa-audil please rebase and squash your changes

lukasholzer (approved):

PR #1703, opened 2020-10-05
build(slider): Bazelify slider component.

SkaillZ:

Bazelify slider component.

#### Type of PR

Other


libs/barista-components/slider/src/slider.spec.ts
```diff
@@ -493,8 +493,7 @@ describe('DtSlider', () => {
       } = getElements(fixture);
       const { slider } = testComponent;
 
-      //<any, any> is needed because _updateSlider is private
-      const update = jest.spyOn<any, any>(slider, '_updateSlider');
+      const update = jest.spyOn(slider, '_updateSlider');
```
SkaillZ, line 496: This lead to errors when trying to build with Bazel

rowa-audil: We should fix this before we can merge this

peter-affenzeller, line 496: @SkaillZ said: This lead to errors when trying to build with Bazel

peter-affenzeller: it works this way:
```suggestion
      const update = jest.spyOn(slider as any, '_updateSlider');
```
then you don't have to change the class.

peter-affenzeller (changes requested):

lukasholzer (approved):

bardosmisi (dismissed):

rowa-audil (approved):

peter-affenzeller (approved):

bardosmisi (approved):

lukasholzer (approved):

PR #1716, opened 2020-10-07
fix(option): Fixes truncate issue in firefox

### LINKED ISSUE #1651: [filter-field] scrollbar truncates items (popup with scrollbar too narrow)
# Bug Report

## Expected Behavior

I expect the full item to be visible, even when scrolling

## Current Behavior

The last bit of the items are truncated, due to the scrollbar.

## Possible Solution

Increase the width of the popup to account for the scrollbar.

## Steps to Reproduce

1. Go to https://stackblitz.com/edit/github-hhfem2?file=src/filter-field/filter-field-default-example/filter-field-default-example.ts
2. Click the filter field
3. Observe the truncated items

1. Remove a few items to avoid the scrollbar
2. Observe no truncated items

## Context (Environment)

So far Firefox seems to be affected, but not Chromium.

When having many suggestions that start with the same string, but differ only in the last (truncated) part, it's impossible for the user to select the right item.

### Used Versions:
- **Firefox**: 81.0
- **@dynatrace/barista-components**: 8.0.0

## Attachments

truncated with scrollbar
image

fully visible without scrollbar
image

rowa-audil:

The autocomplete was behaving weirdly in firefox (fixes #1651)
The text in the option was cut off because the scrollbar is rendered identical to the scrollbar on mac maschines. It overlaps the content beneath it so it shrinks the option element. To accomadate, we added a firefox only selector adding 20px to the width of the autocomplete.
This will result in a wider autocomplete than before when there is no scrolling but solves the truncation of the options.

Thank you @bardosmisi for the support on this issue.

#### Type of PR

#### Checklist

### DIFF
--- a/libs/barista-components/autocomplete/src/autocomplete.scss
+++ b/libs/barista-components/autocomplete/src/autocomplete.scss
@@ -22,6 +22,13 @@ $dt-autocomplete-panel-min-width: 112px !default;
   -webkit-overflow-scrolling: touch; // for momentum scroll on mobile
 }
 
+@supports (-moz-appearance: none) {
+  .dt-autocomplete-panel {
+    width: calc(100% + 20px);
+    flex-shrink: 0; // to not shrink on FF if there is a scroll bar
+  }
+}
+
 .dt-autocomplete-hidden {
   display: none;
 }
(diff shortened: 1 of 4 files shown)


lukasholzer (approved):
LGTM thx for tackling it :)

SkaillZ (approved):

PR #1727, opened 2020-10-09
ci: Run CI on buildkite

andreas-doppelhofer:

This PR includes all relevant files for building barista on buildkite with bazel. 
Heart of the functionality is placed in  `baristacli.py` and the underlying configuration file `baristaPipeline.yml`.

In the `baristaPipeline.yml` you can specify all the relevant build steps as follows:

```
tasks:
  default:               # this is the buildkite queue label
    build:               # step which has to be called first
      trigger:           # define a list of all steps which should be called after, 
                         # here you define the order of all steps which should be called after the initil build step
        - test
      pre_cmd:           # call commands before the bazel call itself
        - echo "Pre command"
      bazel_flags:       # specify bazel flags
        - "--keep_going"
      bazel_cmd:
        - "//..."
      post_cmd:          # call commands after calling bazel 
        - echo "Post command"
    test:                # name of the test step
      disable: false     # disable the step setting it to true, otherwise false
      bazel_flags:
        - "--keep_going"
      bazel_cmd:
        - "//..."
##############################
### specify optional steps ###
##############################
    shell_command:           # name of a additional step
      pre_cmd:               # list of all commands which may be called before
      cmd:                   # list of all commands
        - bazel build //... 
        - bazel test //...
      post_cmd:              # list of all commands which may be called after
```

- buildkite configuration is generated with following step
`"source .buildkite/requirement.sh && .buildkite/baristacli.py pipeline| buildkite-agent pipeline upload"`
- commands on a node are called later with e.g.
`source .buildkite/requirement.sh && python3 .buildkite\baristacli.py exec --platform=default --target=build`
#### Type of PR

- see https://buildkite.com/dynatrace/bazel/builds/548

As a side note, build and tests are currently configured to run on a single node to be fast as possible (at that point in time).
We can switch this later on by specifying following in the yml and setting  `--multistage` in the call of our buildkite pipeline.
e.g. https://buildkite.com/dynatrace/bazel/builds/445
```
    test:
      parallel: 4
      disable: false
      bazel_flags:
        - "--keep_going"
      bazel_cmd:
        - "//..."
```

#### Checklist


.buildkite/baristacli.py
```diff
@@ -0,0 +1,569 @@
...
+        raise BuildKiteConfigError(
+            "{} is not available. Please specify one of {}".format(args.target,
+                                                                   stage.keys()))
+
+    # execute pre commands
+    if "pre_cmd" in stage[args.target]:
+        execute_shell_commands(stage[args.target]['pre_cmd'])
+
+    # bazel build info file
+    build_event_file = "build_event_file.json"
```
lukasholzer, line 124: for what do we need the build event prrotocol?

.buildkite/baristacli.py
```diff
@@ -0,0 +1,569 @@
...
+
+    if args.target == "test" and 'parallel' in stage[args.target] and int(
+            args.shard) != 0:
+        bazel_flags = stage[args.target]["bazel_flags"] if "bazel_flags" in stage[
+            args.target] else []
+        # add build event json
+        bazel_flags += ["--build_event_json_file={}".format(build_event_file)]
```
lukasholzer, line 131: I would add for every bazel command the `--profile=./profile.gz --record_full_profiler_data` build profile. and store it as gz. This is the holy grail if you want to investigate performance issues in the build. would be nice if we can store it as artifact to download it. you can upload the file to `chrome://tracing` and get insights in your build.

.buildkite/baristacli.py
```diff
@@ -0,0 +1,569 @@
...
+        execute_command(
+            [BAZELBIN] + [args.target.split("_")[0]] + test_sharding(
+                int(args.shard)), fail_if_nonzero=not in_ci())
+    else:
+        if 'bazel_cmd' in stage[args.target]:
+            bazel_flags = stage[args.target]["bazel_flags"] if "bazel_flags" in stage[
+                args.target] else []
+            # add build event json
+            bazel_flags += ["--build_event_json_file={}".format(build_event_file)]
```
lukasholzer, line 140: same here with profile\

.buildkite/baristacli.py
```diff
@@ -0,0 +1,569 @@
...
+            )
+        )
+        expanded_test_targets = expand_test_target_patterns(test_targets)
+        test_targets = get_targets_for_shard(expanded_test_targets, shard_id,
+                                             shard_count)
+
+    return test_targets
+
+
+def bubble_sort(array):
```
lukasholzer, line 305: haha nice best way to sort 👍

lukasholzer (approved):
Awesome! So nicely documented 👍 I see bazelci saved us a lot of time 😅

.buildkite/baristacli.py
andreas-doppelhofer: publish tests in buildkite if they fail

lukasholzer:
I think the buildcache setup is missing located under `tools/bazel_build_cache`.

.buildkite/baristacli.py
andreas-doppelhofer: image

.buildkite/baristacli.py
andreas-doppelhofer: done

lukasholzer (approved):

andreas-doppelhofer:
> I think the buildcache setup is missing located under `tools/bazel_build_cache`.

created followup https://github.com/dynatrace-oss/barista/issues/1731

lukasholzer (approved):
Awesome!

lukasholzer:
cherry-picked manually on `8.x` and `8.1.x`

PR #1743, opened 2020-10-15
fix(context-dialog): Focuses the first tabbable element inside the overlay con…

### LINKED ISSUE #1649: [dt-context-dialog] overlay in dt-context-dialog gets focused causing unfriendly experience
# Bug Report

When opening a `dt-context-dialog` with an overly in its content, the gets focussed even if disabled.
This causes unfriendly behaviour for the user especially if a dialog contains scrolled content. The last overlay gets focused causing the whole content to be scrolled to bottom.

## Expected Behavior

When opening a `dt-context-dialog` first DOM element gets the focus so when the content is scrolled it is scrolled to top.

## Current Behavior

As described in the bug report. The content is scrolled to bottom.
## Steps to Reproduce

Checkout https://stackblitz.com/edit/github-lkdc6c?file=src/context-dialog/context-dialog-default-example/context-dialog-default-example.html and see example code that reproduced the issue.

Also reproduced on newest Barista version.

rowa-audil:

Fixes a bug that caused the last tabbable item in an overlay context to be focused. The list of elements in the overlay was scrolled to the bottom as the focus is set to the last element.

_**Fixed**_ by focusing the first tabbable item

fixes #1649 

#### Type of PR

#### Checklist

### DIFF
--- a/libs/barista-components/context-dialog/src/context-dialog.html
+++ b/libs/barista-components/context-dialog/src/context-dialog.html
@@ -14,6 +14,11 @@
 </button>
 <ng-template>
   <div class="dt-context-dialog-panel" #panel role="dialog" dtTheme=":dark">
+    <ng-content select="dt-context-dialog-header"></ng-content>
+    <div class="dt-context-dialog-content" [ngClass]="overlayPanelClass">
+      <ng-content></ng-content>
+    </div>
+    <ng-content select="dt-context-dialog-footer"></ng-content>
     <button
       dt-icon-button
       variant="secondary"
@@ -24,10 +29,5 @@
     >
       <dt-icon name="abort"></dt-icon>
     </button>
-    <ng-content select="dt-context-dialog-header"></ng-content>
-    <div class="dt-context-dialog-content" [ngClass]="overlayPanelClass">
-      <ng-content></ng-content>
-    </div>
-    <ng-content select="dt-context-dialog-footer"></ng-content>
   </div>
 </ng-template>
--- a/libs/barista-components/context-dialog/src/context-dialog.ts
+++ b/libs/barista-components/context-dialog/src/context-dialog.ts
@@ -268,7 +268,7 @@ export class DtContextDialog extends _DtContextDialogMixinBase
         this._overlayRef.overlayElement,
       );
       this._focusTrap
-        .focusLastTabbableElementWhenReady()
+        .focusFirstTabbableElementWhenReady()
         .catch((error: Error) => {
           if (isDevMode()) {
             LOG.debug('Error when trying to set initial focus', error);


ffriedl89 (commented):
What is focused now initially?

rowa-audil:
@ffriedl89 The close button (X) of the overlay

ffriedl89:
> @ffriedl89 The close button (X) of the overlay

I know that initially we did not want to have the x focused right away, since the user's first action is not going to be closing the context dialog right away. Could you change the source order so that the x is not the first but the last element in the context dialog?

apps/components-e2e/src/components/context-dialog/context-dialog.po.ts
```diff
@@ -39,3 +39,14 @@ export const getActiveElementAriaLabel = ClientFunction(() => {
   }
   return element.getAttribute('aria-label') || '';
 });
+
+export const isCloseButtonActiveElement = ClientFunction(() => {
```
lukasholzer, line 43: NIT: please add jsdoc what you are doing there because it is not that describing when you see this selector `('.dt-context-dialog-trigger')`

libs/barista-components/context-dialog/src/context-dialog.ts
```diff
@@ -268,7 +268,7 @@ export class DtContextDialog extends _DtContextDialogMixinBase
         this._overlayRef.overlayElement,
       );
       this._focusTrap
-        .focusLastTabbableElementWhenReady()
+        .focusFirstTabbableElementWhenReady()
```
lukasholzer, line 271: I guess It worked then as expected :D

lukasholzer (approved):

libs/barista-components/context-dialog/src/context-dialog.html
```diff
@@ -24,10 +28,6 @@
     >
       <dt-icon name="abort"></dt-icon>
     </button>
-    <ng-content select="dt-context-dialog-header"></ng-content>
-    <div class="dt-context-dialog-content" [ngClass]="overlayPanelClass">
-      <ng-content></ng-content>
-    </div>
     <ng-content select="dt-context-dialog-footer"></ng-content>
```
ffriedl89, line 31: Please also move the footer up in the source order. All areas where the user could add content should be before the abort button.

ffriedl89 (changes requested):

ffriedl89 (dismissed):

ffriedl89:
Needs a rebase - please change scope from overlay to context-dialog

ffriedl89 (approved):

rowa-audil:
Mistakenly added a commit 😅

ffriedl89 (approved):

PR #1744, opened 2020-10-15
fix(option): Removes white-space: nowrap from the span element.

rowa-audil:

_**Hotfix**_: Fixes an unintended behaviour caused by #1716

#### Type of PR

#### Checklist


ffriedl89 (changes requested):
We need truncation with ellipsis on the option. An option must not be wrapping over 2 lines. Please add nowrap back and test with very long strings inside the option whether the ellipsis still work.

PR #1745, opened 2020-10-16
ci: Enable AWS bazel cache

andreas-doppelhofer:

<hr>
This enables builds which run on buildkite to use the AWS bazel cache to be as fast as possible.

#### Type of PR


.buildkite/build_cache/setup_ci_build_cache.sh
```diff
@@ -0,0 +1,41 @@
...
+set -e
+
+SCRIPT=$(readlink -f "$0")
+DIR=$(dirname "$SCRIPT")
+
+bazel_rc="${DIR}/../bazel.rc"
+
+PULL_BRANCH_REGEX="^pull\/[0-9]+$"
+
+# Forked pull requests have CIRCLE_BRANCH set to pull/XXX
```
ffriedl89, line 13: It might be a really dumb question - but how is buildkite connected to CIRCLE_BRANCH naming?

andreas-doppelhofer: True, this is a leftover. thx for the hint

lukasholzer (approved):

PR #1782, opened 2020-11-03
barista: Fix internal links not properly removed.

ffriedl89:

This fixes an issue that certain links were not properly removed.

### DIFF
--- a/.deployment/barista/public/Jenkinsfile
+++ b/.deployment/barista/public/Jenkinsfile
@@ -74,13 +74,7 @@ pipeline {
           sh '''
             npm run ng run barista-tools:build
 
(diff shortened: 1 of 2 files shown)


lukasholzer:
Seems like a cleanup issue on buildkite `npm WARN checkPermissions Missing write access to /var/lib/buildkite-agent/builds/designops-buildkite-i-09ada0a5ee130cdea-1/dynatrace/bazel/node_modules/@environments/barista-environment.d.ts`

Maybe @andreas-doppelhofer do you have a clue?

@ffriedl89 I think it is strange that the file that fails is the touched environments but it should not be located under `node_modules`

lukasholzer (approved):

PR #1800, opened 2020-11-09
fix(filter-field): Fixes a bug in the async example where the async d…

rowa-audil:

Fixes an issue where the loading indicator kept spinning by adding the data back to the original data source when a tag has been edited.

Fixes APM-266066

#### Type of PR

#### Checklist

### DIFF
--- a/libs/examples/src/filter-field/filter-field-async-example/filter-field-async-example.ts
+++ b/libs/examples/src/filter-field/filter-field-async-example/filter-field-async-example.ts
@@ -55,7 +55,10 @@ export class DtExampleFilterFieldAsync {
   currentFilterChanged(
     event: DtFilterFieldCurrentFilterChangeEvent<any>,
   ): void {
-    if (event.added[0] === this.DATA.autocomplete[0]) {
+    if (
+      event.added[0] === this.DATA.autocomplete[0] ||
+      this.ASYNC_DATA.autocomplete.find((name) => name === event.removed[0])
+    ) {
       // Emulate a http request
       setTimeout(() => {
         this._dataSource.data = this.ASYNC_DATA;


libs/examples/src/filter-field/filter-field-async-example/filter-field-async-example.ts
```diff
@@ -55,7 +55,12 @@ export class DtExampleFilterFieldAsync {
   currentFilterChanged(
     event: DtFilterFieldCurrentFilterChangeEvent<any>,
   ): void {
-    if (event.added[0] === this.DATA.autocomplete[0]) {
+    if (
+      event.added[0] === this.DATA.autocomplete[0] ||
+      this.ASYNC_DATA.autocomplete.find((name) => {
+        return name === event.removed[0];
```
nimrod13, line 61: Look good to me. I think you can even simplify it to one line with `find(name => name === event.removed[0])`  if you want.

nimrod13 (approved):

libs/examples/src/filter-field/filter-field-async-example/filter-field-async-example.ts
rowa-audil: It threw an error originally but that was due to me adding the curly brackets by mistake. Thank you!

nimrod13: Gladly :)

nimrod13 (approved):

lukasholzer (approved):

ffriedl89 (changes requested):
Code change looks good. Please rebase to chore scope please. This is not a fix that should show up in the changelog, since it only enhances an example. And please also reach out to the colleague to explain what was missing to get it working. 
The important part here is that the filter field does not reset the the previously provided autocomplete options if the autocomplete is marked as async. It is the same behavior as on the first selection where the filter field waits (and the loading spinner spins) until new data is provided.

tomheller:
lgtm, please rebase though.

nimrod13 (approved):

tomheller (approved):

thomaspink (approved):

bardosmisi (approved):

PR #1808, opened 2020-11-10
chore: Internal link check in jenkins pipeline

rowa-audil:

Fixes APM-266099

Thank you @tomheller for the help on this issue!

#### Type of PR

### DIFF
--- a/tools/scripts/internal-link-checker.js
+++ b/tools/scripts/internal-link-checker.js
@@ -1,0 +1,32 @@
+const { execSync } = require('child_process');
+const dotenv = require('dotenv');
+
+dotenv.config();
+
+const internalLinks = process.env.INTERNAL_LINKS.split(',');
+
+if (internalLinks.length === 0) {
+  console.log(
+    'No internal link configuration was found, please make sure to provide one in the INTERNAL_LINKS environment variable',
+  );
+  process.exit(1);
+}
+
+const results = [];
+
+for (const internalLink of internalLinks) {
+  const execSyncResult = execSync(
+    `grep --include="./dist/**.json" -rl "./dist" -e "${internalLink}"`,
+  ).toString();
+
+  if (execSyncResult.length > 0) {
+    results.push(execSyncResult);
+  }
+}
+
+if (results.length > 0) {
+  console.log(`Failed because internal links were found in files:\n${results}`);
+  process.exit(1);
+}
+
+process.exit(0);
--- a/.deployment/barista/public/Jenkinsfile
+++ b/.deployment/barista/public/Jenkinsfile
@@ -83,6 +83,16 @@ pipeline {
               ./node_modules/.bin/ng run barista-design-system:static-pages:production
           '''
         }
+
+        configFileProvider([
+          configFile(fileId: "barista-env", variable: "ENV")
+        ]) {
+          sh '''
+            mv $ENV .env
+
+            node tools/scripts/internal-link-checker.js
+          '''
+        }
       }
     }
 


tools/scripts/internal-link-checker.js
```diff
@@ -0,0 +1,22 @@
+const { execSync } = require('child_process');
+const dotenv = require('dotenv');
+
+dotenv.config();
+
+function run() {
```
lukasholzer, line 6: NIT: no need for a function

lukasholzer (dismissed):

tools/scripts/internal-link-checker.js
```diff
@@ -0,0 +1,18 @@
+const { execSync } = require('child_process');
+const dotenv = require('dotenv');
+
+dotenv.config();
+
+const internalLinks = process.env.INTERNAL_LINKS.split(',');
```
ffriedl89, line 6: Please add process.exit(1) in the case when internalLinks is empty. We always need to remove internal links - this must not be empty.

ffriedl89 (changes requested):
Overall a very nice solution. Great teamwork! Only a small change needed

tools/scripts/internal-link-checker.js
```diff
@@ -0,0 +1,23 @@
...
+  );
+  if (result.toString().length > 0) {
+    console.log(
+      `Failed because internal links were found in files:\n${result.toString()}`,
+    );
+    process.exit(1);
+  } else {
+    console.log('No internal links were found.');
+    process.exit(1);
+  }
```
tomheller, line 20: This will always break the CI pipeline. 
Please remove the else condition. it is unnecessary.

tomheller (changes requested):
Please add a check before the loop, if the constant `internalLinks` has links in them. If the _config_ is empty, please also exit the process with 1, to prevent silent passing without an error.

ffriedl89 (changes requested):
If you have a closer look again this program can never exit with exit code 0. It will always fail. The issue is that we need to check whether internalLinks is an empty array. Consider that internalLinks must always be an array with entries.

rowa-audil:
Sorry about that. Thank you for pointing that out @tomheller and @ffriedl89.

tomheller:
No problem, happens to the best. 😉

tools/scripts/internal-link-checker.js
```diff
@@ -0,0 +1,24 @@
+const { execSync } = require('child_process');
+const dotenv = require('dotenv');
+
+dotenv.config();
+
+const internalLinks = process.env.INTERNAL_LINKS.split(',');
+
+if (internalLinks.length === 0) {
+  process.exit(1);
```
tomheller, line 9:
```suggestion
  console.log('No internal link configuration was found, please make sure to provide one in the INTERNAL_LINKS environment variable');
  process.exit(1);
```

tools/scripts/internal-link-checker.js
```diff
@@ -0,0 +1,24 @@
...
+
+for (const internallink of internalLinks) {
+  const result = execSync(
+    `grep --include="./dist/**.json" -rl "./dist" -e "${internallink}"`,
+  );
+  if (result.toString().length > 0) {
+    console.log(
+      `Failed because internal links were found in files:\n${result.toString()}`,
+    );
+    process.exit(1);
```
tomheller, line 20: If there are multiple internal link domains configured, and it detects the first one, it will fail already. 
After fixing the first one it might run into another error for the second match in a consecutive CI run. 

I'm not saying we need to improve this, but it could be handled by collecting the results in an array outside of the loop, and having a check after the loop if the result array contains errors.
This way every domain match violation would be caught in a single run and can be fixed as one as well.

tomheller (approved):

ffriedl89 (approved):
LGTM

tomheller (approved):
Looks good to me!

lukasholzer (approved):

PR #1809, opened 2020-11-11
fix(filter-field): Fixes an issue with async data and programmatically set filters.

tomheller:
When setting the filters programmatically, as the quickfilter does when a filter is set, the
retreival of set filters breaks.
The programmatic setting of the filters runs a function called `tryApplyFilters`, which looks in the
datasource and async dataset for the correct filters.
For some reason, the root of the async def was pushed twice, resulting in a wrong path and a wrong
id.

@thomaspink do you remember the reasoning on why the `def` and `asyncDef` where pushed to the array in the partial async case.
Event with the change in this PR, the partial async example behaves the same and all tests are passing.

Fixes APM-267380

#### Type of PR

Bugfix

#### Checklist

### DIFF
--- a/apps/components-e2e/src/components/quick-filter/quick-filter.e2e.ts
+++ b/apps/components-e2e/src/components/quick-filter/quick-filter.e2e.ts
@@ -21,6 +21,7 @@ import {
   tagDeleteButton,
   clickOption,
   filterFieldRangePanel,
+  options,
 } from '../filter-field/filter-field.po';
 import {
   getGroupItem,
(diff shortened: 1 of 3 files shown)


rowa-audil (dismissed):

rowa-audil (approved):

apps/components-e2e/src/components/quick-filter/quick-filter.e2e.ts
```diff
@@ -302,3 +303,49 @@ test('should check the show more with distinct values', async (testController: T
...
 });
+
+fixture('Quick Filter async')
+  .page('http://localhost:4200/quick-filter/async')
+  .beforeEach(async () => {
+    await resetWindowSizeToDefault();
+    await waitForAngular();
+  });
+
+test('should work with async data and handle distincts correctly', async (testController: TestController) => {
```
ffriedl89, line 314: I might be getting confused, but was the issue visible with this runthrough? I thought you had to select async three times and on the third time Linz was in the options list although it shouldn't have been? 

I know that the ids were already faulty, but I am not sure if the user already saw the issue with selecting a non async autocomplete in between. 
Sorry if i a can't remember right now.

tomheller: It was quite confusing, but yes the issue was visible with this runthrough. 
When selecting the first async filter completely, then choosing the same async filter again, we saw that the first filter was picked up correctly (ids were still fine). So after `Aut (async) Linz`, picking `Aut (async)` gave you the correct options of `Vienna` and `Graz`. 

When choosing a different filter after the first async, it showed the issue right away. So after `Aut (async) Linz`, picking `USA San Francisco` and then `Aut (async)` it gave the wrong options of `Linz`, `Vienna` and `Graz` whereas `Linz` should have been filtered out because it previously has been set.

thomaspink (approved):

PR #1812, opened 2020-11-13
test(combobox): Skip flaky test in the combobox.

tomheller:
Skip the flaky test for now, the combobox is still in experimental and we want to have a stable e2e pipeline right now. 

#### Type of PR

Other

#### Checklist

### DIFF
--- a/apps/components-e2e/src/components/combobox/combobox.ts
+++ b/apps/components-e2e/src/components/combobox/combobox.ts
@@ -74,7 +74,7 @@ export class DtE2ECombobox {
       this._changeDetectorRef.markForCheck();
     }
 
-    timer(1500)
+    timer(1000)
       .pipe(take(1))
       .subscribe(() => {
         this._options = allOptions.filter(optionFilter(event.filter));


apps/components-e2e/src/components/combobox/combobox.e2e.ts
```diff
@@ -97,7 +97,8 @@ test('should filter the correct values and not rewrite the input', async (testCo
     .contains('Value 2');
 });
 
-test('should not reset the value when hitting continuously typing the backspace key', async (testController: TestController) => {
+// tslint:disable-next-line: dt-no-focused-tests
+test.skip('should not reset the value when hitting continuously typing the backspace key', async (testController: TestController) => {
```
ffriedl89, line 101: I am perfectly fine with the skip here, one idea maybe worth trying is to have a different timing in the combobox.ts in the e2e app than the wait time in the test. 
The wait here and the timer in 
https://github.com/dynatrace-oss/barista/blob/master/apps/components-e2e/src/components/combobox/combobox.ts#L77

both currently use 1500ms for the delay.
Reduce the wait time in the combobox.ts might make it less flaky

ffriedl89 (approved):
I am fine with skipping, an idea would be to try to increase the wait time here or decrease the number that the example waits

lukasholzer (dismissed):

apps/components-e2e/src/components/combobox/combobox.e2e.ts
tomheller: Thanks fabian, that helped. 
I will reword the commit to match the solution.

ffriedl89 (approved):

PR #1823, opened 2020-11-17
refactor(switch): Custom properties.

bardosmisi:

Fixes the switch part of APM-266103.

#### Type of PR


libs/barista-components/switch/src/switch.scss
```diff
@@ -3,7 +3,50 @@
 @import '../../core/src/style/transition';
 @import '../../core/src/theming/theming';
```
tomheller, line 4: We can also drop this import, as we no longer need to call the `dt-apply-theme()` mixin.
```suggestion
```

libs/barista-components/switch/src/switch.scss
```diff
@@ -3,7 +3,50 @@
...
 @import '../../style/font-mixins';
-@import './switch-theme';
+
+:host {
+  --content-color: #{$default-font-color};
+  --dt-switch-bar-background-radial: #{$gray-700};
+  --dt-switch-bar-border-color: #{$turquoise-600};
+  --dt-switch-bar-bg-border-color: transparent;
```
tomheller, line 11: Please keep abbreviations consistent. You are sometimes using `bg` and sometimes `background`. 
I understand that some of these variables could get quite long, I would opt for `background`.
```suggestion
  --dt-switch-bar-background-border-color: transparent;
```

peter-affenzeller (approved):
except for the inconsistent abbreviations mentioned by tom - lgtm 🐰

libs/barista-components/switch/src/switch.scss
bardosmisi: Yes, I was thinking about that during implementation, but now I think ```background``` is better.

libs/barista-components/switch/src/switch.scss
bardosmisi: Removed, thanks 👍

tomheller (approved):

lukasholzer (approved):

tomheller:
IN the past we have considered these features. But I'm fine with keeping it as a refactor.

PR #1824, opened 2020-11-18
build(examples): Bazelify Barista examples

SkaillZ:

Adds Bazel build targets to all examples.

#### Type of PR

Other


libs/examples/src/filter-field/BUILD.bazel
```diff
@@ -1,13 +1,32 @@
...
-package(default_visibility = ["//:__subpackages__"])
+package(default_visibility = ["//visibility:public"])
 
 ng_module(
-    name = "filter-field",
-    srcs = glob(["**/*.ts"]),
-    angular_assets = glob(["**/*.html"]),
-    module_name = "@dynatrace/examples/filter-field",
-    tsconfig = "//libs/examples:tsconfig-lib",
+    name = "compile",
```
tomheller, line 6: this name is referenced in the e2e test app, which is why the build now fails on buildkite. 
We might need to change the reference in the e2e app 

https://github.com/dynatrace-oss/barista/blob/a59f8815b59c71b9e95f6a9e14d8b3ac4f762c49/apps/components-e2e/src/components/filter-field/BUILD.bazel#L18

libs/examples/src/quick-filter/BUILD.bazel
```diff
@@ -1,13 +1,22 @@
...
-package(default_visibility = ["//:__subpackages__"])
+package(default_visibility = ["//visibility:public"])
 
 ng_module(
-    name = "quick-filter",
-    srcs = glob(["**/*.ts"]),
-    angular_assets = glob(["**/*.html"]),
-    module_name = "@dynatrace/examples/quick-filter",
-    tsconfig = "//libs/examples:tsconfig-lib",
+    name = "compile",
```
tomheller, line 6: This name is referenced in the e2e test app, we might need to update it there.

https://github.com/dynatrace-oss/barista/blob/a59f8815b59c71b9e95f6a9e14d8b3ac4f762c49/apps/components-e2e/src/components/quick-filter/BUILD.bazel#L23

tomheller (changes requested):

libs/examples/src/filter-field/BUILD.bazel
lukasholzer: That's why the ci is failing :)

lukasholzer (changes requested):
Please name the `ng_module` like the folder to have shorter targets later on.

For example: `ng_module( name = "alert" ...) ` instead of compile.

lukasholzer (changes requested):
CI is still failing:

image

tomheller (approved):

lukasholzer (approved):

PR #1826, opened 2020-11-19
fix(filter-field): Fixes an issue with confusing disabled styles.

tomheller:

Fixes APM-269127

image

Improves the contrast ratio and differentiation of the various states of a tag. Here is the comparison table what is set for the example. Free text examples are added additionally, as the styling of the quotes is sometimes different.

| Tag | Editable | Deletable |
| --- |:---:|:---:| 
| AUT: Linz | false | false |
| AUT: Vienna | true | false |
| AUT: Graz | false | true |
| AUT: Bregenz | true | true |
| USA: Miami |false |false |
| USA: Austin | true| false|
| USA: San Diego |false |true  |
| USA: Bend |true | true|

Two major things that have changed. If a tag is whether deletable and editable, it is considered as disabled. The `X` button on the disabled tag was removed completely. 
Colors of key and value elements have been adjusted to gray / turquoise depending on the fact if they are interactive or not.

#### Type of PR

Bugfix

#### Checklist

### DIFF
--- a/libs/barista-components/filter-field/src/filter-field-tag/filter-field-tag.html
+++ b/libs/barista-components/filter-field/src/filter-field-tag/filter-field-tag.html
@@ -31,6 +31,7 @@
   </button>
 </div>
 <button
+  *ngIf="!(_filterFieldDisabled || disabled) && deletable"
   dt-icon-button
   variant="nested"
   class="dt-filter-field-tag-button"
--- a/libs/barista-components/filter-field/src/filter-field-tag/filter-field-tag.scss
+++ b/libs/barista-components/filter-field/src/filter-field-tag/filter-field-tag.scss
@@ -26,7 +26,18 @@
   &.dt-filter-field-tag-disabled {
     border-color: $gray-200;
     background-color: #ffffff;
-    color: $gray-300;
+
+    .dt-filter-field-tag-key,
+    .dt-filter-field-tag-value {
+      color: $gray-500;
+    }
+  }
+
+  &.dt-filter-field-tag-read-only {
+    .dt-filter-field-tag-key,
+    .dt-filter-field-tag-value {
+      color: $gray-500;
+    }
   }
 }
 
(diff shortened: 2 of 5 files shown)


kathrin-aigner:
lgtm!

lukasholzer (dismissed):

ffriedl89:
I've one question: When the tag is "fully" disabled we do not show the closing icon, but when the deletion is disabled it is shown but disabled. Does it make sense to unify this - i like the approach to hide it - since it reduces clutter in the filter field. I'd go for removing it entirely whenever deletion is not possible. @tomheller @kathrin-aigner

kathrin-aigner:
> I've one question: When the tag is "fully" disabled we do not show the closing icon, but when the deletion is disabled it is shown but disabled. Does it make sense to unify this? - I like the approach to hide it - since it reduces clutter in the filter field. I'd go for removing it entirely whenever deletion is not possible. @tomheller @kathrin-aigner


@ffriedl89 I agree, it does not make sense to show the close icon as disabled. I would hide it as well.

tomheller:
> @ffriedl89 I agree, it does not make sense to show the close icon as disabled. I would hide it as well.

I disagree. In case of the disabled tag, there is no interaction possible at all. The user cannot click / edit the tag. In case of `deletable: false`, the tag can still be interacted with. The value can change, but the tag itself can never be fully deleted. 
Without the _disabled X_ Button, the user would have no indication on why the deletion is not possible. 
Especially if there are no other tags present, which can be deleted (which would make the difference more visible).

kathrin-aigner:
> > @ffriedl89 I agree, it does not make sense to show the close icon as disabled. I would hide it as well.
> 
> @kathrin-aigner I disagree. In case of the disabled tag, there is no interaction possible at all. The user cannot click / edit the tag. In case of `deletable: false`, the tag can still be interacted with. The value can change, but the tag itself can never be fully deleted.
> Without the _disabled X_ Button, the user would have no indication on why the deletion is not possible.
> Especially if there are no other tags present, which can be deleted (which would make the difference more visible).

@tomheller As discussed I still suggest hiding the close button because the delet icon gives no additional value to the user. It can cause more confusion because people will have to interpret why they can't interact with that element. Also when you look at accessibility it does not make sense to show the disabled delete button because it cannot be focused and the screen reader users would get less information because the element would be skipped anyways.

tomheller:
Updated the behavior to reflect requested changes. 
image

ffriedl89 (approved):
LGTM!

PR #1836, opened 2020-11-25
feat(filter-field): Add multiselect

subarroca:
#### Type of PR

Added multi select panel for filter field

#### Checklist

### DIFF
--- a/apps/components-e2e/src/components/filter-field/filter-field.e2e.ts
+++ b/apps/components-e2e/src/components/filter-field/filter-field.e2e.ts
@@ -28,9 +28,13 @@ import {
   tagOverlay,
   setupSecondTestScenario,
   filterFieldRangePanel,
+  clickMultiSelectOption,
+  multiSelectApply,
+  multiSelectPanel,
+  switchToSecondDatasource,
 } from './filter-field.po';
 import { Selector } from 'testcafe';
-import { waitForAngular, resetWindowSizeToDefault } from '../../utils';
+import { resetWindowSizeToDefault, waitForAngular } from '../../utils';
 
 fixture('Filter Field')
   .page('http://localhost:4200/filter-field')
@@ -318,5 +322,136 @@ test('should close the range when blurring the filter field mid filter, returnin
     .pressKey('backspace')
     .wait(250)
     .expect(filterFieldRangePanel.exists)
+    .notOk();
+});
+
+test('should choose a multiselect node with the keyboard and submit the correct value', async (testController: TestController) => {
+  await testController
+    .click(switchToSecondDatasource)
+    // Wait for the filterfield to catch up.
+    .wait(500);
+
+  // Focus the filter field.
+  await focusFilterFieldInput();
+
+  // Select the test autocomplete
+  await testController
+    // Select the multiselect node
+    .pressKey('down down down enter')
+    // Wait for a certain amout fo time to let the filterfield refresh
+    .wait(250)
+    // Select the desired option
+    .pressKey('down space enter')
+    // Wait for a certain amout fo time to let the filterfield refresh
+    .wait(250);
+
+  const tags = await getFilterfieldTags();
+
+  await testController
+    .expect(tags.length)
+    .eql(1)
+    .expect(tags[0])
+    .eql('SeasoningKetchup');
+});
+
+test('should not apply an empty multiselect node with the keyboard', async (testController: TestController) => {
+  await testController
+    .click(switchToSecondDatasource)
+    // Wait for the filterfield to catch up.
+    .wait(500);
+
+  // Focus the filter field.
+  await focusFilterFieldInput();
+
+  // Select the test autocomplete
+  await testController
+    // Select the multiselect node
+    .pressKey('down down down enter')
+    // Wait for a certain amout fo time to let the filterfield refresh
+    .wait(250)
+    // Try to apply without any option
+    .pressKey('enter')
+    // Wait for a certain amout fo time to let the filterfield refresh
+    .wait(250);
+
+  const tags = await getFilterfieldTags();
+
+  await testController.expect(tags.length).eql(0);
+});
+
+test('should choose a multiselect node with the mouse and submit the correct value immediately', async (testController: TestController) => {
+  // Switch to second datasource.
+  await testController
+    .click(switchToSecondDatasource)
+    // Wait for the filterfield to catch up.
+    .wait(500);
+
+  // Select the multiSelect
+  await clickOption(4)
+    // Wait for a certain amount of time to let the filterfield refresh
+    .wait(250);
+
+  // Select the multiSelect option
+  await clickMultiSelectOption(3).wait(250);
+
+  // Submit the value immediately
+  await testController.click(multiSelectApply).wait(250);
+
+  const tags = await getFilterfieldTags();
+
+  await testController
+    .expect(tags.length)
+    .eql(1)
+    .expect(tags[0])
+    .match(/SeasoningMustard/);
+});
+
+test('should not apply an empty multiselect node with the mouse', async (testController: TestController) => {
+  // Switch to second datasource.
+  await testController
+    .click(switchToSecondDatasource)
+    // Wait for the filterfield to catch up.
+    .wait(500);
+
+  // Select the multiSelect
+  await clickOption(4)
+    // Wait for a certain amount of time to let the filterfield refresh
+    .wait(250);
+
+  // Submit the value immediately
+  await testController.click(multiSelectApply).wait(250);
+
+  // panel should not have closed
+  await testController.expect(multiSelectPanel.exists).ok();
+});
+
+test('should close the multiselect when blurring the filter field mid filter, returning back and deleting the current filter', async (testController: TestController) => {
+  // Switch to second datasource.
+  await testController
+    .click(switchToSecondDatasource)
+    // Wait for the filterfield to catch up.
+    .wait(500);
+
+  // Select the multiSelect
+  await clickOption(4)
+    // Wait for a certain amount of time to let the filterfield refresh
+    .wait(250);
+
+  await testController
+    .click(Selector('.outside'))
+    .expect(multiSelectPanel.exists)
+    .notOk();
+
+  // Focus the filter field again,
+  // press backspace - deleting the current filter
+  // the multiSelect should now be closed.
+  await focusFilterFieldInput();
+  await testController
+    .expect(multiSelectPanel.exists)
+    .ok()
+    .wait(250)
+    .pressKey('backspace')
+    .wait(250)
+    .expect(multiSelectPanel.exists)
     .notOk();
 });
--- a/libs/barista-components/filter-field/README.md
+++ b/libs/barista-components/filter-field/README.md
@@ -83,11 +83,12 @@ property, when replacing the datasource with the partially loaded data.
 
 <ba-live-example name="DtExampleFilterFieldPartial"></ba-live-example>
 
-## Unique free-text or range options
+## Unique free-text, range or multiselect options
 
 It is possible to set a free-text or range option to be unique. So it can only
-be added once regardless of the value the user added to the free-text input
-field or into the range inputs.
+be added once regardless of the value the user added to the input field.
+
+Multiselection is unique by default.
 
 <ba-live-example name="DtExampleFilterFieldUnique"></ba-live-example>
 
@@ -103,6 +104,10 @@ Every filter is an array of:
   user. This includes the range operator (`DtFilterFieldRangeOperator`), the
   unit (`string`) of the selected values and the range itself (which can either
   be one `number` value or a `number` tuple).
+- Multi select filters: array of objects that have been selected via a multi
+  select list with checkboxes (the objects are exactly the ones the consumer has
+  provided via the data source). The suggested items can also be refined by
+  typing.
 
 ### Receiving the selected filters
 
@@ -186,10 +191,19 @@ node objects:
 | `isDtOptionDef`       | Whether the provided definition object is of type `NodeDef` and consists of an option definition.                      |
 | `dtGroupDef`          | Creates a node definition object or extends one and applies a group definition object based on the parameters.         |
 | `isDtGroupDef`        | Whether the provided definition object is of type `NodeDef` and consists of a group definition.                        |
+| `dtMultiSelectDef`    | Creates a node definition object or extends one and applies a multi select definition object based on the parameters.  |
+| `isDtMultiSelectDef`  | Whether the provided definition object is of type `NodeDef` and consists of a multi select definition.                 |
 
 ## Behavior
 
 <ba-ux-snippet name="filter-field-behavior"></ba-ux-snippet>
+
+### Multi selection of filter values
+
+Multi select lets users choose more than one value for the same field. The
+result is an array of the same objects the user has passed in the data source.
+
+<ba-live-example name="DtExampleFilterFieldMultiSelect"></ba-live-example>
 
 ### Clear filters
 
@@ -243,10 +257,6 @@ angular validator interface.
 ### Handling operators (not yet implemented)
 
 <ba-ux-snippet name="filter-field-operators"></ba-ux-snippet>
-
-### Multiselection of filter values (not yet implemented)
-
-<ba-ux-snippet name="filter-field-multiselection"></ba-ux-snippet>
 
 ### Help and errors (not yet implemented)
 
--- a/libs/barista-components/filter-field/src/filter-field-module.ts
+++ b/libs/barista-components/filter-field/src/filter-field-module.ts
@@ -19,14 +19,17 @@ import { NgModule } from '@angular/core';
 
 import { DtAutocompleteModule } from '@dynatrace/barista-components/autocomplete';
 import { DtButtonModule } from '@dynatrace/barista-components/button';
-import { DtButtonGroupModule } from '@dynatrace/barista-components/button-group';
-import { DtOptionModule } from '@dynatrace/barista-components/core';
 import { DtIconModule } from '@dynatrace/barista-components/icon';
+import { DtButtonGroupModule } from '@dynatrace/barista-components/button-group';
+import { DtCheckboxModule } from '@dynatrace/barista-components/checkbox';
+import { DtOptionModule } from '@dynatrace/barista-components/core';
 import { DtInputModule } from '@dynatrace/barista-components/input';
 import { DtLoadingDistractorModule } from '@dynatrace/barista-components/loading-distractor';
 import { DtOverlayModule } from '@dynatrace/barista-components/overlay';
 
 import { DtFilterField } from './filter-field';
+import { DtFilterFieldMultiSelect } from './filter-field-multi-select/filter-field-multi-select';
+import { DtFilterFieldMultiSelectTrigger } from './filter-field-multi-select/filter-field-multi-select-trigger';
 import { DtFilterFieldRange } from './filter-field-range/filter-field-range';
 import { DtFilterFieldRangeTrigger } from './filter-field-range/filter-field-range-trigger';
 import { DtFilterFieldTag } from './filter-field-tag/filter-field-tag';
@@ -44,6 +47,7 @@ import { DtHighlightModule } from '@dynatrace/barista-components/highlight';
     DtLoadingDistractorModule,
     DtOverlayModule,
     DtHighlightModule,
+    DtCheckboxModule,
   ],
   exports: [
     DtAutocompleteModule,
@@ -56,6 +60,8 @@ import { DtHighlightModule } from '@dynatrace/barista-components/highlight';
     DtFilterFieldTag,
     DtFilterFieldRange,
     DtFilterFieldRangeTrigger,
+    DtFilterFieldMultiSelect,
+    DtFilterFieldMultiSelectTrigger,
   ],
 })
 export class DtFilterFieldModule {}
--- a/libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select-trigger.ts
+++ b/libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select-trigger.ts
@@ -1,0 +1,185 @@
+/**
+ * @license
+ * Copyright 2020 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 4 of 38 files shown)


apps/components-e2e/src/components/filter-field/filter-field.e2e.ts
```diff
@@ -17,20 +17,24 @@
...
   setupSecondTestScenario,
+  switchToSecondDatasource,
+  clickMultiSelectOption,
+  input,
+  errorBox,
+  tagOverlay,
   filterFieldRangePanel,
 } from './filter-field.po';
-import { Selector } from 'testcafe';
-import { waitForAngular, resetWindowSizeToDefault } from '../../utils';
```
tomheller, line 33: This imports have primarily resorted. It would be great to disable automatic import sorting. These changesets are really hard to grasp during a review. Maybe you can clean them up?

libs/barista-components/filter-field/README.md
```diff
@@ -101,6 +101,10 @@ Every filter is an array of:
   user. This includes the range operator (`DtFilterFieldRangeOperator`), the
   unit (`string`) of the selected values and the range itself (which can either
   be one `number` value or a `number` tuple).
+- Multi select filters: array of objects that have been selected via an multi
+  selct list with checkboxes (the objects are exactly the ones the consumer has
```
tomheller, line 105:
```suggestion
  select list with checkboxes (the objects are exactly the ones the consumer has
```

libs/barista-components/filter-field/README.md
```diff
@@ -184,11 +188,20 @@ node objects:
...
+| `dtMultiSelectDef`    | Creates a node definition object or extends one and applies a multi select definition object based on the parameters.  |
+| `isDtMultiSelectDef`  | Whether the provided definition object is of type `NodeDef` and consists of a multi select definition.                 |
 
 ## Behavior
 
 <ba-ux-snippet name="filter-field-behavior"></ba-ux-snippet>
 
+### Multi selection of filter values
+
+Multi select let's users choose more than one value for the same field. The
```
tomheller, line 200:
```suggestion
Multi select lets users choose more than one value for the same field. The
```

libs/barista-components/filter-field/src/filter-field-module.ts
```diff
@@ -16,34 +16,36 @@
...
 import { DtLoadingDistractorModule } from '@dynatrace/barista-components/loading-distractor';
 import { DtOverlayModule } from '@dynatrace/barista-components/overlay';
-
 import { DtFilterField } from './filter-field';
+import { DtFilterFieldMultiSelect } from './filter-field-multi-select/filter-field-multi-select';
+import { DtFilterFieldMultiSelectTrigger } from './filter-field-multi-select/filter-field-multi-select-trigger';
 import { DtFilterFieldRange } from './filter-field-range/filter-field-range';
 import { DtFilterFieldRangeTrigger } from './filter-field-range/filter-field-range-trigger';
 import { DtFilterFieldTag } from './filter-field-tag/filter-field-tag';
-import { DtHighlightModule } from '@dynatrace/barista-components/highlight';
```
tomheller, line 33: Please consider resetting the import sort.

libs/barista-components/filter-field/src/filter-field-module.ts
```diff
@@ -16,34 +16,36 @@
...
     DtButtonGroupModule,
+    DtButtonModule,
+    DtCheckboxModule,
+    DtHighlightModule,
+    DtIconModule,
+    DtInputModule,
     DtLoadingDistractorModule,
+    DtOptionModule,
     DtOverlayModule,
-    DtHighlightModule,
```
tomheller, line 46: Please reset the order to what it was, these sorts of changesets are really hard to decode.

libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select-trigger.ts
```diff
@@ -0,0 +1,758 @@
...
+        this._scrollToOption();
+      }
+    }
+  }
+
+  /** Determines whether the panel can be opened. */
+  // private _canOpen(): boolean {
+  //   const element = this._elementRef.nativeElement;
+  //   return !element.readOnly && !element.disabled && !this._multiSelectDisabled;
+  // }
```
tomheller, line 452: Code is duplicated below.
```suggestion
```

libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select-trigger.ts
```diff
@@ -0,0 +1,758 @@
...
+  /**
+   * Strategy that is used to position the panel.
+   */
+  private _positionStrategy: DtFlexibleConnectedPositionStrategy;
+
+  /** Stream of keyboard events that can close the panel. */
+  private readonly _closeKeyEventStream = new Subject<void>();
+
+  /** Subscription to viewport size changes. */
+  // private _viewportSubscription = EMPTY.subscribe();
```
tomheller, line 242: remove unused code.
```suggestion
```

libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select-trigger.ts
```diff
@@ -0,0 +1,758 @@
...
+    return this._multiSelect;
+  }
+  set multiSelect(value: DtFilterFieldMultiSelect<T>) {
+    this._multiSelect = value;
+    this._detachOverlay();
+  }
+  private _multiSelect: DtFilterFieldMultiSelect<T>;
+
+  /** autocomplete` attribute to be set on the input element. */
+  @Input('autocomplete') autocompleteAttribute = 'off';
```
tomheller, line 130: Which autocomplete is this. I don't see why the multi select trigger should require an autocomplete. Maybe you can explain this a bit. 
Imho this trigger does not need a lot more functionality than the range-trigger.

libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select-trigger.ts
```diff
@@ -0,0 +1,758 @@
...
+   * Reference relative to which to position the autocomplete panel.
+   * Defaults to the autocomplete trigger element.
+   */
+  @Input('dtMultiSelectConnectedTo') connectedTo: DtAutocompleteOrigin;
+
+  /**
+   * Reference relative to which to position the autocomplete panel.
+   * Defaults to the autocomplete trigger element.
+   */
+  // @Input('dtMultiSelectConnectedTo') connectedTo: DtAutocompleteOrigin;
```
tomheller, line 155: Unused
```suggestion
```

libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select-trigger.ts
```diff
@@ -0,0 +1,758 @@
...
+      '(multiSelectDisabled || !panelOpen) ? null : multiSelect?.id',
+    '(focusin)': '_handleFocus()',
+    '(blur)': '_handleBlur()',
+    '(input)': '_handleInput($event)',
+    '(keydown)': '_handleKeydown($event)',
+  },
+  providers: [DT_AUTOCOMPLETE_VALUE_ACCESSOR],
+})
+export class DtFilterFieldMultiSelectTrigger<T>
+  implements ControlValueAccessor, OnDestroy {
```
tomheller, line 110: Why does the Multiselect trigger need to implement the ControlValueAccessor? This trigger should function quite the same as the filterfield-range-trigger, but seems far more complicated. 
I don't quite understand why the trigger needs it's own autocomplete that it can disable, and why it needs a controlValueAccessor? 

Maybe you can explain this to me, because I think this can be done similarly to the filter-field-range-trigger.
Please ignore further comments in this file for now.

libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select.ts
```diff
@@ -0,0 +1,344 @@
...
+
+/** Injection token to be used to override the default options for `dt-multi-select`. */
+export const DT_MULTI_SELECT_DEFAULT_OPTIONS = new InjectionToken<
+  DtMultiSelectDefaultOptions
+>('dt-multi-select-default-options', {
+  providedIn: 'root',
+  factory: DT_MULTI_SELECT_DEFAULT_OPTIONS_FACTORY,
+});
+
+/** @docs-private */
```
tomheller, line 72: Has no effect in our repo. please use @internal and prefix it with an `_`
```suggestion
@internal
```

tomheller (changes requested):
A couple of things I noticed: 

- Multi select modal cannot be navigated to or interacted with the keyboard. If a multiselect option is openend, I do think the modal should get the focus and it should be able to check the checkboxes with the keyboard. I know there is a test for it and I didn't check if this one is already passing, but when I tried it, this did not work at all.

- In the multiselect, I'm not sure what the specification says, but I don't think that you should be able to add the same filter twice. Maybe it is possible to remove the "ketchup" option for the second filter (similar to what the autocomplete does). Or is there a specific use case that needs to allow this?
image

- Option filtering by typing: Imho we should consider including the option group as a filterable element as well. When typing home, do I not want to see all condiments within the `Homemade` option group?

- A small bug I noticed when trying to edit in the e2e test app: 
  - Add a Seasoning: Ketchup filter
  - Click on seasoning again, but without completing the filter, try to edit the first filter by clicking on it.
  - The filter field ends up in weird state, where there are 2 flaps open (one is a bit hidden, but you can see it when blurring the filter field.
  - Afterwards, every consecutive filter you want to add in `Seasoning`, will already have ketchup prefilled. It seems there is a check missing if entering the edit mode is possible. 

- When closing the multiselect with `ESC`, there is no way to open it back up again. In the autocomplete, the flap opens again when typing. This does not happen with the multi-select.

-

libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select-trigger.ts
tomheller: After thinking about this a bit more, I do think I understand why you need access to the autocomplete. 

In order to filter down the choices that you present in the select, you need access to the input control. I do see that as problematic though, as we now control the autocomplete (input) from two directions. Once from the filterfield itself and once by the multiselect. 
I would suggest finding a way where the filterfield passes the current input value into multiselect, rather than accessing the whole autocomplete. 
Would that make sense?

libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select.ts
```diff
@@ -0,0 +1,344 @@
...
+    this._checkApplyDisable();
+  }
+
+  /** @internal Toggle option from template */
+  _toggleOptionFromTemplate(event: DtCheckboxChange<T>): void {
+    if (event.checked) this._currentSelection.push(event.source.value);
+    else
+      this._currentSelection = this._currentSelection.filter(
+        (opt) => opt !== event.source.value,
+      );
```
ffriedl89, line 300:
```suggestion
    if (event.checked) {
      this._currentSelection.push(event.source.value);
    } else {
      this._currentSelection = this._currentSelection.filter(
        (opt) => opt !== event.source.value,
      );
    }
```
Please use curly brackets for if statements - this is a potential pitfall

libs/barista-components/filter-field/src/filter-field.spec.ts
```diff
@@ -43,25 +43,100 @@ import {
   FILTER_FIELD_TEST_DATA_SINGLE_DISTINCT,
   FILTER_FIELD_TEST_DATA_SINGLE_OPTION,
 } from '@dynatrace/testing/fixtures';
-import { TEST_DATA_EDITMODE_ASYNC } from './filter-field-edit-mode.spec';
 import {
-  TEST_DATA_SUGGESTIONS,
-  TEST_DATA_EDITMODE,
```
ffriedl89, line 49: Why did you move those fixtures into this file?

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -656,6 +677,8 @@ export class DtFilterField<T = any>
 
   /** @internal */
   _handleInputKeyDown(event: KeyboardEvent): void {
+    console.log('filter-field', this._loading);
+
```
ffriedl89, line 681:
```suggestion

```

ffriedl89 (changes requested):
Thanks @subarroca for putting the work in. I've added a couple of comments regarding the code. The point for improvement is the trigger for me - this is where I see the some potential problems as well, as pointed out by @tomheller. Overall you added a lot tests and examples as well as documentation. Good job :) I think we will find a good solution for the trigger that helps us maintain the different types of overlays a bit better in the future.

libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select.ts
```diff
@@ -0,0 +1,344 @@
...
+   * @internal
+   * Manages active item in option list based on key events.
+   */
+  _keyManager: ActiveDescendantKeyManager<DtOption<T>>;
+
+  /** @internal */
+  _portal: TemplatePortal;
+
+  /** @internal Holds the current values of the input field for the from value */
+  _currentSelection: T[] = [];
```
thomaspink, line 176: Angular CDK can help you with selections. It provides a `SelectionModel` class that you can use. Especially powerful for multi selects as you do.
https://github.com/angular/components/blob/master/src/cdk/collections/selection-model.ts

libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select-trigger.ts
```diff
@@ -0,0 +1,758 @@
...
+
+    this._optionHeight = heightConfig.height;
+    this._maxPanelHeight = heightConfig.maxPanelHeight;
+  }
+
+  ngOnDestroy(): void {
+    this._destroy$.next();
+    this._destroy$.complete();
+    this._closingActionsSubscription.unsubscribe();
+    this._windowBlurSubscription.unsubscribe();
```
thomaspink, line 311: You already used `.takeUntil(this._destroy$)` for the `_windowBlurSubscription`. So unsubscribing is not necessary.

libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select-trigger.ts
thomaspink: As fabian and tom suggested:
- Implementing ControlValueAccessor is not necessary.
- You share a lot of code with the range. Let's create a shared class that both extend from?

thomaspink (changes requested):

subarroca:
> A couple of things I noticed:
> 
> * Multi select modal cannot be navigated to or interacted with the keyboard. If a multiselect option is openend, I do think the modal should get the focus and it should be able to check the checkboxes with the keyboard. I know there is a test for it and I didn't check if this one is already passing, but when I tried it, this did not work at all.
> * In the multiselect, I'm not sure what the specification says, but I don't think that you should be able to add the same filter twice. Maybe it is possible to remove the "ketchup" option for the second filter (similar to what the autocomplete does). Or is there a specific use case that needs to allow this?
>   image
> * Option filtering by typing: Imho we should consider including the option group as a filterable element as well. When typing home, do I not want to see all condiments within the `Homemade` option group?
> * A small bug I noticed when trying to edit in the e2e test app:
>   
>   * Add a Seasoning: Ketchup filter
>   * Click on seasoning again, but without completing the filter, try to edit the first filter by clicking on it.
>   * The filter field ends up in weird state, where there are 2 flaps open (one is a bit hidden, but you can see it when blurring the filter field.
>   * Afterwards, every consecutive filter you want to add in `Seasoning`, will already have ketchup prefilled. It seems there is a check missing if entering the edit mode is possible.
> * When closing the multiselect with `ESC`, there is no way to open it back up again. In the autocomplete, the flap opens again when typing. This does not happen with the multi-select.

I'll address these issues. Thanks for the extensive testing 👍

libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select-trigger.ts
subarroca: My mistake. I removed it

subarroca:
TBH The development has been done in 2 stages, one for the functionality and 4 months later for the keyboard interaction. After the first iteration I did an extensive clean-up but when I faced the second one I could not make the keyboard interaction work so I put back lots of code just to achieve it and I did not do the clean-up again.

Thanks for your thorough review and comments. I'll try to address everything in the following week(s)

libs/barista-components/filter-field/src/filter-field.spec.ts
subarroca: My fault. It was done some time ago and then I found that you separated the files and tried to keep up to date. I'll put it back

subarroca:
**This is @tomheller  task list**

- [x] Multiselect should be unique by default
- [x] Include option group as a filterable element. (No highlight possible at the moment)
- [ ] Multi select modal cannot be navigated to or interacted with the keyboard. If a multiselect option is openend, I do think the modal should get the focus and it should be able to check the checkboxes with the keyboard. I know there is a test for it and I didn't check if this one is already passing, but when I tried it, this did not work at all.

- [ ] A small bug I noticed when trying to edit in the e2e test app:

- Add a Seasoning: Ketchup filter
- Click on seasoning again, but without completing the filter, try to edit the first filter by clicking on it.
- The filter field ends up in weird state, where there are 2 flaps open (one is a bit hidden, but you can see it when blurring the filter field.
- Afterwards, every consecutive filter you want to add in Seasoning, will already have ketchup prefilled. It seems there is a check missing if entering the edit mode is possible.

- [ ] When closing the multiselect with ESC, there is no way to open it back up again. In the autocomplete, the flap opens again when typing. This does not happen with the multi-select.

**This is @ffriedl89 and @thomaspink list**
- [ ] Remove ControlValueAccessor implementation.
- [ ] Create a  generic class for range and filter and extend them

tomheller:
@subarroca thank you for updating the PR. As of right now, we are quite busy so please be patient as we want to do a thorough review of this PR and this will take some time.

tomheller (commented):
Hi @subarroca!

This looks quite good already. Here are a couple of scenarios where I think the behavior could be improved in a second iteration (I don't think this is quite necessary in v1). 

1. When `Month` is selected in the dev app and the focus stays in the input field. `Spacebar` has 2 meanings. First it adds a space to the filtered value, second it would select the checkbox. We might need to do some more focus management here.

2. When selecting `Month` with `ENTER`, then typing `Janu`, then hit `SPACE` and `ENTER`. The selected value is `All`, where only january was filtered. We might need to handle the first active value here a bit differently.

Good work :tada:

tomheller (dismissed):

tomheller:
I have rebased master into the branch and fixed the examples bazel build. 
sadly buildkite is a bit flaky at the moment. The issue has been reported to the responsible team and should be handled shortly.

libs/barista-components/filter-field/README.md
```diff
@@ -103,6 +104,10 @@ Every filter is an array of:
   user. This includes the range operator (`DtFilterFieldRangeOperator`), the
   unit (`string`) of the selected values and the range itself (which can either
   be one `number` value or a `number` tuple).
+- Multi select filters: array of objects that have been selected via an multi
```
tomheller, line 107:
```suggestion
- Multi select filters: array of objects that have been selected via a multi
```

libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select-trigger.ts
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  constructor(
+    protected _elementRef: ElementRef,
+    protected _overlay: Overlay,
+    protected _changeDetectorRef: ChangeDetectorRef,
+    protected _viewportRuler: ViewportRuler,
+    protected _platform: Platform,
+    protected _overlayContainer: OverlayContainer,
+    _zone: NgZone,
```
tomheller, line 83: Nitpick: 
Not quite necessary to prefix this with a `_` as this is not a class member

tomheller (approved):

libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select-trigger.ts
```diff
@@ -80,7 +85,7 @@ export class DtFilterFieldMultiSelectTrigger<T>
     protected _viewportRuler: ViewportRuler,
     protected _platform: Platform,
     protected _overlayContainer: OverlayContainer,
-    _zone: NgZone,
+    protected _zone: NgZone,
```
tomheller, line 88: Sorry, not quite what I meant in the previous review. 
The zone does not need to be a member, because you are simply passing it to the super constructor. 

it's fine either way though.

subarroca: Yes, I know you meant to remove the underscore but as all the others had this format I preferred adding the `protected`

tomheller:
This looks good now,  buildkite seems to have passed as well. 
@thomaspink & @ffriedl89 you have still requested changes. Could we get this over the finish line?

libs/barista-components/filter-field/src/filter-field.html
```diff
@@ -95,6 +99,19 @@
...
+    <dt-filter-field-multi-select
+      #multiSelect="dtFilterFieldMultiSelect"
+      (multiSelectSubmitted)="_handleMultiSelectSubmitted($event)"
+      [optionsOrGroups]="
+        _currentDef &&
+        !!_currentDef!.multiSelect &&
+        _currentDef!.multiSelect!.multiOptions
+      "
+      [inputValue]="_inputValue"
+      panelWidth="400px"
```
ffriedl89, line 112: this is gonna be a problem soon, but i am fine with this being in there for v1

ffriedl89 (approved):
LGTM! Thanks @subarroca for the big big contribution. Ready to merge from my point of view.

libs/barista-components/filter-field/src/filter-field.html
tomheller: Autocomplete panel is also somewhere clipped. Let's see when it will be a problem.

tomheller (approved):

PR #1840, opened 2020-11-26
feat: Custom properties for row(s).

bardosmisi:

Fixes the row, expandable-row and tree-table-row parts of APM-266103.

#### Type of PR


libs/barista-components/table/src/row.scss
```diff
@@ -24,10 +28,17 @@
 }
 
 :host-context(.dt-table-interactive-rows):hover {
-  border: solid 1px $gray-500;
+  border: solid 1px var(--interactive-border-color);
```
tomheller, line 31: Hmm, this was gray-500 before and is now gray-100?
```suggestion
  border: solid 1px var(--interactive-border-color);
```

bardosmisi: I am not sure how this happened, good catch.

tomheller (approved):

tomheller:
When squashing this, please reword to a feature.

lukasholzer (approved):

lukasholzer (dismissed):

lukasholzer:
@bardosmisi please reword to `refactor(tree-table): Change styling to use css-custom properties.` We don't have a `feature` commit scope. You can use `feat` instead for features [comitlint.conif.js](https://github.com/dynatrace-oss/barista/blob/master/commitlint.config.js#L5). Should be caught by the commit lint have you commited with `--no-verify`?

But In my opinion, it is a refactor as it does not add additional functionality it refactors existing behaviour to a newer technology.

tomheller:
> But In my opinion, it is a refactor as it does not add additional functionality it refactors existing behaviour to a newer technology.

I agree, but we added all of the refactor of custom properties as a feature to show up in the changelog. For consistency, I would stick to that.

tomheller:
Reworded the commit to match the commit guidelines and the other custom property updates.

rowa-audil (approved):
LGTM

tomheller (approved):

PR #1913, opened 2021-01-12
fix(sunburst-chart): Fix tooltip flicker on click

subarroca:
#### Bugfix (non-breaking change which fixes an issue)

Currently tooltip gets repositioned on segment click and before any other mouse move.
Now the overlay has been moved to the segment without manual handling.
Also, to preserve the possibility of opening the overlay from the outside I've kept the functionality in the main component

#### Checklist

### DIFF
--- a/libs/barista-components/sunburst-chart/src/sunburst-chart.spec.ts
+++ b/libs/barista-components/sunburst-chart/src/sunburst-chart.spec.ts
@@ -389,36 +389,25 @@ describe('DtSunburstChart', () => {
       expect(overlayContainer).toBeDefined();
     });
 
-    it('should display an overlay when hovering over a slice', () => {
-      const firstSlice = fixture.debugElement.query(By.css(selectors.segment));
-
-      dispatchFakeEvent(firstSlice.nativeElement, 'mouseenter');
+    it('should display and hide an overlay when calledff', () => {
+      component.openOverlay(component.slices[0]);
       fixture.detectChanges();
 
-      const overlayPane = overlayContainerElement.querySelector(
+      let overlayPane = overlayContainerElement.querySelector(
         selectors.overlay,
       );
       expect(overlayPane).toBeDefined();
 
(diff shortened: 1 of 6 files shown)


tomheller (commented):
Hey @subarroca ! 
Thank you for this fix. We are in the middle of a breaking version release, so this might not be merged before we can safely release v9 (hopefully in this or next week). 

I noticed that due to the changes you made, there are now some unit tests failing. Could you please adjust them accordingly?

subarroca:
Yes, I'm dealing with the unit, it slipped my mind, thanks for letting me know too.
No problem if this goes a bit later in time :)

libs/barista-components/sunburst-chart/src/sunburst-chart.spec.ts
```diff
@@ -389,36 +389,25 @@ describe('DtSunburstChart', () => {
...
+      let overlayPane = overlayContainerElement.querySelector(
         selectors.overlay,
       );
       expect(overlayPane).toBeDefined();
 
       const overlayContent = (overlayPane!.textContent ?? '').trim();
       expect(overlayContent).toBe('Purple');
-    });
-
-    it('should remove the overlay when moving the mouse away from the slice', () => {
```
subarroca, line 407: This is checked in e2e, so removing it from here

libs/barista-components/sunburst-chart/src/sunburst-chart.spec.ts
```diff
@@ -389,36 +389,25 @@ describe('DtSunburstChart', () => {
       expect(overlayContainer).toBeDefined();
     });
 
-    it('should display an overlay when hovering over a slice', () => {
```
subarroca, line 392: This is checked in e2e, so removing it from here

tomheller (approved):

rowa-audil (approved):
LGTM

tomheller (approved):

PR #1919, opened 2021-01-14
feat(filter-field): Adds a quick input option (default search option)

rowa-audil:

Adds the quick input feature to the filter field. By using a freeText set to defaultSearch you can try it out.
README explains functionality

Thank you a lot @thomaspink for the guidance! Was a great learning experience.

#### Type of PR

- Feature (non-breaking change which adds functionality)

#### Checklist

### DIFF
--- a/apps/components-e2e/src/components/quick-filter/quick-filter.e2e.ts
+++ b/apps/components-e2e/src/components/quick-filter/quick-filter.e2e.ts
@@ -329,7 +329,6 @@ test('should work with async data and handle distincts correctly', async (testCo
   await clickOption(2);
   // Click option Los Angeles
   await clickOption(2);
-
   // Expect the filter to be set
   await testController
     .expect(getFilterfieldTags())
--- a/libs/barista-components/filter-field/src/filter-field.ts
+++ b/libs/barista-components/filter-field/src/filter-field.ts
@@ -75,7 +75,6 @@ import {
   Subscription,
 } from 'rxjs';
 import {
-  debounceTime,
   delay,
   distinctUntilChanged,
   filter,
@@ -116,6 +115,7 @@ import {
   filterAutocompleteDef,
   filterFreeTextDef,
   filterMultiSelectDef,
+  findDefaultSearch,
   findFilterValuesForSources,
   isDtAutocompleteValueEqual,
   isDtFreeTextValueEqual,
@@ -124,13 +124,16 @@ import {
 } from './filter-field-util';
 import { DtFilterFieldControl } from './filter-field-validation';
 import {
+  DefaultSearchOption,
   DtAutocompleteValue,
   DtFilterFieldTagData,
   DtFilterValue,
+  DtFreeTextDef,
   DtNodeDef,
   DtOptionDef,
   isAsyncDtAutocompleteDef,
   isAsyncDtMultiSelectDef,
+  isDefaultSearchOption,
   isDtAutocompleteDef,
   isDtAutocompleteValue,
   isDtFreeTextDef,
@@ -167,8 +170,6 @@ export class DtFilterFieldCurrentFilterChangeEvent<T> {
     public filters: T[][],
   ) {}
 }
-
-export const DT_FILTER_FIELD_TYPING_DEBOUNCE = 200;
 
 // We need to save the instance of the filterField that currently
 // has a flap open, to make sure we can close the old one.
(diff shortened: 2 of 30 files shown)


tomheller (dismissed):
LGTM 👍

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -1389,7 +1422,20 @@ export class DtFilterField<T = any>
 
   /** Updates the list of options or groups displayed in the autocomplete overlay */
   private _updateAutocompleteOptionsOrGroups(): void {
-    const currentDef = this._currentDef;
+    // In order to only display non defaultSearchDef's by filtering, we need to clone deep here.
+    const currentDef = cloneDeep(this._currentDef);
```
thomaspink, line 1426: I am not sure if the filtering here is correct. I think we can revert this whole function back. Let's hopp on a call

thomaspink (changes requested):

lukasholzer (approved):

ffriedl89:
Please squash your commits and we are ready to go :)

tomheller (approved):
Looks good to me.

thomaspink (dismissed):

rowa-audil:
@tomheller Rebased and Squashed

tomheller (approved):

tomheller:
@rowa-audil Looks good now. as we are still in major release stage, please do not add `merge-ready` label, as this would confuse the cherrypicker.

ffriedl89:
I've had a session with Holger and there are still some issues that might be a bit tricky to get fixed. Try testing by typing and hitting enter fast. The suggestion only gets applied when its already visible which has a little bit of a delay. You can end up with empty filters or just select Asia or some characters might be cut off. 

For further infos - i also have some screen recordings if needed. Just don't hesitate to reach out.

apps/components-e2e/src/components/quick-filter/quick-filter.e2e.ts
```diff
@@ -343,10 +343,10 @@ test('should work with async data and handle distincts correctly', async (testCo
     // Expect Vienna and Graz to still be in the list.
     .expect(options.nth(0).textContent)
     // textContent is duplicated because of the highlight within the option
-    .eql('ViennaVienna')
+    .eql(' ViennaVienna')
```
tomheller, line 346: Why is there a space in there now? I'm confused.

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -169,8 +168,6 @@ export class DtFilterFieldCurrentFilterChangeEvent<T> {
   ) {}
 }
 
-export const DT_FILTER_FIELD_TYPING_DEBOUNCE = 200;
```
tomheller, line 168: Alright, this seems a bit drastic but I get where the change is coming from. 
If it still works and does not produce side effects. After all we are emitting way `inputChange` outputs now than before. There probably are cases where consumers bind to this event and go off an fetch async data.

This would be a good hint to include into the changelog, so people are aware.

tomheller (approved):

apps/components-e2e/src/components/quick-filter/quick-filter.e2e.ts
rowa-audil: I get this error otherwise:

Screenshot 2021-02-01 at 10 36 30

tomheller: I'm almost 100% certain this can be tracked back to this change:
https://github.com/dynatrace-oss/barista/pull/1919/files#diff-e38531c013cd8f799b6f772d2d6f8dcddb0a621551c466d38b3eebbee33b76a5L86-R111

tomheller (approved):

apps/components-e2e/src/components/quick-filter/quick-filter.e2e.ts
rowa-audil: Solved. TY!

tomheller (dismissed):

ffriedl89 (approved):
Really good PR - feels super snappy.

Please add the following to the commit body.
`NOTE: The debounce for the inputChange event was removed. This leads to the inputChange event being fired more often. Please cancel requests that are inflight or debounce the event yourself.`

rowa-audil:
> Really good PR - feels super snappy.
> 
> Please add the following to the commit body.
> `NOTE: The debounce for the inputChange event was removed. This leads to the inputChange event being fired more often. Please cancel requests that are inflight or debounce the event yourself.`

@ffriedl89 
Reworded the commit.

tomheller (approved):

ffriedl89 (approved):

thomaspink (approved):

PR #1955, opened 2021-02-05
feat(filter-field): Keyboard support

rowa-audil:

Adds keyboard support to the filter-field

#### Type of PR

#### Checklist

### DIFF
--- a/libs/barista-components/filter-field/src/filter-field.ts
+++ b/libs/barista-components/filter-field/src/filter-field.ts
@@ -28,6 +28,7 @@ import {
   DELETE,
   ENTER,
   ESCAPE,
+  LEFT_ARROW,
   UP_ARROW,
 } from '@angular/cdk/keycodes';
 import { DOCUMENT } from '@angular/common';
@@ -516,6 +517,18 @@ export class DtFilterField<T = any>
   /** Whether the filter field or one of it's child elements is focused. */
   private _isFocused = false;
 
+  /** Whether to allow focusing the tags when navigating in the input element using arrow keys */
+  private _allowArrowKeyFocusOnTags = false;
+
+  private _currentFocusState: 'none' | 'input' | 'inputboundary' = 'none';
+
+  private _currentlyFocusedTag:
+    | 'none'
+    | 'editButton'
+    | 'deleteButton'
+    | 'last'
+    | 'first' = 'none';
+
   /** Merges the autocomplete and range opened events */
   private _mergedAutocompleteRangeOpened: Observable<unknown>;
 
@@ -770,7 +783,7 @@ export class DtFilterField<T = any>
     this._changeDetectorRef.markForCheck();
   }
 
-  /** @internal */
+  /** @internal Handles keydown on the input */
   _handleInputKeyDown(event: KeyboardEvent): void {
     if (this._loading) {
       return;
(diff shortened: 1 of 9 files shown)


rowa-audil:
Adding tests

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -494,6 +496,15 @@ export class DtFilterField<T = any>
   /** Whether the filter field or one of it's child elements is focused. */
   private _isFocused = false;
 
+  /** Whether to allow focusing the tags when navigating in the input element using arrow keys */
+  private allowArrowKeyFocusOnTags = false;
```
ffriedl89, line 500: Please prefix with an _ for internal

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -494,6 +496,15 @@ export class DtFilterField<T = any>
...
+
+  /** Whether the focus in the suffix tags or not */
+  private inSuffix = false;
```
ffriedl89, line 503: I'd reword the variables to e.g. _hasFocusInSuffix - also think about using one focus state variable instead as well. The focus can never be in both (suffix and prefix) so they are mutually exclusive. If you combine them into one variable, it gets easier to avoid impossible state. 
pseudo code:
```
private _currentFocusState: 'none' | 'input' | 'inputstart' | 'inputend' | 'prefix' | 'suffix' = 'none';
```

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -748,12 +759,21 @@ export class DtFilterField<T = any>
     this._changeDetectorRef.markForCheck();
   }
 
-  /** @internal */
-  _handleInputKeyDown(event: KeyboardEvent): void {
+  /** @internal Handles keydown on the input */
+  _handleInputKeyDown(event: KeyboardEvent): boolean | void {
```
ffriedl89, line 763: no need to add the boolean return value here. You only return void

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -748,12 +759,21 @@ export class DtFilterField<T = any>
...
     if (this._loading) {
       return;
     }
 
+    if (
```
ffriedl89, line 768: NIT: 
You can always use a ternary operator for assignments. 
```
this.allowArrowFocusOnTags = this._inputEl.nativeElement.selectionStart === 0 ||
      this._inputEl.nativeElement.selectionStart === this._inputValue.length;

File: libs/barista-components/filter-field/src/filter-field.ts [hunk 5]
```diff
@@ -805,6 +825,222 @@ export class DtFilterField<T = any>
...
+      if (this.allowArrowKeyFocusOnTags || !this._inputValue.length) {
+        this._handleArrowKeyNavigation({ right: true });
+      } else {
+        this.allowArrowKeyFocusOnTags = true;
+      }
+    }
+  }
+
+  /** @internal Handles the navigation through the list of tags when using arrow keys */
+  _handleArrowKeyNavigation(navigationEvent: {
```
tomheller, line 860: This function is majestically complicated. If nesting is about 5 levels deep. Could we break this up a little bit? 
I know the focus handling is quite complex, if it is not possible/worth it reducing this function, it's fine as well.

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -805,6 +825,222 @@ export class DtFilterField<T = any>
...
+    /** Contains the currently focused tag from which the event was fired from */
+    tag?: DtFilterFieldTag;
+    /** Whether to step back */
+    left?: boolean;
```
ffriedl89, line 864:
```suggestion
    direction: 'left' | 'right';
```
Again here - right and left can never be true at the same time, use one variable instead - makes it easier

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -805,6 +825,222 @@ export class DtFilterField<T = any>
...
+    /** Whether to step forward */
+    right?: boolean;
+  }): void {
+    const focusedTagIndex = this._tags
+      .toArray()
+      .findIndex((tag) => tag.editButton === navigationEvent.tag?.editButton);
+    const inputElementHasFocus =
+      document.activeElement === this._inputEl.nativeElement;
+    const inSplittedMode = !!this._suffixTagData.length;
```
tomheller, line 873: Nitpick: `Splitted` does not exist, please use `split`. 

Also it is quite confusing to double negate, if in fact you are simply checking if the lenght is not 0
```suggestion
    const inSplitMode = this._suffixTagData.length !== 0;
```

ffriedl89, line 873: NIT:
```suggestion
    const inSplitMode = !!this._suffixTagData.length;
```

libs/barista-components/filter-field/src/filter-field.ts
tomheller, line 503:
```suggestion
  private isFocusInSuffix = false;
```

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -494,6 +496,15 @@ export class DtFilterField<T = any>
...
   private _isFocused = false;
 
+  /** Whether to allow focusing the tags when navigating in the input element using arrow keys */
+  private allowArrowKeyFocusOnTags = false;
+
+  /** Whether the focus in the suffix tags or not */
+  private inSuffix = false;
+
+  /** Whether the focus in the suffix tags or not */
+  private inPrefix = false;
```
tomheller, line 506:
```suggestion
  private isFocusInPrefix = false;
```

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -805,6 +825,222 @@ export class DtFilterField<T = any>
...
+      this._tags.first.editButton.nativeElement === document.activeElement;
+    const lastElementHasFocus =
+      this._getNextFocusableTagButton(this._tags.last, true).nativeElement ===
+      document.activeElement;
+    const editButtonHasFocus =
+      navigationEvent.tag?.editButton.nativeElement === document.activeElement;
+    const deleteButtonHasFocus =
+      navigationEvent.tag !== undefined &&
+      this._getNextFocusableTagButton(navigationEvent.tag!, true)
+        .nativeElement === document.activeElement;
```
ffriedl89, line 912: This could be moved into a function that returns the current focused element - only one can be focused ;) 
So e.g. return 'edit-button' or 'delete-button'

libs/barista-components/filter-field/src/filter-field.ts
ffriedl89, line 860: Try to split up this method into smaller methods that return values. I got lost along the way - There are too many if else in there for me :D

ffriedl89 (changes requested):

libs/barista-components/filter-field/src/filter-field.ts
tomheller, line 768: Add a comment on what this actually checks. If your unfamiliar with selectionStart and what it does, this could be bit confusing.

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -748,12 +759,21 @@ export class DtFilterField<T = any>
...
     }
 
+    if (
+      this._inputEl.nativeElement.selectionStart !== 0 &&
+      this._inputEl.nativeElement.selectionStart !== this._inputValue.length
+    ) {
+      this.allowArrowKeyFocusOnTags = false;
+    } else {
+      this.allowArrowKeyFocusOnTags = true;
+    }
```
tomheller, line 775:
```suggestion
    this.allowArrowKeyFocusOnTags = !(this._inputEl.nativeElement.selectionStart !== 0 &&
      this._inputEl.nativeElement.selectionStart !== this._inputValue.length);
```

tomheller (changes requested):
The code looks good in general, the `_handleArrowKeyNavigation` function is fairly complex...

One thing that I noticed during testing:
When starting in a completely empty filter-field, pressing either the left or right arrow key result in an error: `Can't access property "deleteButton"|"editButton", tag is undefined`

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -516,6 +518,18 @@ export class DtFilterField<T = any>
   /** Whether the filter field or one of it's child elements is focused. */
   private _isFocused = false;
 
+  /** Whether to allow focusing the tags when navigating in the input element using arrow keys */
+  private _allowArrowKeyFocusOnTags = false;
+
+  private _currentFocusState: 'none' | 'input' | 'inputboundry' = 'none';
```
tomheller, line 524: Please add jsDoc here as as well as in line 526.
```suggestion
  private _currentFocusState: 'none' | 'input' | 'inputboundary' = 'none';
```

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -827,6 +841,223 @@ export class DtFilterField<T = any>
...
+  /** @internal Handles keyup on the input */
+  _handleInputKeyUp(event: KeyboardEvent): void {
+    if (!this._tags.length) {
+      return;
+    }
+
+    this._getCurrentFocusState();
+
+    const keyCode = _readKeyCode(event);
+    if (keyCode === LEFT_ARROW && this._currentFocusState === 'inputboundry') {
```
tomheller, line 853: You only react to things, if `currentFocusState` is `inputboundary`, maybe we can add a check first, and early exit.

```
if (this._currentFocusState !== 'inputboundary') {
  return;
}
```

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -827,6 +841,223 @@ export class DtFilterField<T = any>
...
+    const editButtonHasFocus =
+      currentTag.editButton.nativeElement === document.activeElement;
+    const deleteButtonHasFocus =
+      this._getNextFocusableTagButton(currentTag, true).nativeElement ===
+      document.activeElement;
+
+    if (firstTagHasFocus) this._currentlyFocusedTag = 'first';
+    else if (lastTagHasFocus) this._currentlyFocusedTag = 'last';
+    else if (editButtonHasFocus) this._currentlyFocusedTag = 'editButton';
+    else if (deleteButtonHasFocus) this._currentlyFocusedTag = 'deleteButton';
```
tomheller, line 1002: No... just no 😉 There are no if statements without braces :)
```suggestion
    if (firstTagHasFocus) {
     this._currentlyFocusedTag = 'first';
    } else if (lastTagHasFocus) {
      this._currentlyFocusedTag = 'last';
    } else if (editButtonHasFocus) { 
      this._currentlyFocusedTag = 'editButton'; 
    } else if (deleteButtonHasFocus) {
      this._currentlyFocusedTag = 'deleteButton';
    }
```

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -827,6 +841,223 @@ export class DtFilterField<T = any>
...
+  _getNextFocusableTagButton(
+    tag: DtFilterFieldTag,
+    deleteButtonFirst: boolean,
+  ): ElementRef<HTMLButtonElement> {
+    if (deleteButtonFirst) {
+      if (tag.deleteButton) {
+        return tag.deleteButton;
+      } else {
+        return tag.editButton;
+      }
```
tomheller, line 1018: Unnecessary else. As you already return in the if condition, you can do sth like this: 
```suggestion
        return tag.editButton;
```

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -827,6 +841,223 @@ export class DtFilterField<T = any>
...
+    } else {
+      if (tag.editButton) {
+        return tag.editButton;
+      } else {
+        return tag.deleteButton;
+      }
```
tomheller, line 1024: Same here, unnecessary else.

tomheller (changes requested):
Way better and cleaner code. This is way more understandable. only a couple of notes.

libs/barista-components/filter-field/src/filter-field.ts
tomheller: This seems to be still in the code. I'm fine with it, consider it a nitpick.

tomheller (dismissed):
Generally looks good now. Thank you for taking care of this. 

A couple of my comments have not been applied, but these can be considered nitpicks as well. 

Seems that tests are still failing.

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -827,6 +840,216 @@ export class DtFilterField<T = any>
...
+
+    const focusedTagIndex = this._tags
+      .toArray()
+      .findIndex((tag) => tag.editButton === currentTag?.editButton);
+
+    this._getCurrentlyFocusedTagType(currentTag!);
+
+    if (direction === 'left') {
+      this._findPreviousTagDefaultMode(
+        currentTag!,
```
tomheller, line 889: Should not be necessary, as you check for undefined above already
```suggestion
        currentTag,
```

ffriedl89 (dismissed):

libs/barista-components/filter-field/src/filter-field.ts
```diff
@@ -827,6 +840,216 @@ export class DtFilterField<T = any>
...
+  private _getCurrentFocusState(): void {
+    const isStart = this._inputEl.nativeElement.selectionStart === 0;
+    const isEnd =
+      this._inputEl.nativeElement.selectionStart === this._inputValue.length;
+    const isInput = document.activeElement === this._inputEl.nativeElement;
+
+    if (isStart) {
+      this._currentFocusState = 'inputboundary';
+    } else if (isEnd) {
+      this._currentFocusState = 'inputboundary';
```
tomheller, line 1035:
```suggestion
    if (isStart || isEnd) {
      this._currentFocusState = 'inputboundary';
```

tomheller (approved):
Good work, 2 small nitpicks 🎉

tomheller (approved):

thomaspink (approved):

PR #1957, opened 2021-02-08
fix(table): Fixes an issue with partial keyframes on expandable row.

tomheller:

This was quite impossible to reproduce reliably and to be honest, we are not entirely sure if this will fix the problem.
Regarding to this issue in the angular components, this should help though and we were not seeing the error after this fix in our development environment.
https://github.com/angular/components/issues/13970

Fixes APM-276923

#### Type of PR

Bugfix

#### Checklist

### DIFF
--- a/libs/barista-components/table/src/expandable/expandable-row.ts
+++ b/libs/barista-components/table/src/expandable/expandable-row.ts
@@ -61,7 +61,7 @@ const COLLAPSE_ANIMATE = [
       style({ height: 'auto', visibility: 'hidden', offset: 0 }),
       style({
         height: '0px',
-        minHeight: '0',
+        minHeight: '0px',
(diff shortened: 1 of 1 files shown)


libs/barista-components/table/src/expandable/expandable-row.ts
```diff
@@ -61,7 +61,7 @@ const COLLAPSE_ANIMATE = [
       style({ height: 'auto', visibility: 'hidden', offset: 0 }),
       style({
         height: '0px',
-        minHeight: '0',
+        minHeight: '0px',
```
rowa-audil, line 64: 😄

rowa-audil (approved):

ffriedl89 (approved):

Issue #1959, opened 2021-02-08
Radial chart legend percentages


tonidalmases:
# Bug Report

## Expected Behavior

When using the radial chart with percentages, without _Max value_ (full chart), the sum of percentages in the legend should be equal to 100%.

## Current Behavior

In some occasions, the sum is not equal to 100, since the chart does not consider a precision, nor rounded percentages.

The current behavior (the sum is equal to 99.98)
![imagen](https://user-images.githubusercontent.com/27827642/107201905-dbc7a700-69f9-11eb-84d4-d922f97be927.png)

## Possible Solution

Adding `precision` to the chart would fix the number of decimals to display in the legend.
![imagen](https://user-images.githubusercontent.com/27827642/107202685-c010d080-69fa-11eb-9f56-404ef388217a.png)

However, note that the sum would be equal to 99.9, since those percentages are not rounded to 100. A possible implementation, described [Here](https://revs.runtime-revolution.com/getting-100-with-rounded-percentages-273ffa70252b), would also solve the issue by distributing the percentages in a way that the final result is equal to 100.
![imagen](https://user-images.githubusercontent.com/27827642/107202932-0b2ae380-69fb-11eb-84a0-078b730d15f1.png)

## Steps to Reproduce

https://stackblitz.com/edit/github-xzqveu?file=src/radial-chart/radial-chart-overlay-example/radial-chart-overlay-example.ts

Issue also described in APM-277510.

This issue is extensive to all kind of charts in barista whose percentages must sum up 100%.

### Used Versions:

- **typescript**: 3.8.3
- **angular**: 11
- **@angular/cdk**: 11.1.0
- **@dynatrace/barista-components**: 9.0.0-rc.3

tomheller:
Hi @tonidalmases! We are no longer tracking issues here in github. please read #1786 and reopen the ticket in Jira.

PR #1960, opened 2021-02-08
fix(radial-chart): Display of rounded percentage with precision

tonidalmases:

This PR fixes the issues with the **radial-chart** legend percentages described in #1959 .

#### Type of PR

#### Checklist

### DIFF
--- a/libs/barista-components/radial-chart/src/utils/radial-chart-utils.ts
+++ b/libs/barista-components/radial-chart/src/utils/radial-chart-utils.ts
@@ -85,3 +85,63 @@ export function generatePathData(
     endAngle,
   });
 }
+
+/**
+ * Returns an array of percentages with decimal precision.
+ */
+export function getPercentages(
+  values: number[],
+  total: number,
+  precision: number,
+): number[] {
+  const sumAllValues = getSum(values);
+  if (total > sumAllValues) {
+    return values.map((value) => +((value / total) * 100).toFixed(precision));
+  }
+
+  return getRoundedPercentages(values, precision);
+}
+
+/**
+ * Returns an array of rounded percentages so that the sum of
+ * all values is equal to 100.
+ */
+export function getRoundedPercentages(
+  values: number[],
+  precision: number,
+): number[] {
+  const sumAllValues = getSum(values);
+  const precisionPower = Math.pow(10, precision);
+
+  // Temp object to keep the percentage with precision decimals and initial position.
+  const tempPercentages = values.map((value, i) => ({
+    percentage: (value / sumAllValues) * 100 * precisionPower,
+    position: i,
+  }));
+
+  // Difference between the total and the sum of the rounded values.
+  const diff =
+    Math.pow(10, precision + 2) -
+    tempPercentages.reduce((sum, x) => sum + Math.floor(x.percentage), 0);
+
+  // Sorting the items in decimal decreasing order.
+  tempPercentages.sort(
+    (x, y) =>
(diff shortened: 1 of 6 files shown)


libs/barista-components/radial-chart/src/utils/radial-chart-utils.ts
```diff
@@ -85,3 +85,64 @@ export function generatePathData(
     endAngle,
   });
 }
+
+/**
+ * Returns an array of percentages with decimal precision.
+ */
+export function getPercentages(
```
rowa-audil, line 92: We could move this to the percentage formatter or to the core. 
@tomheller What do you think?

libs/barista-components/radial-chart/src/utils/radial-chart-utils.ts
```diff
@@ -85,3 +85,64 @@ export function generatePathData(
...
+  }
+
+  return getRoundedPercentages(values, precision);
+}
+
+/**
+ * Returns an array of rounded percentages so that the sum of
+ * all values is equal to 100.
+ */
+export function getRoundedPercentages(
```
rowa-audil, line 109: Dito

libs/barista-components/radial-chart/src/utils/radial-chart-utils.ts
rowa-audil: For now, it looks good! I suggest a follow-up issue if we want to move the functions

rowa-audil (commented):
Looks good to me, but I'll stand by with approving

libs/barista-components/radial-chart/src/utils/radial-chart-utils.ts
```diff
@@ -85,3 +85,64 @@ export function generatePathData(
...
+  return (
+    tempPercentages
+      // Distributing the difference to the first items.
+      .map((x, i) => ({
+        percentage:
+          i < diff ? Math.floor(x.percentage) + 1 : Math.floor(x.percentage),
+        position: x.position,
+      }))
+      // Sorting the items in the initial position.
+      .sort((x, y) => x.position - y.position)
```
ffriedl89, line 144: Sort does not return a new array but mutate the array. 
I'd move the sort out of the chain - and do the sort before returning tempPercentages.

ffriedl89 (dismissed):

libs/barista-components/radial-chart/src/utils/radial-chart-utils.ts
tonidalmases: Thank you @rowa-audil, as I comment in the [Issue](https://github.com/dynatrace-oss/barista/issues/1959), we might have this behavior for all charts whose percentages must sum up 100. I am open to move this function to an upper level, so that other components can use it as well.

libs/barista-components/radial-chart/src/utils/radial-chart-utils.ts
tonidalmases: Thanks for your comment @ffriedl89, I will take care of it.

tomheller:
Please squash your commits, before we can merge this.

subarroca (dismissed):

tomheller (approved):

tomheller (approved):

tomheller:
Cherry picked into 
9.x d08b8085643ce3461b90f3804eb369bd9459a4eb
9.1.x 66e42019f5996f60b8d5715d081c77ac0965c1e1

PR #1973, opened 2021-02-11
feat(filter-field): Adds seperate disabled member in tags

rowa-audil:

APM-283060

#### Type of PR

#### Checklist

### DIFF
--- a/libs/barista-components/filter-field/src/filter-field-tag/filter-field-tag.scss
+++ b/libs/barista-components/filter-field/src/filter-field-tag/filter-field-tag.scss
@@ -49,7 +49,7 @@
   color: $gray-500;
 
   &[disabled] {
-    color: $gray-300;
+    background-color: $white;
   }
 
   .dt-button-icon {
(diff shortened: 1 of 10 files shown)


tomheller:
Depends on #1955 consider this one blocked until it is merged.

ffriedl89:
I am really unsure about this - maybe we can get some eyes on this from UX. This seems to be really distracting - the more I try this the more I get annoyed by the tags constantly changing states. What about a little different solution - to not change any visuals but use a block cursor and maybe add a title attribute to the tags that states you are currently creating / editing another tag?

libs/barista-components/filter-field/src/filter-field-tag/filter-field-tag.scss
```diff
@@ -49,7 +49,7 @@
   color: $gray-500;
 
   &[disabled] {
-    color: $gray-300;
+    background-color: #ffffff;
```
tomheller, line 52: could be $white for consistency, but is fine like that as well.

libs/barista-components/filter-field/src/filter-field-tag/filter-field-tag.scss
```diff
@@ -29,7 +29,7 @@
 
     .dt-filter-field-tag-key,
     .dt-filter-field-tag-value {
-      color: $gray-500;
+      color: $gray-300;
```
tomheller, line 32: why are we changing the color here?

libs/barista-components/filter-field/src/filter-field-tag/filter-field-tag.ts
```diff
@@ -84,30 +84,41 @@ export class DtFilterFieldTag implements OnDestroy {
...
   set disabled(value: boolean) {
     const coercedValue = coerceBooleanProperty(value);
+    this._disabled = coercedValue;
     this.editable = !coercedValue;
-    this.deletable = !coercedValue;
     this._changeDetectorRef.markForCheck();
   }
+  private _disabled = false;
 
+  /** Whether the tag is temporarily disabled or not. Based on editmode state */
```
tomheller, line 100:
```suggestion
  /** @internal Whether the tag is temporarily disabled or not. Based on editmode state */
```

tomheller (dismissed):

tomheller (approved):

tomheller:
Cherry picker failed to CP this into the minor branch, because the original feature is not yet in the minor. This is fine.

PR #1985, opened 2021-02-16
fix(filter-field): The id a consumer provides via data of an option …

thomaspink:
feat(filter-field): The id a consumer provides via data of an option will be used to track it over time when using the DefaultDataSource.

### DIFF
--- a/libs/barista-components/quick-filter/src/state/reducer.ts
+++ b/libs/barista-components/quick-filter/src/state/reducer.ts
@@ -20,7 +20,7 @@ import {
 } from '@dynatrace/barista-components/core';
 import { DtAutocompleteValue } from '@dynatrace/barista-components/filter-field';
 import { Action, ActionType } from './actions';
-import { initialState, QuickFilterState } from './store';
+import { QuickFilterState } from './store';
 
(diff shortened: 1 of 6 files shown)


libs/barista-components/quick-filter/src/state/reducer.ts
```diff
@@ -46,14 +46,14 @@ export function quickFilterReducer(
 
   switch (action.type) {
     case ActionType.SWITCH_DATA_SOURCE:
+      const stateBackup = state;
       if (state.dataSource) {
         state.dataSource.disconnect();
       }
-      return { ...initialState, dataSource: action.payload };
+      return { ...stateBackup, dataSource: action.payload };
```
ffriedl89, line 53:
```suggestion
      return { ...state, dataSource: action.payload };
```
You can spread the state directly, no need to store it in a variable :)

ffriedl89 (changes requested):
No real concerns, looks really good 👍

tomheller (approved):

ffriedl89 (approved):

tomheller (approved):

PR #1986, opened 2021-02-16
fix(filter-field): Improving isOptionSelected method for async values.

PedroMosquera:
#### Type of PR

Bugfix (non-breaking change which fixes an issue).

#### In this PR
- Keeping already selected values, on edit mode, for asynchronous values
- Adding test for current async implementation


tomheller (approved):
Looks good, thank you for this. 
👍 
nevermind the buildkite status, it has some hickups on an infrastructure level.

subarroca (approved):

PR #2029, opened 2021-03-22
fix(filter-field): Fixes the defaultSearch not being displayed anymore

rowa-audil:

Fixes the defaultSearch not being displayed anymore after regression by version 9.3.0

#### Type of PR

#### Checklist

### DIFF
--- a/libs/barista-components/filter-field/src/filter-field-util.ts
+++ b/libs/barista-components/filter-field/src/filter-field-util.ts
@@ -78,24 +78,22 @@ export function filterAutocompleteDef(
 export function filterFreeTextDef(
   def: DtNodeDef,
   filterText?: string,
-): DtNodeDef | null {
+): DtNodeDef {
   const suggestions = def.freeText!.suggestions
     ? def.freeText!.suggestions.filter((option) =>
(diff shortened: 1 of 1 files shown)


libs/barista-components/filter-field/src/filter-field-util.ts
```diff
@@ -78,24 +78,22 @@ export function filterAutocompleteDef(
...
   filterText?: string,
-): DtNodeDef | null {
+): DtNodeDef {
   const suggestions = def.freeText!.suggestions
     ? def.freeText!.suggestions.filter((option) =>
         filterOptionDef(option, new Set(), filterText),
       )
     : [];
 
-  return def.freeText!.async || suggestions.length
```
rowa-audil, line 88: a freeText doesn't always have to be async or have suggestions

tomheller (approved):

libs/barista-components/filter-field/src/filter-field-util.ts
ffriedl89: I've tested the async behavior and it still works for freetext with these changes applied.

ffriedl89 (approved):

PR #2035, opened 2021-03-25
fix(key-value-list): Adds a zero-width white space to split the key from the value.

tomheller:

In chrome, when double clicking the last word of the value, it selected the first part of the value as well. Because the value is set to display: inline; which chrome interprets as a continuous word.
With the inserted 0-width white space, this forces chrome to separate the words.

#### Type of PR
Bugfix

#### Checklist


ffriedl89 (approved):
That is brilliant!

PR #2080, opened 2021-04-22
feat(event-chart): APM-295388 Implemented "heatfield" into event-chart

kacper-olszanski:
event chart heatfields

### DIFF
--- a/libs/barista-components/event-chart/src/event-chart.ts
+++ b/libs/barista-components/event-chart/src/event-chart.ts
@@ -73,20 +73,32 @@ import {
   DtEventChartEvent,
   DtEventChartLane,
   DtEventChartLegendItem,
+  DtEventChartField,
   DtEventChartOverlay,
+  DtEventChartHeatfieldOverlay,
   DtEventChartSelectedEvent,
   DT_EVENT_CHART_COLORS,
+  DtEventChartSelectedField,
 } from './event-chart-directives';
 import { DtEventChartLegend } from './event-chart-legend';
 import { dtCreateEventPath } from './merge-and-path/create-event-path';
-import { dtEventChartMergeEvents } from './merge-and-path/merge-events';
-import { RenderEvent } from './render-event.interface';
+import {
+  dtEventChartMergeEvents,
+  dtEventChartMergeFields,
+} from './merge-and-path/merge-events';
+import {
+  RenderEvent,
+  RenderField,
+  isRenderField,
+  isRenderEvent,
+} from './render-event.interface';
 import {
   DtTimeUnit,
   DtFormattedValue,
   formatDuration,
 } from '@dynatrace/barista-components/formatters';
 
+const FIELD_BUBBLE_SIZE = 4;
 const EVENT_BUBBLE_SIZE = 16;
 const EVENT_BUBBLE_SPACING = 4;
 
@@ -95,6 +107,8 @@ const TICK_HEIGHT = 24;
 const TICK_WIDTH = 140;
 
 const LANE_HEIGHT = EVENT_BUBBLE_SIZE * 3;
+
+const HEATFIELD_OFFSET = 16;
 
 let patternDefsOutlet: PortalOutlet;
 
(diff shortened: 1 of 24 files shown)


libs/barista-components/event-chart/src/event-chart.ts
```diff
@@ -460,15 +589,106 @@ export class DtEventChart<T> implements AfterContentInit, OnInit, OnDestroy {
...
+    );
+
+    path.push('z');
+
+    return path.join(' ');
+  }
+
+  _calculateRects(renderEvent: RenderField<T>): string {
+    const magic_number = FIELD_BUBBLE_SIZE / 2;
+    const magic_number_2 = magic_number / 2;
```
tomheller, line 635: It will be much easier to read the already fairly crazy genius of svg pathing if the variables have at least some semantic meaning. 

```suggestion
    const field_bubble_half = FIELD_BUBBLE_SIZE / 2;
    const field_bubble_quarter = magic_number / 2;
```

libs/barista-components/event-chart/src/event-chart.ts
```diff
@@ -460,15 +589,106 @@ export class DtEventChart<T> implements AfterContentInit, OnInit, OnDestroy {
     return path.join(' ');
   }
 
+  _calculateFieldOutline(
+    renderEvent: RenderField<T>,
+    offset: number = 0,
+  ): string {
+    const eventBubbleRadius = FIELD_BUBBLE_SIZE + offset;
```
tomheller, line 596: You are creating a rectangle with rounded corners, if I'm not mistaken here. 
Any reason why you are not using `rx` on a rectangle to add the rounded corners like this:
`<rect x="120" width="100" height="100" rx="15" />`

tomheller (changes requested):
I tried it locally and it seems there is some weird offset if you hover over the heatfield. Are you aware of that. 
image

It would be great to add a demo in the examples library that can be showcased in barista and shows how heatfields should be used in the eventchart. 

It would be great to have some tests added that test the heatfield functionality

Please also add how to use the heatfield and heatfield documentation in the readme as well.

libs/barista-components/event-chart/src/event-chart.ts
kacper-olszanski: I've had some strange problems with the `rect`

kacper-olszanski:
@tomheller 
> I tried it locally and it seems there is some weird offset if you hover over the heatfield. Are you aware of that.
I am aware if the offfset: I tried to reproduce the style of the heatfield in the normal chart and also implemented the same offsets

kacper-olszanski:
@tomheller 
> I tried it locally and it seems there is some weird offset if you hover over the heatfield. Are you aware of that.

I am aware of the hover offset. I implemented it sop it is more clear if you are selecting overlapping heatfields but I could remove it.

tomheller:
> I am aware of the hover offset. I implemented it sop it is more clear if you are selecting overlapping heatfields but I could remove it.

I'm not saying you need to remove it, but is seems a bit odd with its shape. Imho it looks a bit broken.

tomheller:
Great job @kacper-olszanski !
This is also a cool edition with the overlay on clicked heatfield. Just one thing, can we either connect the position of the overlay to the heatfield or close the overlay on scroll. 

I was able to produce this state by activating one of the heatfields and then scrolling the page. I guess this is not the expected behavior.
image

kacper-olszanski:
> Great job @kacper-olszanski !
> This is also a cool edition with the overlay on clicked heatfield. Just one thing, can we either connect the position of the overlay to the heatfield or close the overlay on scroll.
> 
> I was able to produce this state by activating one of the heatfields and then scrolling the page. I guess this is not the expected behavior.
> image

fixed it for the heatfields and for the other event bubbles

tomheller (approved):
Perfect, this looks good to me now. Thank you for this amazing contribution 🎉 

From your end, you should be done now. The next steps are that the PR will be merged and cherry picked to the respective release branches.  
Your change will be included in the next barista version, which I expect to be released next week.

tomheller:
Cherry picked to the minor release branch 9.x, new head 73219a8d6ea3cba6ee82d9a032f2fb984d11d8d8

PR #2127, opened 2021-05-20
feat(filter-field): Display a hint when partial options are loaded.

giordy:
Fixes #APM-277597

#### Type of PR

#### Checklist

preview

https://user-images.githubusercontent.com/6016950/118957572-ee3e6e80-b960-11eb-9dff-50b62d9a77e8.mp4

### DIFF
--- a/libs/barista-components/filter-field/src/filter-field.html
+++ b/libs/barista-components/filter-field/src/filter-field.html
@@ -87,34 +87,42 @@
           >
         </dt-option>
       </ng-container>
-      <ng-container
-        *ngFor="let optionOrGroupDef of _autocompleteOptionsOrGroups"
-      >
-        <dt-optgroup
-          *ngIf="optionOrGroupDef.group; else options"
-          [label]="optionOrGroupDef.group!.label"
+      <ng-container>
+        <div
+          class="dt-filter-field-hint-partial"
+          *ngIf="_currentDef?.autocomplete?.partial"
         >
-          <dt-option
-            *ngFor="let optionDef of optionOrGroupDef.group!.options"
-            [value]="optionDef"
-            [disabled]="optionDef.option!.disabled"
-            [attr.title]="_getTitle(optionDef.option)"
+          {{ _filterFieldConfig.partialHintMessage }}
+        </div>
+        <ng-container
+          *ngFor="let optionOrGroupDef of _autocompleteOptionsOrGroups"
+        >
+          <dt-optgroup
+            *ngIf="optionOrGroupDef.group; else options"
+            [label]="optionOrGroupDef.group!.label"
           >
-            <dt-highlight [term]="_inputValue">{{
-              optionDef.option!.viewValue
-            }}</dt-highlight>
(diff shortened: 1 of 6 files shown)


libs/barista-components/filter-field/src/filter-field.html
```diff
@@ -87,34 +87,42 @@
...
-            <dt-highlight [term]="_inputValue">{{
-              optionDef.option!.viewValue
-            }}</dt-highlight>
-          </dt-option>
-        </dt-optgroup>
-        <ng-template #options>
-          <dt-option
-            [value]="optionOrGroupDef"
-            [attr.title]="_getTitle(optionOrGroupDef.option)"
+          Pick or provide free text
```
ffriedl89, line 95: We need to provide this string to the component as something the user can set / change. 
Either an Input or via a InjectionToken - e.g. a DtFilterFieldConfig that has a field partialHintMessage.

ffriedl89: What do you prefer @thomaspink ?

thomaspink: Agree with @ffriedl89 that this text has to be provided by the consumer, but not sure what the best approach is. This heavily depends on the use case. Is this text going to be changed a lot in the lifetime of a FilterField? e.g. When traversing down the option and a wild range input appears? Then back to options and onto free-text fields? I guess this text then has to change all the time then?

The options we have (order by the amount of possible changes):

1. Via config: With this option the consumer has the benefit of easily change the text for multiple filter fields but not to update the text a lot of times. This should be used if the text is mostly the same even for multiple filter field instances

2. Via input: Setting the text for each filter field individually, but it is hard to change it multiple times in the lifetime of the FF or when the user traverses the options and other inputs. This should be used if most filter fields have a different text, but it does not change to often in its lifetime. (it is possible for the consumer to change it in the lifetime but this requires some extra steps to achieve this - listening for changes and updating the input accordingly)

3. Via a field in the data (source) itself: This is the best option if the text changes a lot during the lifetime of the FF, because you can set the text for each node individually. We could also implement it in a way that if no property is present in the current node we take the one from the next parent. This would allow setting the text for whole data brackens but still have the option to override it for individual nodes.

4. Combining options above: We can also combine the options (or some of them) to achieve a broader range of possibilities for the user (e.g. input and config, or config and data). With this solution we need a way that one option overrides the other if both are present.

Your thoughts? @ffriedl89

thomaspink (changes requested):

libs/barista-components/filter-field/src/filter-field.html
ffriedl89: I'd go with option 1 and have the current hardcoded string as the default value.

giordy: Thanks for the feedback, I will update the code to have the string message set via config 👍🏻

ffriedl89 (approved):

thomaspink (approved):

PR #2134, opened 2021-05-26
fix(stacked-series-chart): Fixes the initial display of the Y axis.

tonidalmases:

The Y axis in a stacked series chart is not correctly rendered when the chart is loaded with `[visibleValueAxis]="true"`

image

The rendering of this Y axis is done based on a `Subject` that is triggered on `set` functions. However, the subscription to the `Subject` is done after setting all parameters, so the rendering is not performed.

This PR consists on changing the type to `ReplaySubject`, so that all events happening before the subscription will be also considered and the Y axis will be correctly displayed.

image

#### Type of PR


ffriedl89 (approved):

ffriedl89:
Thx for the fix 🔥

PR #2145, opened 2021-05-31
feat(stacked-series-chart): Adding auto scrolling on horizontal overflow

tonidalmases:

This PR implements two changes in stacked series chart, as described in APM-295657.

#### 1. Enabling automatic horizontal scrolling when `mode = 'column'` and the series need more space to be correctly displayed.

Before: https://gyazo.com/7afc817ff23c1493b606b102ec691b46
After: https://gyazo.com/9eb53ddc9a57402b4e94d1b81b2fb4ac

CSS is also adapted for having a correct display on dark mode (and dark mode with the series placed on a Card element).

#### 2. Displaying the cursor as `pointer`  for the whole stack when `selectionMode = 'stack'`.

Before: https://gyazo.com/0aa5573a4b64eeff56c4afe83506b4d9
After: https://gyazo.com/235b64e421660bb072fb7d0f81033d29

e2e for testing the correct behavior depending on the `selectionMode`.

### Type of PR

### DIFF
--- a/libs/barista-components/stacked-series-chart/src/_stacked-series-chart-theme.scss
+++ b/libs/barista-components/stacked-series-chart/src/_stacked-series-chart-theme.scss
@@ -1,0 +1,13 @@
+@import '../../core/src/style/colors';
+
+.dt-theme-dark :host {
+  .dt-stacked-series-chart-value-axis.dt-stacked-series-chart-value-axis {
+    background-color: $gray-860;
+  }
+}
+
+.dt-theme-dark .dt-card :host {
+  .dt-stacked-series-chart-value-axis.dt-stacked-series-chart-value-axis {
+    background-color: #505050;
+  }
+}
--- a/apps/components-e2e/src/components/stacked-series-chart/stacked-series-chart.e2e.ts
+++ b/apps/components-e2e/src/components/stacked-series-chart/stacked-series-chart.e2e.ts
@@ -21,8 +21,8 @@ import {
   barBtn,
   barChart,
   body,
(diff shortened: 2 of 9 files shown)


libs/barista-components/stacked-series-chart/src/_stacked-series-chart-theme.scss
```diff
@@ -0,0 +1,13 @@
...
+
+.dt-theme-dark :host {
+  .dt-stacked-series-chart-value-axis {
+    background-color: $gray-860 !important;
+  }
+}
+
+.dt-theme-dark .dt-card :host {
+  .dt-stacked-series-chart-value-axis {
+    background-color: #505050 !important;
```
ffriedl89, line 11: Please avoid using !important, if you have rules that have the same specificity you can duplicate a class for instance 
```
.dt-stacked-series-chart-value-axis.dt-stacked-series-chart-value-axis
```

ffriedl89 (changes requested):

ffriedl89:
Maybe the commit is a bit misleading since it does more than just adding overflow. Maybe you can rephrase that to fit the changeset a bit better :)

libs/barista-components/stacked-series-chart/src/_stacked-series-chart-theme.scss
tonidalmases: Thanks for the input, I'll apply this change

ffriedl89 (approved):

PR #2146, opened 2021-06-01
fix(sort): Fix default table sorting when the sort direction is set. …

nimrod13:
…Closes APM-305714.

Adjusted the order of the data in the "CPU" column from the table sorting example, which was misleadingly already sorted. Then it could be seen that the initial descending sorting of the "CPU" column would have failed in the previous version.

#### Type of PR

#### Checklist


ffriedl89 (approved):
🔥  Nice, well done!

PR #2151, opened 2021-06-09
style(filter-field): Expands input element inside the filter-field in…

rowa-audil:

Moved the max-width from the input to the infix, which only applies when in edit mode and made the infix grow to remaining space

#### Type of PR

#### Checklist


thomaspink (approved):
LGTM. Well done 👍

PR #2158, opened 2021-06-14
feat(stacked-series-chart): Show highest value in Y axis.

tonidalmases:

Implementation of the APM-295667for the stacked series chart.

**1. Displaying a higher value in the Y axis close to the highest series value. Moreover, adjust the Y axis for the visible nodes when the user clicks on an element of the legend in order to hide it**

Before
image

After
https://gyazo.com/8c703c68b5915e40a8aefffa8c080d32

**2, Fixing the Y axis display when `displayMode = 'percent'`**

Before
image

After
image

#### Type of PR

#### Checklist

### DIFF
--- a/libs/barista-components/stacked-series-chart/src/stacked-series-chart.ts
+++ b/libs/barista-components/stacked-series-chart/src/stacked-series-chart.ts
@@ -144,7 +144,9 @@ export class DtStackedSeriesChart implements OnDestroy, OnInit {
   /** Max value in the chart */
   @Input()
   get max(): number | undefined {
-    return this._max !== undefined ? this._max : getTotalMaxValue(this.series);
+    return this._max !== undefined
+      ? this._max
+      : getTotalMaxValue(this._filledSeries);
   }
   set max(value: number | undefined) {
     if (value !== this._max) {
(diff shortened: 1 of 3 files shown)


libs/barista-components/stacked-series-chart/src/stacked-series-chart.ts
```diff
@@ -596,12 +599,15 @@ export class DtStackedSeriesChart implements OnDestroy, OnInit {
             0.6 +
           1.5,
         relative:
-          (this._axisTicks.slice(-1)[0].valueRelative.toString().length + 1) *
+          ((this._axisTicks.slice(-1)[0].valueRelative * 100).toString()
```
thomaspink, line 602: Not sure what this code is for. Can you help us out there? Seems pretty random with all these numbers :)

tonidalmases: Hi @thomaspink, it's a way to fix the second case (when `displayMode = percent`, only the last characters are displayed `0 %`, instead of the whole number `100 %`, `75 %`, `50 %`, ...). So I propose the adaptation of this line to consider the extra width the Y axis need: `this._axisTicks.slice(-1)[0].valueRelative` returns the highest relative value, a number between 0 and 1. In order to compute the needed width for displaying the percentage numbers I multiply this relative value length by 100.

ffriedl89: @tonidalmases could you add some comments in code please :)

ffriedl89 (approved):

thomaspink (approved):

ffriedl89:
cherry pick done to 9.x

PR #2194, opened 2021-08-02
fix(chart): Fixed chart selection area focus.

thomaspink:


dyladan:
[CLA assistant check](https://cla.oss.dynalabs.io/dynatrace-oss/barista?pullRequest=2194) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.oss.dynalabs.io/dynatrace-oss/barista?pullRequest=2194) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.oss.dynalabs.io/check/dynatrace-oss/barista?pullRequest=2194) it.</sub>

libs/barista-components/chart/src/chart-module.ts
```diff
@@ -40,6 +41,8 @@ const COMPONENTS = [
   DtChartTimestamp,
   DtChartTooltip,
   DtChartSelectionAreaAction,
+  DtChartFocusAnchor,
```
ffriedl89, line 44: Do we really want to export this from the module? isn't this just used internally?

libs/barista-components/chart/src/range/range.spec.ts
```diff
@@ -41,6 +43,10 @@ import { RangeStateChangedEvent } from './range';
 
 // tslint:disable:no-magic-numbers no-unbound-method no-use-before-declare
 
+export class MockChart {
+  _focusTargets = new Set<DtChartFocusTarget>();
+}
+
 describe('DtChart Range', () => {
```
ffriedl89, line 50: I think it would make sense to add a test to check whether focus management works now as expected.

ffriedl89 (changes requested):

libs/barista-components/chart/src/chart-module.ts
thomaspink: yeah. we can make this internal

libs/barista-components/chart/src/range/range.spec.ts
thomaspink: As discussed in our 1on1 session, I added tests for shifting the focus on the anchor directive.

ffriedl89 (approved):

PR #2208, opened 2021-08-16
chore(workspace): Update snyk policy for security checks.

tomheller:
Update security policies. 

Skipping d3 as it will be updated in barista version 10


tomheller:
Every reviewer possible is on vacation, sorry for force merging

PR #2215, opened 2021-08-19
fix(filter-field): Fix multiselect apply button disabled

tonidalmases:

This PR fixes the bug described in APM-311774. For a multiselect filter, when trying to modify the values of the filter, as long as there are any values already checked, the "Apply" button is disabled if the user doesn't make any change (select or unselect items).

The fix consist on enabling the button when editing, even if the user doesn't make any change. The button will be disabled only in the case there are no items selected.

Before: https://gyazo.com/81520da20f273237f704f6c19fa24272
After: https://gyazo.com/bd69ee33b3f88fd4ba67c6b92cb764c0

#### Type of PR

#### Checklist

### DIFF
--- a/apps/components-e2e/src/components/filter-field/filter-field.e2e.ts
+++ b/apps/components-e2e/src/components/filter-field/filter-field.e2e.ts
@@ -473,3 +473,27 @@ test('should close the multiselect when blurring the filter field mid filter, re
     .expect(multiSelectPanel.exists)
     .notOk();
 });
+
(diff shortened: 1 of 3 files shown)


tomheller (approved):

tomheller:
Manually cherry picked to:
9.x 00a29117ce3977929e632280123ec278934e24e8
9.10.x fd3b56395f37fa980489028f992042cbe399fd8d

PR #2172, opened 2021-07-05
feat(icon): Retry loading hiccup resources.

piotrl:
- internal issue: APM-310496 APM-282779
- nowadays CDNs are distributed systems that does not guarantees your request to reach right place. It started happening few hounders times per day for us that some icon couldn't be loaded
- CloudFront explicitly marks that requests *needs* to be retried
- mechanism is fairly simple, we retry any error from the server side (AWS throws 500 / 504 / 503 / 403)
  - 3 retries is enough for other resources we already handle, it should be fine here too
- docs: https://www.learnrxjs.io/learn-rxjs/operators/error_handling/retry

#### Type of PR

Feature (non-breaking change which adds functionality)

#### Checklist

- [*] I have read the CONTRIBUTING doc and I follow the PR guidelines
- [*] Lint and unit tests pass locally with my changes
- [ *] I have added tests that prove my fix is effective or that my feature works
- [ *] I have added necessary documentation (if appropriate)

PR #2257, opened 2021-09-23
feat(icon): Remove log for missing loading resources.

piotrl:
* Previous fix PR #2172 improved spam a lot, we have now instead 500 logs daily, only 1-2
* But still we have them daily, preferably we'd like to not see it again as a hiccup ;)
* Suggestion: Keep severity as INFO instead WARN


thomaspink (approved):

thomaspink:
LGTM. Thank you!

PR #2297, opened 2021-10-27
feat(formatters): Use binary bytes prefixes for bytes formatted using binary factor

adameq11:

<hr>

This change provides a binary prefixes using logic that was prepared for such case. We already had a binary multiplier concept and it is used in library consumers. So whenever a consumer use `formatBytes(n, {factor: KIBI_MULTIPLIER})` method a binary prefixes will be attached to resulting value. Right now we would always get decimal prefixes for both multipliers.

In this change I didn't changed/improved how formatter works - it was not possible before and it is not possible now to convert between decimally and binary formatted values using inputUnit outputUnit parameters. So if you use formatBytes() with binary multiplier you should not use decimal (like MB) inputUnit. That is inherent problem of existing library method and it was not my goal to fix it.<br>
I also didn't change bit formatter as for me binary formatting for bits is not really used in real life.

Internal DT issue for this change - APM-330040.
This change is related to changes in GWT DT codebase which introduce similar changes (see APM-330039). 

Examples from test app:
barista_ex2
barista_ex1

#### Type of PR

It is a Feature, but can also be perceived as a fix, since binary prefixes becoming a standard right now and we should have visual differentiation between binary and decimally formatted bytes.

#### Checklist


tomheller:
Cherry picked to 10.x, new HEAD 5ab0ffb9c2664014f025ba9ccbd7594f1b47f888

PR #2302, opened 2021-11-02
Release stage/10.1.0

tomheller:
Feature release for 10.1.0


tomheller:
An issue with the stacked series chart has been noticed and the release will be canceled. Will be re-opened once I found and fixed the issue.

PR #2333, opened 2021-11-22
Add exportCSV to dt-table

LucasHocker:

<hr>

  often helpful).
  guidelines.

#### Type of PR

Feature: added a context dialogue with an "Export CSV" button

#### Checklist

- [n/a] I have added tests that prove my fix is effective or that my feature works
- [n/a] I have added necessary documentation (if appropriate)

### DIFF
--- a/libs/barista-components/table/src/table.ts
+++ b/libs/barista-components/table/src/table.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { coerceBooleanProperty } from '@angular/cdk/coercion';
+import { BooleanInput, coerceBooleanProperty } from '@angular/cdk/coercion';
 import { Platform } from '@angular/cdk/platform';
 import { CdkPortalOutlet, TemplatePortal } from '@angular/cdk/portal';
 import { DOCUMENT } from '@angular/common';
@@ -62,6 +62,7 @@ import {
   _COALESCED_STYLE_SCHEDULER,
 } from '@angular/cdk/table';
 import { ViewportRuler } from '@angular/cdk/scrolling';
+import { DtTableSelection } from './selection/selection';
 
 interface SimpleColumnsAccessorMaps<T> {
   displayAccessorMap: Map<string, DtSimpleColumnDisplayAccessorFunction<T>>;
@@ -95,6 +96,7 @@ export class DtTable<T> extends _DtTableBase<T> implements OnDestroy {
   private _multiExpand: boolean; // TODO: discuss default value with UX, should maybe change from false to true
   private _loading: boolean;
   private _destroy$ = new Subject<void>();
+  private _showExportButton: boolean = false; //Revert to opt-in instead of opt-out per request
 
   /** Sort accessor map that holds all sort accessor functions from the registered simple columns. */
   private _sortAccessorMap = new Map<
@@ -118,6 +120,32 @@ export class DtTable<T> extends _DtTableBase<T> implements OnDestroy {
   /** @internal Whether a expandable row is registered with the table */
   _hasExpandableRows = false;
 
+  /** @internal Reference to filteredData in table-data-source */
+  _filteredData: T[];
+
+  /** @internal Reference to exportable data. */
+  _exporter: {
+    filteredData: T[];
+    selection: DtTableSelection<T> | null;
+  };
+
+  /** @internal Determine if selection is enabled and properly connected. */
+  get _selectionEnabledAndConnected(): boolean {
+    if (this._exporter && this._exporter.selection) return true;
+    else return false;
+  }
+
+  /** The number of rows that have been selected by the user. */
+  get numSelectedRows(): number {
+    if (
+      this._exporter &&
+      this._exporter.selection != null &&
+      this._exporter.selection.selected
+    )
+      return this._exporter.selection.selected.length;
+    else return 0;
+  }
+
   /** Whether the loading state should be displayed. */
   @Input()
   get loading(): boolean {
@@ -140,6 +168,20 @@ export class DtTable<T> extends _DtTableBase<T> implements OnDestroy {
   get isEmptyDataSource(): boolean {
     return !(this._data && this._data.length);
   }
+
+  /**
+   * Experimental!
+   * Opt-in to the possibility to export all table data, visible table data or
+   * selected table data into a csv file.
+   */
+  @Input()
+  get showExportButton(): boolean {
+    return this._showExportButton;
+  }
+  set showExportButton(value: boolean) {
+    this._showExportButton = coerceBooleanProperty(value);
+  }
+  static ngAcceptInputType_showExportButton: BooleanInput;
 
   /** @internal The snapshot of the current data */
   get _dataSnapshot(): T[] | readonly T[] {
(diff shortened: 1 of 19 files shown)


tomheller (changes requested):
Hi @LucasHocker! Thank you for your contribution.

This looks very good so far. A couple of things to note here, please make the display / usage of the context dialog an `opt-in`. The way you have implemented this now, it will appear on every barista table, without the consumer having the ability to influence this behavior.

With an opt-in switch you can make sure that there is also enough space on the top of the table to position the context-dialog correctly. The negative top and margin values are usually not something we work with, because it can create unpredictable outcome, depending on where on the page the table is rendered. Here is an example from our demos application, that shows that this might be a bit broken.
image

Additionally, please make sure that you revert the changes from the `package-lock.json`, they should not be part of this PR.

LucasHocker:
Thanks Thomas. I suspected there might be a couple concerns with the mandatory approach. Would you be open to talking about an opt-out approach versus opt-in? Relying on each page owner to go back and opt-in could have very low adoption rates due to various other priorities. As an end user, I want to be able to export tabular data any time I encounter it. I don’t think that could be achieved with an opt-in. Thoughts?

I’ll work on a better way to position the context dialogue rather than relative positioning. Most examples of dt-table seem to have white space above the right side of the table, so I was hoping to utilize that rather than creating another row of white space. I’ll play around with some more examples.

The package.json was unintentional; I’ll be sure to remove that change.

LucasHocker:
Made the following changes:
- Moved export button to bottom, aligns with pagination row if present.
- Added opt-out Input.
- Added demo showing opt-out.
- Updated doc to explain opt-out.

Example (w/ pagination):
Screen Shot 2021-11-23 at 11 30 03 AM

Example (w/o pagination):
Screen Shot 2021-11-23 at 11 31 46 AM

Example (opt-out):
Screen Shot 2021-11-23 at 12 06 56 PM

LucasHocker:
Eventually got all the CI tests to succeed, but `pull_request_labeler` step seems to be stuck. Hopefully you can bump that when you get a change to review the change.

tomheller:
> Eventually got all the CI tests to succeed, but pull_request_labeler step seems to be stuck. Hopefully you can bump that when you get a change to review the change.

This is design to only run if certian finishing labels are being applied. It's fine if it is stuck for now. 

One more observation, can you please switch the `opt-out` switch to an `opt-in` one as outlined in my comment above. We do not want to touch every table in the product with an update like this.

LucasHocker:
Updated (hopefully) ready to merge

libs/barista-components/table/src/table.ts
```diff
@@ -62,6 +62,7 @@ import {
   _COALESCED_STYLE_SCHEDULER,
 } from '@angular/cdk/table';
 import { ViewportRuler } from '@angular/cdk/scrolling';
+import { DtTableDataSource } from '..';
```
tomheller, line 65: Please do not import from a barrel file here. Will fix that for you during cleanup, but this would lead to cyclic dependencies as this files is also exported in the barrel.

libs/barista-components/table/src/table.ts
```diff
@@ -141,6 +143,15 @@ export class DtTable<T> extends _DtTableBase<T> implements OnDestroy {
     return !(this._data && this._data.length);
   }
 
+  /** Whether the export button should be displayed. */
+  @Input()
+  get exportButton(): boolean {
```
tomheller, line 148:
```suggestion
  get showExportButton(): boolean {
```

libs/barista-components/table/src/table.ts
```diff
@@ -321,4 +332,41 @@ export class DtTable<T> extends _DtTableBase<T> implements OnDestroy {
 
   /** CSS class added to any row or cell that has sticky positioning applied. */
   protected stickyCssClass = 'dt-table-sticky';
+
+  exportCSV(): void {
```
tomheller, line 336: Missing JSDoc, also this can be considered an internal function. Would need to mark it as such in the JSDoc and prefix with an _

```suggestion
/** @internal Converts the datasources data to csv format, adds it to a blob and initiates a download
  _exportCSV(): void {
```

libs/barista-components/table/src/table.ts
```diff
@@ -321,4 +332,41 @@ export class DtTable<T> extends _DtTableBase<T> implements OnDestroy {
...
+    let ds = this.dataSource as DtTableDataSource<T>;
```
tomheller, line 337: Most of the variable declarations here could be `const. We try to use constants wherever possible

libs/barista-components/table/src/table.ts
```diff
@@ -321,4 +332,41 @@ export class DtTable<T> extends _DtTableBase<T> implements OnDestroy {
...
+    let exportData = ds.filteredData;
+    let csv = '';
+    let keys: string[] = [];
+    //get list of keys
+    if (typeof exportData[0] == 'object') {
+      keys = Object.keys(exportData[0]);
+    }
+    if (!keys.length) return;
```
tomheller, line 345: Codestyle
```suggestion
    if (!keys.length) {
      return;
    }
```

libs/barista-components/table/src/table.ts
```diff
@@ -321,4 +332,41 @@ export class DtTable<T> extends _DtTableBase<T> implements OnDestroy {
...
+    csv += keys.join(',') + '\n';
+
+    exportData.forEach((row) => {
```
tomheller, line 348: Codestyle: prefer a for ... of loop when possible.

libs/barista-components/table/src/table.ts
```diff
@@ -321,4 +332,41 @@ export class DtTable<T> extends _DtTableBase<T> implements OnDestroy {
...
+      let vals = Object.values(row);
+      keys.forEach((key, idx) => {
+        let val = row[key];
+        if (typeof val != 'string') val = val.toString();
```
tomheller, line 352: @LucasHocker The datasource can also hold complex values, not only primitives like the ones in your example. Data structures like this are not uncommon. 
```
 {
      host: 'et-demo-2-win4',
      cpu: 30,
      memory: {
        memoryPerc: 38,
        memoryTotal: 5830000000,
      },
      traffic: 98700000,
    },
  ...
  ```

With your implementation this would very likely result in a [Object object] stringification. I think we need another iteration on this part of the logic to get this rendering correctly.

File: libs/barista-components/table/src/table.ts [hunk 8]
```diff
@@ -321,4 +332,41 @@ export class DtTable<T> extends _DtTableBase<T> implements OnDestroy {
...
+        if (val.includes(',')) csv += `"${val}"`;
+        else csv += val;
+        if (idx < vals.length) csv += ',';
+      });
+      csv += '\n';
+    });
+
+    //make csv document
+    let o = new Blob([csv], { type: 'text/csv' });
```
tomheller, line 361: We might want to split this into a utility function. It does not necessarily be part of the member.

tomheller (changes requested):

LucasHocker:
@tomheller Fixed above requested changes. Also did some refactoring after adding in selection specific export as well. Export functionality is now also documented in the Readme.md with live examples.
Screen Shot 2021-12-07 at 10 46 41 AM

Please take a look when you have a moment to see if we're aligned for a merge.

tomheller:
@LucasHocker Thank you for the effort. I will check it on Thursday, as we have a public holiday tomorrow in Austria.

tomheller:
@LucasHocker All looking good, I was trying to clean up some of the commits and code styles. You will need to give me permission to push to your fork (there should be a button somewhere that allows maintainers to push updates to your branch).

tomheller:
Cherry picked to release branch 10.x 9034a8be7aec7b7e43f3a50f711c561015da8d4f

PR #2342, opened 2021-12-06
feat(components): Increase coverage of translatable messages.

piotrl:
### Goal

this PR increases coverage of i18n messages understood by Angular. Avoids "parts" of the product UI being untranslated.

### Technical details

* Since Angular 11 and Ivy i18n, Angular understands libraries when extracting and building messages
* catched all through tslint `template-i18n` rule, however I didn't enable it as we have same `tslint.json` and no `tslint.spec.json` that would disable this rule for `.spec.ts` (producing many false positives)

#### Type of PR

### DIFF
--- a/libs/barista-components/experimental/datepicker/src/calendar.html
+++ b/libs/barista-components/experimental/datepicker/src/calendar.html
@@ -6,6 +6,7 @@
     (click)="_addMonths(-12)"
     class="dt-calendar-header-button dt-calendar-header-button-prev-year"
     aria-label="Previous Year"
+    i18n-aria-label
   >
     <dt-icon name="learn-more"></dt-icon>
   </button>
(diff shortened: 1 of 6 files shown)


tomheller:
Nice @piotrl! Would you mind rebasing your branch to an up to date version, it seems that you have based your changes on a very outdated master, which means you are still running `bazel` tasks

piotrl:
thanks @tomheller, overlooked it!

tomheller (approved):

tomheller:
The scope of `*` might not be ideal here. I think the correct scope would be `components`. Would you mind rewording the commit?

piotrl:
ok, updated commit message ;)

tomheller:
cherry picked into ac5320f51244c13026e1b2e580feeb27717dc382

PR #2379, opened 2022-01-20
fix(autocomplete): Fixed alignment for autocomplete-overlay if not enough space is available on the right.

mikekuss:
Fixed alignment for autocomplete-overlay if not enough space is available on the right.


tomheller:
Looks good, security checks seem to fail though.

tomheller (approved):

PR #2407, opened 2022-02-02
feat(table): Add possibility to configure table selection in a way th…

pawelszkiler:

In the case of tables with pagination, the global checkbox selects all rows in the currently displayed page (with respect to selection limit and selection enabled / disabled for a given row). 
Now, it is possible to configure table selection so that master checkbox select all rows across all pages.

`providers: [
    {
      provide: DT_TABLE_SELECTION_CONFIG,
      useValue: { globalSelection: true },
    },
  ],`

#### Type of PR

Feature

#### Checklist


libs/barista-components/table/src/selection/selection.ts
```diff
@@ -36,6 +36,13 @@ export interface DtTableSelectionConfig {
    * By default there is no limit
    */
   selectionLimit?: number;
+
+  /**
+   * When set to true and data source is of type DtTableDataSource, master checkbox
```
tomheller, line 41:
```suggestion
   * When set to true and data source is of type DtTableDataSource, global checkbox
```

libs/barista-components/table/src/selection/selection.ts
```diff
@@ -36,6 +36,13 @@ export interface DtTableSelectionConfig {
...
+   * selects rows across all pages (respecting selectionLimit if specified).
+   * By default, master checkbox selection is limited to current page.
```
tomheller, line 43:
```suggestion
   * By default, global checkbox selection is limited to current page.
```

tomheller (commented):
@pawelszkiler I see the use case and the code looks good to me. I however have bit of concern regarding the usability/accessibility of that feature. 
To the user, there is no way of knowing how many items are actually selected (which can be quite problematic). Some tables would only select the current page, some would select the whole dataset. For the user there is no visual feedback or any other feedback what is selected. 

When thinking about action on selection examples (like delete), this would be quite worrisome.

pawelszkiler:
> Some tables would only select the current page, some would select the whole dataset. For the user there is no visual feedback or any other feedback what is selected.
@tomheller I agree with that one. Do you have some suggestions on how this could be improved? At first, I was thinking about some kind of popup/toast that is displayed once the user toggles the master checkbox when `globalSelection` is enabled. However, it may get irritating pretty soon once such a user is used to this screen and knows exactly what are the consequences of clicking on the master checkbox.
Some tooltip when you hover over the checkbox? - can be easily missed

In the UI that will use `globalSelection` feature we have added our custom component that shows selection status ( `numer_of_selected_rows` / `total_rows_count` - maybe that's a way to go? Leave it to dev using table with `globalSelection` enabled? 

Given that there are plenty of solutions here, could we implement this additional feature ('making the user aware that master checkbox selects the whole data set') in a separate PR?

tomheller:
> In the UI that will use globalSelection feature we have added our custom component that shows selection status ( numer_of_selected_rows / total_rows_count - maybe that's a way to go? Leave it to dev using table with globalSelection enabled?

That would definitely be one approach. The only problem I see with that is, that it's now up to the app developer to care and think about this, which can be missed fairly easily as well. 

>Given that there are plenty of solutions here, could this improvement / additional feature ('making the user aware that master checkbox selects the whole data set') be done in a separate PR?

I would rather have this added immediately / in one go because history shows that these _optional additions_ usually happen late, or not at all. 

If you are under time pressure, are you able to shim this on your end as there is no visual difference for this? Could you, for example extend from this component and handle your _one-off_ usecase like this? This would at least get rid of the concerns for consistency on the barista end, as we would need to uphold some standard as everyone is affected by these changes. 
https://github.com/dynatrace-oss/barista/blob/dda55a5ba6a5ce72dd4127ec12b69a2e8ea29ddd/libs/barista-components/table/src/selection/selectors.ts#L101-L121

PR #2466, opened 2022-03-31
Further updates to Table export

LucasHocker:
- Add tests
- Handle gettings labels from complex columns
- Handle case where data is inserted directly in table instead of data source
- Handle up to 5 levels of object recusion for complex data
- Allow devs to pick only visible or only table data
 
<hr>

  often helpful).
  guidelines.

#### Type of PR

#### Checklist

- [X] I have read the CONTRIBUTING doc and I follow the PR guidelines
- [X] Lint and unit tests pass locally with my changes
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have added necessary documentation (if appropriate)

### DIFF
--- a/libs/barista-components/table/src/table.ts
+++ b/libs/barista-components/table/src/table.ts
@@ -25,6 +25,7 @@ import {
   Component,
   ContentChildren,
   ElementRef,
+  EmbeddedViewRef,
   Inject,
   Input,
   IterableDiffers,
@@ -49,6 +50,8 @@ import {
   DtSimpleColumnSortAccessorFunction,
   DtSimpleColumnBase,
 } from './simple-columns/simple-column-base';
+import { DtHeaderCell } from './header/header-cell';
+import { DtRow } from './row';
 import {
   _DisposeViewRepeaterStrategy,
   _ViewRepeater,
@@ -99,7 +102,7 @@ export class DtTable<T> extends _DtTableBase<T> implements OnDestroy {
   private _loading: boolean;
   private _destroy$ = new Subject<void>();
 
-  private _showExportButton: boolean = false; //Revert to opt-in instead of opt-out per request
+  private _showExportButton: boolean | 'visible' | 'table' = false; //Display button for Visible, Table, both (true), or neither (false)
   private _exportExcludeList: string[] = [];
 
   /** Sort accessor map that holds all sort accessor functions from the registered simple columns. */
@@ -179,11 +182,13 @@ export class DtTable<T> extends _DtTableBase<T> implements OnDestroy {
    * selected table data into a csv file.
    */
   @Input()
-  get showExportButton(): boolean {
+  get showExportButton(): boolean | 'visible' | 'table' {
     return this._showExportButton;
   }
-  set showExportButton(value: boolean) {
-    this._showExportButton = coerceBooleanProperty(value);
+  set showExportButton(value: boolean | 'visible' | 'table') {
+    if (value === 'visible' || value === 'table')
+      this._showExportButton = value;
+    else this._showExportButton = coerceBooleanProperty(value);
   }
   static ngAcceptInputType_showExportButton: BooleanInput;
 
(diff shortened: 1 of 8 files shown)


libs/barista-components/table/src/table.ts
```diff
@@ -99,7 +102,7 @@ export class DtTable<T> extends _DtTableBase<T> implements OnDestroy {
   private _loading: boolean;
   private _destroy$ = new Subject<void>();
 
-  private _showExportButton: boolean = false; //Revert to opt-in instead of opt-out per request
+  private _showExportButton: boolean | 'visible' | 'table' = false; //Display button for Visible, Table, both (true), or neither (false)
```
tomheller, line 105: What exactly is the difference between true and `'visible'`? That type seems a bit complicated, can you please elaborate why this extension was necessary?

libs/barista-components/table/src/table.ts
```diff
@@ -212,8 +217,11 @@ export class DtTable<T> extends _DtTableBase<T> implements OnDestroy {
   @ViewChild(CdkPortalOutlet, { static: true })
   _portalOutlet: CdkPortalOutlet;
 
-  /** @internal A list of columns */
-  @ContentChildren(DtSimpleColumnBase) _childColumns: QueryList<
+  /** @internal A list of complex columns */
+  @ContentChildren(DtHeaderCell, { descendants: true })
+  _childComplexColumns: QueryList<DtHeaderCell>;
```
tomheller, line 222: The presence of a HeaderCell makes it a complex definition? What you mean by this is that you are querying for non simple columns, is  this correct?

libs/barista-components/table/src/table.ts
```diff
@@ -212,8 +217,11 @@ export class DtTable<T> extends _DtTableBase<T> implements OnDestroy {
...
+  /** @internal A list of complex columns */
```
tomheller, line 223: Comment is duplicated for the inverse query, please adjust.

libs/examples/src/table/table-export-selection-example/table-export-selection-example.html
```diff
@@ -2,7 +2,7 @@
   [dataSource]="dataSource"
   dtSort
   dtTableSelection
-  showExportButton="true"
+  showExportButton="visible"
```
tomheller, line 5: Why is this change in the API necessary?

tomheller (changes requested):

libs/barista-components/table/src/table.ts
LucasHocker: `visible` only gives the option for the Visible data versus both visible and underlying table data. Mohamed had a usecase where the table data includes many more columns than what’s visible. This would be an alternative to a very long exclude list. I’ll update docs to indicate this is an option, but that devs should prefer `true`.

libs/barista-components/table/src/table.ts
LucasHocker: Correct. Would you prefer something like `nonSimple` instead of complex?

libs/barista-components/table/src/table.ts
LucasHocker: Will fix.

libs/examples/src/table/table-export-selection-example/table-export-selection-example.html
LucasHocker: See above for api change. This example was changed to show the missing table button.

LucasHocker:
I had a question on my unit tests as well. I’m now testing to see if the context dialog shows up based on `showExportButton` but I do not check that the correct buttons show up in the dialog. Do you feel this is necessary? Since the opened dialog shows up in a CdkOutlet, testing this is much more complicated. If think this should be added, I might need assistance from a more experienced angular dev.

tomheller:
> I had a question on my unit tests as well. I’m now testing to see if the context dialog shows up based on showExportButton but I do not check that the correct buttons show up in the dialog. Do you feel this is necessary? Since the opened dialog shows up in a CdkOutlet, testing this is much more complicated. If think this should be added, I might need assistance from a more experienced angular dev.

I don't think it is necessary to check if the context dialog shows up.

libs/barista-components/table/src/table.ts
tomheller: to be quite honest, I don't know what I would prefer here. At least a comment that outlines why this is necessary would be benefitial.

tomheller (approved):

LucasHocker:
Fixed the above and added doc updates. Squashed commits.

tomheller:
> @tomheller, is this good to be merged into 10.x branch?

Looks good from my end, still awaiting review from @ffriedl89

ffriedl89 (approved):

PR #2478, opened 2022-04-11
fix(filter-field): Fixes an issue where the displayed label in edit m…

thomaspink:
fix(filter-field): Fixes an issue where the displayed label in edit mode might overflow in certain cases.


tomheller (approved):

tomheller:
Can we base this one against 10.x as well?

tomheller:
Please don't let the cherry picker run, we are still in a pre-release stage

PR #2497, opened 2022-04-21
fix(loading-distractor): Loading distractor text is now announced by screen readers.

Sherif-Elhefnawy:


libs/barista-components/loading-distractor/src/loading-distractor.ts
```diff
@@ -31,7 +36,23 @@ let nextUniqueId = 0;
...
   _labelId = `"dt-loading-distractor-label-${nextUniqueId++}`;
+
+  /**
+   * @internal
+   * Instance of loading distractor text container
+   */
+  @ViewChild('loadingLabelContainer', { static: false })
+  _loadingLabelContainer: ElementRef;
+
+  constructor(private _zone: NgZone) {}
```
tomheller, line 50: This would be a breaking change.

libs/barista-components/loading-distractor/src/loading-distractor.ts
```diff
@@ -31,7 +36,23 @@ let nextUniqueId = 0;
...
+
+  ngAfterViewInit(): void {
+    // Focusing on the container allows the screen reader to read the label when the loading distractor is shown
+    this._zone.onStable.pipe(take(1)).subscribe(() => {
+      this._loadingLabelContainer.nativeElement.focus();
```
tomheller, line 55: hmm, I'm not sure if this is the right approach. on appearance of the loading distractor this would immediately focus the element. even if the focus was previously somewhere else or the user does not even navigate with a keyboard. 

Imho this will trigger more violations on the a11y tests than it would solve.

tomheller (changes requested):

Sherif-Elhefnawy:
As discussed, we will instead shift to using role='alert'

tomheller (approved):

Sherif-Elhefnawy:
Cherry picked into:
10.7.x: 9181d8e1b3cc6793af9b62ddae95a4f5e3432a88
a11y-fixes-barista-11: 7e5897e1f0d0696dd6f49bff87ded4d16e27bd3d

PR #2499, opened 2022-04-27
fix(card): Reflow of title actions is now possible at small screen si…

Sherif-Elhefnawy:
…zes. Closes APM-358314

Before:
image

After:
image

### DIFF
--- a/libs/barista-components/card/src/card.scss
+++ b/libs/barista-components/card/src/card.scss
@@ -37,6 +37,12 @@
   display: flex;
   width: 100%;
   padding-bottom: 16px;
+
+  // Allow reflow of elements when under a specific screen width
+  @media (max-width: 460px) {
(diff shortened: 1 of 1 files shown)


tomheller (approved):

tomheller:
In general we are trying to go with a mobile first approach, so we would go for min-width queries rather than max-width ones. 

This is only a partial fix though, as we do not have access to container queries, this will only work if the card is a full width one. If nested in another layout, using only half the screen, this will still cause reflow issues. But that should be handled by the surrounding layout

Sherif-Elhefnawy:
Cherry picked into:
10.7.x: abeb9beab92ac34e1f4e536574f74d834398500a
a11y-fixes-barista-11: e0d0a2ed6c9b16169c62d84a8c32ea2728f2e2e1

PR #2571, opened 2022-08-17
feat(filter-field): Adding keyboard functionality to multiselect dropbox

manulopez2dynatrace:

<hr>

#### Type of PR

#### Checklist

#### Changes

User now is able to navigate from the filter bar to the multiselect options and the other way around, due to that, the auto focus is not working anymore (autoActiveFirstOption) as the focus is set via arrow key navigation.

Cases to consider: 

***Acceptance Criteria 1***
HAVING the filter bar
THEN selecting a multiselect filter
AND start typing
THEN no option should be focused by default
THEN the user press the down arrow key
AND the first option should be focused

***Acceptance Criteria 2***
HAVING the filter bar
THEN selecting a multiselect filter
AND start typing
THEN no option should be focused by default
THEN the user press the up arrow key
THEN it scrolls down to the end of the list
AND the last option must be focused

***Acceptance Criteria 3***
HAVING the first multiselect option focused
THEN the user press the up arrow key
AND the filter field should be focused
AND no option in the list should be focused

***Acceptance Criteria 4***
HAVING the last multiselect option focused
THEN the user press the down arrow key
AND the filter field should be focused
AND no option in the list should be focused

*Note: Navigating through the options with the arrow keys should be still working as it is at this moment*

### DIFF
--- a/libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select-trigger.ts
+++ b/libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select-trigger.ts
@@ -78,6 +78,15 @@ export class DtFilterFieldMultiSelectTrigger<T>
     super.elementDisabled = value;
   }
 
+  /**
+   * It will allow to check if the key manager is already focused on the first active item on multi select panel
+   */
+  isKeyManagerFocusAlreadyOnTop = false;
+  /**
+   * It will allow to check if the key manager is already focused on the last active item on multi select panel
+   */
+  isKeyManagerFocusAlreadyOnBottom = false;
+
   constructor(
     protected _elementRef: ElementRef,
     protected _overlay: Overlay,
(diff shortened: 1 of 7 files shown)


Sherif-Elhefnawy:
@manulopez2dynatrace, seems that the dropdown in case of Multi-select stays open on hitting Tab, which shouldn't be the case.

libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select-trigger.ts
```diff
@@ -106,10 +109,105 @@ export class DtFilterFieldMultiSelectTrigger<T>
   /** Opens the filter-field multiSelect panel. */
   openPanel(): void {
     if (!this.element._isOpen) {
+      this.resetActiveItemAndKeyManagerLimitsOnPanel();
```
manulopez2dynatrace, line 112: Add comment about why this was done.

libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select-trigger.ts
```diff
@@ -124,7 +222,7 @@ export class DtFilterFieldMultiSelectTrigger<T>
 
     if (!this.element._applyDisabled && keyCode === ENTER && this.panelOpen) {
       this.element._handleSubmit(event);
-      this._resetActiveItem();
```
manulopez2dynatrace, line 127: Multi-select panel should close on pressing TAB

libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select-trigger.ts
```diff
@@ -147,15 +245,16 @@ export class DtFilterFieldMultiSelectTrigger<T>
         this.element._keyManager.activeItem !== prevActiveItem
       ) {
         this._scrollToOption();
+        this.element._keyManager.activeItem
```
manulopez2dynatrace, line 248: Add comment about what this does

libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-multi-select.ts
```diff
@@ -64,19 +63,6 @@ export class DtFilterFieldMultiSelectSubmittedEvent<T> {
 export class DtFilterFieldMultiSelect<T>
   implements DtFilterFieldElement<T>, AfterViewInit
 {
-  /**
-   * Whether the first option should be highlighted when the multi-select panel is opened.
-   * Can be configured globally through the `DT_MULTI_SELECT_DEFAULT_OPTIONS` token.
-   */
-  @Input()
```
Sherif-Elhefnawy, line 71: @ffriedl89, how do you feel about the whole auto focus removal?

The issue is that the Multi-select is kind of broken due to the following:
1- Auto focus only works the first time the panel is opened. If the user closes the panel and then opens it again, no option is auto focused. 
2- While Panel is open, pressing Tab DOESN'T close the panel and just tabs through normally while keeping the panel open.
3- Entering a text which contains a space, while having an option in focus, caused the input not to accept space as a valid character.

libs/barista-components/filter-field/src/filter-field-multi-select/filter-field-with-multiselect.spec.ts
```diff
@@ -62,10 +65,14 @@ describe('DtFilterField', () => {
...
   ) => void;
 
+  const scrollIntoViewMock = jest.fn();
+  window.HTMLElement.prototype.scrollIntoView = scrollIntoViewMock;
+
   beforeEach(() => {
     ({
       fixture,
       filterField,
+      inputEl,
```
Sherif-Elhefnawy, line 75:
```suggestion
      inputElement,
```

libs/barista-components/filter-field/src/filter-field.html
```diff
@@ -50,8 +54,16 @@
       "
       (keydown)="_handleInputKeyDown($event)"
       (keyup)="_handleInputKeyUp($event)"
+      (focus)="_handleFocus()"
+      (mousemove)="_handleMouseMove()"
       [value]="_inputValue"
       [disabled]="disabled"
+      [readonly]="
+        loading ||
```
Sherif-Elhefnawy, line 62: Move this into a method :)

libs/barista-components/filter-field/src/filter-field.scss
```diff
@@ -126,6 +126,10 @@ $dt-filter-field-inner-height: 30px;
   outline: none;
   background: transparent;
   z-index: 1;
+  &.dt-not-blinking {
```
Sherif-Elhefnawy, line 129: You don't need this

Sherif-Elhefnawy (changes requested):

Sherif-Elhefnawy (approved):

Sherif-Elhefnawy (changes requested):
- Please squash all commits into one

Sherif-Elhefnawy (approved):

PR #2572, opened 2022-08-17
fix(table): Table export prefers data in DS.

LucasHocker:
ie if new data is added in DS, now included in table export. (APM-384237)

<hr>

  often helpful).
  guidelines.

#### Type of PR

#### Checklist

- [ x] I have read the CONTRIBUTING doc and I follow the PR guidelines
- [ x] Lint and unit tests pass locally with my changes
- [ x] I have added tests that prove my fix is effective or that my feature works
- [ x] I have added necessary documentation (if appropriate)

### DIFF
--- a/libs/barista-components/table/src/table.ts
+++ b/libs/barista-components/table/src/table.ts
@@ -53,6 +53,7 @@ import {
 import { DtHeaderCell } from './header/header-cell';
 import { DtRow } from './row';
 import {
+  isDataSource,
   _DisposeViewRepeaterStrategy,
   _ViewRepeater,
   _VIEW_REPEATER_STRATEGY,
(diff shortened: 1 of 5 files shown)


libs/barista-components/table/src/table.ts
```diff
@@ -442,6 +437,38 @@ export class DtTable<T> extends _DtTableBase<T> implements OnDestroy {
     }
     return csvObj;
 
+    /** @internal data can end up in a table many ways, figure out the best to use for exporting table data */
+    function determineDataToExport(table: DtTable<T>): T[] | null {
```
Sherif-Elhefnawy, line 441: This method needs rethinking since it is internal in another method, which uses `DS` from the body of its parent method. It isn't very easy to follow and can be simplified. 

The assumption that `DS` is always a `Datasource` is incorrect.`

`table._filteredData` is actually empty if DS was a data array and is only updated once on `connect` called under `DataSoruce`

Sherif-Elhefnawy (changes requested):

libs/barista-components/table/src/table.ts
LucasHocker: Refactored this a bit to hopefully be more clear and to not transparently use anything from the parent function. `DataSource` is now handled safely. I think all of the various ways data gets into a table are now handled, but please correct me if I'm missing something.

Sherif-Elhefnawy (approved):

Sherif-Elhefnawy:
@LucasHocker, could you please open a PR for master as well?

PR #2589, opened 2022-09-14
fix(filter-field): Focus is now placed on 'view more' button after ex…

Sherif-Elhefnawy:
After change:
quick-filter-group-focus

### DIFF
--- a/libs/barista-components/quick-filter/src/quick-filter-group.html
+++ b/libs/barista-components/quick-filter/src/quick-filter-group.html
@@ -50,7 +50,9 @@
       }"
     ></ng-container>
 
-    <button class="dt-show-more" (click)="_showMore()" i18n>View more</button>
+    <button #viewMoreButton class="dt-show-more" (click)="_showMore()" i18n>
+      <span>View more</span>
+    </button>
   </ng-container>
 
   <ng-template #checkBox let-item>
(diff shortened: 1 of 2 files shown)


libs/barista-components/quick-filter/src/quick-filter-group.html
```diff
@@ -50,7 +50,9 @@
       }"
     ></ng-container>
 
-    <button class="dt-show-more" (click)="_showMore()" i18n>View more</button>
+    <button #viewMoreButton class="dt-show-more" (click)="_showMore()" i18n>
+      <span>View more</span>
```
tomheller, line 54: Why is this span necessary in the button now? Should not make any difference.

tomheller (approved):
Can you check which `View More` gets focused, if you go into the details of the second group?

libs/barista-components/quick-filter/src/quick-filter-group.html
Sherif-Elhefnawy: Due to some formatting shenanigans, space is added around 'View more', a span is added here to ensure this doesn't happen

Sherif-Elhefnawy:
> Can you check which `View More` gets focused, if you go into the details of the second group?

Just for you, :D
quick-filter-group-focus-2

Sherif-Elhefnawy:
Cherry-picked into:
10.8.x: 7fcfc918e6af74deaeff8c1fd34f5f3e7a85e062
a11y-fixes-barista-11: 73e1d64fcae47a6311ff87de927621662c4fb800

PR #2591, opened 2022-09-14
feat(chart): Add default aria-label and ability to override it.

raymond-shanley-dynatrace:
* Add default aria-label which describes the series in the chart.
* Add the ability to override the default aria-label via an @input.

ADES-138


tomheller (approved):

raymond-shanley-dynatrace:
Cherry-picked 9bb0e37d5aafbf9254a25328b7b6f5a06870d4ee to 10.8.x and a11y-fixes-barista-11

PR #2620, opened 2022-12-01
fix(progress-bar): Fix faulty progress calculation when min and max a…

heartdisease:
Fixed faulty algorithm for calculation the progress percentage of the progress component mixin. We did not account for a division-by-zero error.

### DIFF
--- a/libs/barista-components/progress-bar/src/progress-bar.spec.ts
+++ b/libs/barista-components/progress-bar/src/progress-bar.spec.ts
@@ -92,6 +92,58 @@ describe('DtProgressBar', () => {
     expect(progressComponent.value).toBe(0);
   });
 
+  it('should set progress to 0 if min is greater than max', () => {
+    const fixture = createComponent(BasicProgressBar);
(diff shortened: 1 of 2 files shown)


Sherif-Elhefnawy (approved):

raymond-shanley-dynatrace:
> Fixed faulty algorithm for calculation the progress percentage 

"calculation" -> "calculating" or "calculation of"

libs/barista-components/progress-bar/src/progress-bar.spec.ts
```diff
@@ -92,6 +92,58 @@ describe('DtProgressBar', () => {
     expect(progressComponent.value).toBe(0);
   });
 
+  it('should set progress to 0 if min is greater than or equal to max', () => {
```
raymond-shanley-dynatrace, line 95: > should set progress to 0 if min is greater than or equal to max
I would leave out "or equal to" in the title of this test because of the separate test for "equal to"

tomheller (approved):

PR #2664, opened 2023-02-21
feat(stacked-series-chart): Disabled node.

robert-jankowski-dt:
### <strong>New feature | Stacked series chart - disabled node</strong>
Single node of stacked series chart now cannot be set to 'disabled', meaning that it can be hovered over, but cannot be selected. Switch is optional and doesn't do anything unless it is set true.

#### Checklist

- [V] I have read the CONTRIBUTING doc and I follow the PR guidelines
- [V] Lint and unit tests pass locally with my changes
- [V] I have added tests that prove my fix is effective or that my feature works
- [V] I have added necessary documentation (if appropriate)

### DIFF
--- a/libs/barista-components/stacked-series-chart/src/stacked-series-chart.spec.ts
+++ b/libs/barista-components/stacked-series-chart/src/stacked-series-chart.spec.ts
@@ -35,7 +35,10 @@ import { DtIconModule } from '@dynatrace/barista-components/icon';
 import { DtThemingModule } from '@dynatrace/barista-components/theming';
 import { createComponent, dispatchFakeEvent } from '@dynatrace/testing/browser';
 import { DtStackedSeriesChart } from './stacked-series-chart';
-import { stackedSeriesChartDemoDataCoffee } from './stacked-series-chart.mock';
+import {
+  stackedSeriesChartDemoDataCoffee,
+  stackedSeriesChartDemoDataWithDisableNode,
+} from './stacked-series-chart.mock';
 import { DtStackedSeriesChartModule } from './stacked-series-chart.module';
 import {
   DtStackedSeriesChartFillMode,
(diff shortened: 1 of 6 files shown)


libs/barista-components/stacked-series-chart/src/stacked-series-chart.spec.ts
```diff
@@ -366,6 +369,21 @@ describe('DtStackedSeriesChart', () => {
 
       expect(getSelectedSlice()).toBe(null);
     }));
+
+    it('should not select disable node', fakeAsync(() => {
```
tomheller, line 373:
```suggestion
    it('should not select disabled node', fakeAsync(() => {
```

libs/barista-components/stacked-series-chart/src/stacked-series-chart.ts
```diff
@@ -541,7 +541,11 @@ export class DtStackedSeriesChart implements OnDestroy, OnInit {
     series?: DtStackedSeriesChartSeries,
     node?: DtStackedSeriesChartNode,
   ): false | void {
-    return this._isNodeSelectionMode && this._toggleSelect(series, node);
+    return (
+      this._isNodeSelectionMode &&
+      !node?.disabled &&
```
tomheller, line 546: Nit: We could be a bit more explicit here as we had shifty boolean value checks in the past.
```suggestion
      node?.disabled !== false &&
```

tomheller (dismissed):
Looks good to me. 

Please note that a PR against master will contribute to Barista version 11 (the next breaking change version, which is currently only in a release candidate state). 

If you want to have that in the current major version, please also add a PR with the same changes based off 10.x branch.

libs/barista-components/stacked-series-chart/src/stacked-series-chart.ts
robert-jankowski-dt: That way we also need to invert boolean value, so it will be `node?.disabled !== true`

robert-jankowski-dt: That way we need to invert boolean value, so `node?.disabled !== true`

